PR #140, opened 2012-04-20
Do not require exception messages to exactly match MRI

jfirebaugh:
Different Ruby implementations deserve latitude in the error messages, e.g.
to provide more precise diagnostics than MRI.

In particular, do not rely on the fact that RuntimeError generated by MRI
when calling bare `raise` without a current exception has an empty message.
That is not the case on Rubinius; it is arguably a bug in MRI.

Following this commit, all test pass on Rubinius head.


yvettep321 (commented):
thank you

PR #255, opened 2013-05-23
have_{more/less}_than matchers

christhekeele:
I frequently find myself using `have_at_least(3+1)`, when what I really want is `have_more_than(3)`, so I have to add one to my expected value. Eventually I'll introduce an off-by-one error into my tests and lose a little time debugging. It's the same story with `have_at_most` => `have_less_than`. So, I added these two new matchers.

An added benefit of these matchers is that the helpful negative failure messages for the `have_relative` family can recommend contrapositives that don't require you to change the expected value by one. I implemented this too:

```
FactoryGirl.create_list(:user, 5)
expect( User.all ).not_to have_at_most(5).users

# Used to report on failure:
Failure/Error: expect( User.all ).not_to have_at_most(5).users
   Isn't life confusing enough?
   Instead of having to figure out the meaning of this:
     expect(actual).not_to have_at_most(5).users
   We recommend that you use this instead:
     expect(actual).to have_at_least(6).users

# Now reports on failure:
Failure/Error: expect( User.all ).not_to have_at_most(5).users
   Isn't life confusing enough?
   Instead of having to figure out the meaning of this:
     expect(actual).not_to have_at_most(5).users
   We recommend that you use this instead:
     expect(actual).to have_more_than(5).users
```

And etcetera for the other three `have_relative` helpers.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 
 Enhancements
 
+* Add `have_more_than` and `have_less_than` matchers. (christhekeele)
 * Expand `yield_control` so that you can specify an exact or relative
   number of times: `expect { }.to yield_control.exactly(3).times`,
   `expect { }.to yield_control.at_least(2).times`, etc (Bartek
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -462,6 +462,18 @@ module RSpec
       BuiltIn::Have.new(n, :at_most)
     end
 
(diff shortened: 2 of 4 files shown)


NameYourOwn (commented):
Welp, it is nice to have a discussion before anything else. That's what I think, anyway hopefully I'm not migrating to early. Bit concerned with phishing attack which they label as "social engineering".

PR #502, opened 2014-03-18
Add support for key word argument matching with `respond_to`

### LINKED ISSUE #471: respond_to(:method).with(some).arguments should use method signature from rspec-mocks
The [arity check](https://github.com/rspec/rspec-expectations/blob/b447ee4f6c6eee2fa6072a3cf2859e682eb92468/lib/rspec/matchers/built_in/respond_to.rb#L51-L61) in the matcher doesn't handle many cases that it should.  We should leverage the method signature logic from rspec-mocks by moving it into rspec-support.  Then we can use it for `respond_to`.  We can also make it support keyword args -- e.g. `expect(obj).to respond_to(:foo).with(:x, :y).keyword_arguments` or similar.

JonRowe:
First pass at the final part of #471, probably needs more specs but WDYT in general?

(Fixes #471 so it'll autoclose)

### DIFF
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -19,8 +19,6 @@ module RSpec
         # @example
         #   expect(obj).to respond_to(:message).with(1).argument
         #   expect(obj).to respond_to(:message).with(3).arguments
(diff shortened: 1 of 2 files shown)


myronmarston:
This is nearly a year old.  Gonna close it since it's so stale.  #471 is still open so hopefully someone will tackle that at some point.

PR #709, opened 2015-01-02
Add `be_a_subclass_of` matcher.

thomasfedb:
I've added a `be_subclass_of` matcher to the builtins.

I realise that it is possible to use the `be` matcher for this, like so `expect(String).to be < Object`. This however, feels rather cryptic. I feel that `expect(String).to be_subclass_of Object` is clearer.

I've included a specification, as well as a cucumber feature based off of the `be_kind_of` feature.

Let me know if there's anything I can do to make this easier to merge.

### DIFF
--- a/features/built_in_matchers/class.feature
+++ b/features/built_in_matchers/class.feature
@@ -1,0 +1,56 @@
+Feature: Class matchers
+
+  rspec-expectations a matcher to specify the nature of classes:
+
+    * `expect(Klass).to be_subclass_of(Klazz)`: calls `Klass.ancestors.include?(Klazz)`, which returns true if
+        Klazz is in Klass's class hierarchy or is a module and is included in a class in Klass's
+        class hierarchy. This matcher is aliased as `be_a_subclass_of` and `a_subclass_of`.
+
+
+  Scenario: be_(a_)subclass_of matcher
+    Given a file named "be_subclass_of_matcher_spec.rb" with:
+      """ruby
+      module Qux
+
+      end
+
+      class Foo
+
+      end
+      
+      class Bar < Foo
+
+      end
+
+      class Baz < Bar
+        include Qux
+      end
+
+      RSpec.describe Baz do
+        subject { Baz }
+
+        # the superclass
+        specify { expect(subject).to be_subclass_of(Bar) }
+
+        # the super-superclass
+        specify { expect(subject).to be_subclass_of(Foo) }
+
+        # Ruby internals
+        specify { expect(subject).to be_subclass_of(Object) }
+
+        # an included module
+        specify { expect(subject).to be_subclass_of(Qux) }
+
+        # negative passing case
+        it { is_expected.not_to be_subclass_of(String) }
+
+        # deliberate failures
+        specify { expect(subject).not_to be_subclass_of(Foo) }
+        specify { expect(subject).to be_subclass_of(Fixnum) }
+      end
+      """
+    When I run `rspec be_subclass_of_matcher_spec.rb`
+    Then the output should contain all of these:
+      | 7 examples, 2 failures                   |
+      | expected Baz not to be a subclass of Foo |
+      | expected Baz to be a subclass of Fixnum  |
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -352,6 +352,19 @@ module RSpec
     alias_method :be_kind_of, :be_a_kind_of
     alias_matcher :a_kind_of,  :be_a_kind_of
 
+    # Passes if actual.ancestors.include?(expected)
+    #
+    # @example
+    #
+    #   expect(Integer).to    be_a_kind_of(Fixnum)
+    #   expect(String).to     be_a_kind_of(Object)
+    #   expect(String).not_to be_a_kind_of(Integer)
+    def be_a_subclass_of(expected)
+      BuiltIn::BeASubclassOf.new(expected)
+    end
+    alias_method :be_subclass_of, :be_a_subclass_of
+    alias_matcher :a_subclass_of,  :be_a_subclass_of
+
     # Passes if actual.between?(min, max). Works with any Comparable object,
     # including String, Symbol, Time, or Numeric (Fixnum, Bignum, Integer,
     # Float, Complex, and Rational).
--- a/lib/rspec/matchers/built_in.rb
+++ b/lib/rspec/matchers/built_in.rb
@@ -12,6 +12,7 @@ module RSpec
     module BuiltIn
       autoload :BeAKindOf,               'rspec/matchers/built_in/be_kind_of'
       autoload :BeAnInstanceOf,          'rspec/matchers/built_in/be_instance_of'
+      autoload :BeASubclassOf,           'rspec/matchers/built_in/be_subclass_of'
       autoload :BeBetween,               'rspec/matchers/built_in/be_between'
       autoload :Be,                      'rspec/matchers/built_in/be'
       autoload :BeComparedTo,            'rspec/matchers/built_in/be'
--- a/spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
+++ b/spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
@@ -1,0 +1,45 @@
+SuperKlass = Class.new
+SubKlass = Class.new(SuperKlass)
+SubSubKlass = Class.new(SubKlass)
+OtherKlass = Class.new
+
+module RSpec
+  module Matcher
+    [:be_a_subclass_of, :be_subclass_of].each do |method|
+      RSpec.describe "expect(actual).to #{method}(expected)" do
+
+        it_behaves_like "an RSpec matcher", :valid_value => SubKlass, :invalid_value => OtherKlass do
+          let(:matcher) { send(method, SuperKlass) }
+        end
+
+        it "passes if actual is subclass of expected class" do
+          expect(SubKlass).to send(method, SuperKlass)
+        end
+
+        it "passes if actual is subklass of subclass of expected class" do
+          expect(SubSubKlass).to send(method, SuperKlass)
+        end
+
+        it "fails with failure message for should unless actual is subclass of expected class" do
+          expect {
+            expect(OtherKlass).to send(method, SuperKlass)
+          }.to fail_with(%Q{expected OtherKlass to be a subclass of SuperKlass})
+        end
+
+        it "provides a description" do
+          matcher = send(method, SuperKlass)
+          matcher.matches?(OtherKlass)
+          expect(matcher.description).to eq "be a subclass of SuperKlass"
+        end
+      end
+
+      RSpec.describe "expect(actual).not_to #{method}(expected)" do
+        it "fails with failure message for should_not if actual is subclass of expected class" do
+          expect {
+            expect(SubKlass).not_to send(method, SuperKlass)
+          }.to fail_with(%Q{expected SubKlass not to be a subclass of SuperKlass})
+        end
+      end
+    end
+  end
+end
--- a/lib/rspec/matchers/built_in/be_subclass_of.rb
+++ b/lib/rspec/matchers/built_in/be_subclass_of.rb
@@ -1,0 +1,16 @@
+module RSpec
+  module Matchers
+    module BuiltIn
+      # @api private
+      # Provides the implementation for `be_a_subclass_of`.
+      # Not intended to be instantiated directly.
+      class BeASubclassOf < BaseMatcher
+      private
+
+        def match(expected, actual)
+          actual.ancestors.include? expected
+        end
+      end
+    end
+  end
+end


myronmarston:
Thanks for the well written PR, @thomasfedb.  I do have some concerns about this, however:

* `expect(klass).to be_a_subclass_of(klass)` passes for any class even though we would never say a class is a subclass of itself, since classes are in their own ancestors list.
* `expect(Array).to be_a_subclass_of(Enumerable)` also passes, but I wouldn't normally speak of a class subclassing its mixins.
* In ruby it seems like a code smell to specify that a class is a subclass of another class in its tests.  It goes against the spirit of duck typing, IMO.
* I can't think of a single time I've ever wanted this or heard of any users wanting this (before your PR).

Given my concerns above and the fact that it's literally 3 lines of code to define it in projects that want it:

``` ruby
RSpec::Matchers.define :be_a_subclass_of do |ancestor|
  match { |klass| klass.ancestors.include?(ancestor) }
end
```

...my instinct is to not merge this and instead to leave it to the few users who want it to define it in their projects.

thomasfedb:
Hey @myronmarston 

The first two points I can resolve quite simply I believe. Something along the lines of `actual != expected && expected.kind_of?(Class)`. Or perhaps better to raise an exception if the matcher is instantiated with a non-Class. Let me know if you think it's worth going ahead to resolve.

With regards to your concern about the use of such a matcher being a smell, I do think there are valid cases for its use. Particularly in situations involving factory objects, decorators, and the like, especially given that Ruby is a reasonably metaprogramming-happy language.

For example, I worked on some code recently that dynamically created classes, and wished to assert that they descended from a particular other class depending on the details of their creation. I've certainly seen such assertions in code I've worked in, and found them reasonable in their context.

I think that if you are including `is_kind_of` then `is_subclass_of` isn't a particularly odd divergence. A lot of them time in Ruby we just want our object to `#quack`, but sometimes you need to be sure you're dealing with a `Duck`.

myronmarston:
Thanks for following up, @thomasfedb.  I'm still on the fence about this but you've brought me a bit closer to being in favor.  I'm curious what other RSpec core folks think... /cc @rspec/rspec ?

> The first two points I can resolve quite simply I believe. Something along the lines of `actual != expected && expected.kind_of?(Class)`. Or perhaps better to raise an exception if the matcher is instantiated with a non-Class. Let me know if you think it's worth going ahead to resolve.

The `actual != expected` check makes sense for the `expect(klass).to be_a_subclass_of(klass)` case.  For the module case, I'm on the fence about what the semantics should be.

> I think that if you are including `is_kind_of` then `is_subclass_of` isn't a particularly odd divergence. A lot of them time in Ruby we just want our object to #quack, but sometimes you need to be sure you're dealing with a Duck.

`be_kind_of` and `be_subclass_of` are quite different, IMO.  The former operates on objects, and specifies something about it's class.  The latter operates only on classes, and specifies something about it.  I see it as different because in most cases `be_subclass_of` is a tautology (e.g. `class User < ActiveRecord::Base` already states `User` is a subclass of `ActiveRecord::Base`, so `expect(User).to be_a_subclass_of(ActiveRecord::Base)` is redundant).  `be_kind_of` doesn't run into the same redundancy.

You mention wanting this for a special case of dynamically creating classes, and in that case I can see it serving a purpose but that feels like a special case to me.  I don't see this matcher ever being useful for the "normal" way of defining classes.

Anyhow, I'm still curious what others think.  I'm going to comment on the diff now but don't worry about addressing it until we've decided if this is going to be included or not.

features/built_in_matchers/class.feature
```diff
@@ -0,0 +1,56 @@
+Feature: Class matchers
+
+  rspec-expectations a matcher to specify the nature of classes:
```
myronmarston: This is missing a verb.

features/built_in_matchers/class.feature
```diff
@@ -0,0 +1,56 @@
...
+
+    * `expect(Klass).to be_subclass_of(Klazz)`: calls `Klass.ancestors.include?(Klazz)`, which returns true if
+        Klazz is in Klass's class hierarchy or is a module and is included in a class in Klass's
+        class hierarchy. This matcher is aliased as `be_a_subclass_of` and `a_subclass_of`.
```
myronmarston: It seems weird to use a bullet list for only one thing.  Can this be combined with the sentence above into a single paragraph?

features/built_in_matchers/class.feature
```diff
@@ -0,0 +1,56 @@
...
+
+      end
+      
+      class Bar < Foo
+
+      end
+
+      class Baz < Bar
+        include Qux
+      end
```
myronmarston: These module and class definitions are taking up a lot of room.  I'd prefer to see them in a more compact style like `module Qux; end`.

features/built_in_matchers/class.feature
```diff
@@ -0,0 +1,56 @@
...
+
+        # an included module
+        specify { expect(subject).to be_subclass_of(Qux) }
+
+        # negative passing case
+        it { is_expected.not_to be_subclass_of(String) }
+
+        # deliberate failures
+        specify { expect(subject).not_to be_subclass_of(Foo) }
+        specify { expect(subject).to be_subclass_of(Fixnum) }
```
myronmarston: We consider explicitly referencing the `subject` from an example to be a a code smell:

http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/

In addition, you've mixed `expect(subject)` with `is_expected` (which does the exact same thing). It would be good to stick with one style. Elsewhere we've tended to use `is_expected` for its terseness (to keep the focus on the matchers) but this case is slightly weird in that the default subject is an instance of `Baz` (thus forcing you to explicitly define it).  Given that, I think I might favor using `specify { expect(Baz).to ... }`.

features/built_in_matchers/class.feature
```diff
@@ -0,0 +1,56 @@
...
+      end
+      """
+    When I run `rspec be_subclass_of_matcher_spec.rb`
+    Then the output should contain all of these:
+      | 7 examples, 2 failures                   |
+      | expected Baz not to be a subclass of Foo |
+      | expected Baz to be a subclass of Fixnum  |
```
myronmarston: This needs a line break at the end of the file.

lib/rspec/matchers.rb
```diff
@@ -352,6 +352,19 @@ def be_a_kind_of(expected)
     alias_method :be_kind_of, :be_a_kind_of
     alias_matcher :a_kind_of,  :be_a_kind_of
 
+    # Passes if actual.ancestors.include?(expected)
+    #
+    # @example
+    #
+    #   expect(Integer).to    be_a_kind_of(Fixnum)
+    #   expect(String).to     be_a_kind_of(Object)
+    #   expect(String).not_to be_a_kind_of(Integer)
```
myronmarston: These docs are using the wrong matcher...

lib/rspec/matchers/built_in.rb
```diff
@@ -12,6 +12,7 @@ module Matchers
     module BuiltIn
       autoload :BeAKindOf,               'rspec/matchers/built_in/be_kind_of'
       autoload :BeAnInstanceOf,          'rspec/matchers/built_in/be_instance_of'
+      autoload :BeASubclassOf,           'rspec/matchers/built_in/be_subclass_of'
```
myronmarston: It seems odd for the primary method to be `be_a_subclass_of` but the implementation file to be `be_subclass_of`.  Can we make these consistent?

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect(matcher.description).to eq "be a subclass of SuperKlass"
+        end
+      end
+
+      RSpec.describe "expect(actual).not_to #{method}(expected)" do
+        it "fails with failure message for should_not if actual is subclass of expected class" do
+          expect {
+            expect(SubKlass).not_to send(method, SuperKlass)
+          }.to fail_with(%Q{expected SubKlass not to be a subclass of SuperKlass})
+        end
```
myronmarston: We'll need a passing example for the negated case, too.  As you've spec'd it, the implementation could simply always fail in the negated case and all specs would pass.

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        it "provides a description" do
+          matcher = send(method, SuperKlass)
+          matcher.matches?(OtherKlass)
+          expect(matcher.description).to eq "be a subclass of SuperKlass"
+        end
+      end
+
+      RSpec.describe "expect(actual).not_to #{method}(expected)" do
+        it "fails with failure message for should_not if actual is subclass of expected class" do
```
myronmarston: This doc string shouldn't mention `should_not`.  Really, you don't need to mention the failure message in the doc string here.

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        it "fails with failure message for should unless actual is subclass of expected class" do
+          expect {
+            expect(OtherKlass).to send(method, SuperKlass)
+          }.to fail_with(%Q{expected OtherKlass to be a subclass of SuperKlass})
+        end
+
+        it "provides a description" do
+          matcher = send(method, SuperKlass)
+          matcher.matches?(OtherKlass)
```
myronmarston: I think we've done this elsewhere but it's not necessary to match before calling `description` so I'd prefer to see this line removed.

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        it_behaves_like "an RSpec matcher", :valid_value => SubKlass, :invalid_value => OtherKlass do
+          let(:matcher) { send(method, SuperKlass) }
+        end
+
+        it "passes if actual is subclass of expected class" do
+          expect(SubKlass).to send(method, SuperKlass)
+        end
+
+        it "passes if actual is subklass of subclass of expected class" do
```
myronmarston: s/subklass/subclass/

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect(SubSubKlass).to send(method, SuperKlass)
+        end
+
+        it "fails with failure message for should unless actual is subclass of expected class" do
```
myronmarston: The "for should" part feels out of place since that's not the main syntax anymore.  In fact, I wouldn't even mention the failure message in the doc string here (but keep it in the expectation below).

spec/rspec/matchers/built_in/be_a_subclass_of_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect {
+            expect(OtherKlass).to send(method, SuperKlass)
+          }.to fail_with(%Q{expected OtherKlass to be a subclass of SuperKlass})
+        end
```
myronmarston: Whatever we decide about mixins should be spec'd here as well.

lib/rspec/matchers.rb
```diff
@@ -352,6 +352,19 @@ def be_a_kind_of(expected)
...
+    def be_a_subclass_of(expected)
+      BuiltIn::BeASubclassOf.new(expected)
+    end
+    alias_method :be_subclass_of, :be_a_subclass_of
+    alias_matcher :a_subclass_of,  :be_a_subclass_of
```
myronmarston: All the matcher aliases are spec'd in `spec/rspec/matchers/aliases_spec.rb` so please add one for this as well.

lib/rspec/matchers/built_in.rb
thomasfedb: I was attempting to match the implementation of `be_a_kind_of` (primary method), which is backed by a matcher class called `BeAKindOf` in a file named `be_kind_of.rb`.

I will happily change to make the method and file name the same. Should the same be done for `BeAKindOf` (currently in `be_kind_of.rb`) and `BeAnInstanceOf` (currently in `be_instance_of.rb`) also?

Note that, as far as I can tell, the matcher class names contain the `A` and `An` for the purpose of automatically generating the matcher descriptions rather than anything else.

thomasfedb:
Thanks for your diff comments @myronmarston, I'll wait to see what others think before resolving them.

I do agree that there are many cases, eg: `class User < ActiveRecord::Base` where there is no need to be including `expect(User).to be_a_subclass_of(ActiveRecord::Base)` in specifications.

There are however, cases where an class's hierarchy is a product of less declarative, procedural code, and these are valid situations to check. I submitted this against rspec-expectations itself, less because I think this matcher will be used in every single codebase, but more because it is a basic matcher that probably fits in here best.

thomasfedb:
@myronmarston Do you know if any other contributors have an opinion on this?

myronmarston:
Nope, they would comment here with their opinion.  People have been busy as of late.

@rspec/rspec -- we'd love some of your opinions about this :).

soulcutter:
I'm not convinced this is general-enough purpose.

samphippen:
I'm :-1: on adding this as a core feature. I can't think of a case where I've ever wanted to check that something is a subclass of something else.

myronmarston:
Seems like there's a general consensus that this doesn't belong in rspec given the reasons outlined above and how easy it is for users to add to the few projects that would need it.  I'm going to close this.  Thanks for suggesting it, though, @thomasfedb -- RSpec is moved forward by contributions like these and I hope you'll contribute again.

Issue #712, opened 2015-01-08
Compound matcher diff problem


myronmarston:
Consider this spec:

``` ruby
RSpec::Matchers.define_negated_matcher :exclude, :include

RSpec.describe "Combining include and exclude" do
  let(:multiline_string) do
    <<-EOS
    Foo 1
    Bar 2
    Bazz 3
    EOS
  end

  example "include().and exclude()" do
    expect(multiline_string).to include("    Bar 3").and exclude("    Bar 4")
  end

  example "separate expectations" do
    expect(multiline_string).to include("    Bar 3")
    expect(multiline_string).to exclude("    Bar 4")
  end
end
```

Both specs fail (as is intentional) but notice the difference in the failures:

```
Failures:

  1) Combining include and exclude include().and exclude()
     Failure/Error: expect(multiline_string).to include("    Bar 3").and exclude("    Bar 4")
       expected "    Foo 1\n    Bar 2\n    Bazz 3\n" to include "    Bar 3"
     # ./spec/compound_diff_problem_spec.rb:13:in `block (2 levels) in <top (required)>'

  2) Combining include and exclude separate expectations
     Failure/Error: expect(multiline_string).to include("    Bar 3")
       expected "    Foo 1\n    Bar 2\n    Bazz 3\n" to include "    Bar 3"
       Diff:
       @@ -1,2 +1,4 @@
       -    Bar 3
       +    Foo 1
       +    Bar 2
       +    Bazz 3

     # ./spec/compound_diff_problem_spec.rb:17:in `block (2 levels) in <top (required)>'

Finished in 0.00369 seconds (files took 0.15202 seconds to load)
2 examples, 2 failures
```

You lose the diff when you make it a compound matcher expression.  This is unfortunate, as the diff is so useful.  We should find a way to preserve the diff.  Unfortunately, the diff is not produced by the matcher itself but instead by rspec-expectations by comparing `matcher.expected` to `matcher.actual`...but with a compound matcher there are multiple `expected` values (one per matcher in the compound expression) and there's not a simple way to diff them.  We really need it to diff each matcher individually, and then include that in the failure output...but diffing isn't a responsibility of matchers; matchers simply say they are diffable and expose expected/actual, and I don't particularly want to make diffing a responsibility of the matchers.

Anyone got an idea for how best to solve this?

/cc @rspec/rspec

waterlink:
[...]

PR #713, opened 2015-01-13
Compound matcher diff

waterlink:
related to #712 

This is work in progress but some input required.

So I tried first to approach it by using special return value from `#expected` - an instance of class `ExpectedForMultipleDiffs`, but it became a mess very fast.
Then I tried to actually return more general `Expected` value and shifted responsibility of generating diffs to it.
But then it puzzled me, all the logic about composing `failure_message`-s of matchers is hidden inside of `CompoundMatcher`'s private methods, particularly in method `#failure_message`, and since diffs need to be injected inside of this `failure_message` at multiple places, I thought about actually having a special value for `failure_message` - `FailureMessage`

So the final idea is:
- have a class `LazyFailureMessage` and shift responsibility of generating diffs to it
- `lift` (or `coerce`) all `failure_message` values to it and give it a differ factory and let it return proper `failure_message` with diffs inside
- allow some matchers (`Compound` and `All`) to make evaluation of `failure_message` lazy, allowing it to wait for differ to pass it around to child matchers

You can look at the commits, they currently non-squashed and show failed attempts to tackle it.

Basically this variant of design works surprisingly well, but still it could be hard to understand, so I need some input and/or ideas. I thought, for example, about actually shifting responsibility of generating of `failure_message` for compound matchers somewhere outside of it - it could work out well too.

If this kind of design is ok to go, then I will just clean it up a bit and put some docs strings there.

/cc @myronmarston @JonRowe

### DIFF
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -24,8 +24,7 @@ module RSpec
                                "appropriate failure_message[_when_negated] method to return a string?"
         end
 
-        diff = differ.diff(actual, expected)
-        message = "#{message}\nDiff:#{diff}" unless diff.empty?
+        message = ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(expected).message_with_diff(message, method(:differ), actual)
 
         raise RSpec::Expectations::ExpectationNotMetError, message
       end
--- a/lib/rspec/matchers/built_in/compound.rb
+++ b/lib/rspec/matchers/built_in/compound.rb
@@ -3,6 +3,7 @@ module RSpec
     module BuiltIn
       # @api private
       # Base class for `and` and `or` compound matchers.
+      # rubocop:disable ClassLength
       class Compound < BaseMatcher
         # @private
         attr_reader :matcher_1, :matcher_2, :evaluator
--- a/lib/rspec/matchers/built_in/compound.rb
+++ b/lib/rspec/matchers/built_in/compound.rb
@@ -5,7 +5,7 @@ module RSpec
       # Base class for `and` and `or` compound matchers.
       class Compound < BaseMatcher
         # @private
-        attr_reader :matcher_1, :matcher_2
+        attr_reader :matcher_1, :matcher_2, :evaluator
 
         def initialize(matcher_1, matcher_2)
           @matcher_1 = matcher_1
@@ -34,6 +34,28 @@ module RSpec
         def expects_call_stack_jump?
           NestedEvaluator.matcher_expects_call_stack_jump?(matcher_1) ||
           NestedEvaluator.matcher_expects_call_stack_jump?(matcher_2)
+        end
+
+        # @api private
+        # @return [Boolean]
+        def diffable?
+          matcher_is_diffable?(matcher_1) || matcher_is_diffable?(matcher_2)
+        end
+
+        # @api private
+        # @return [RSpec::Matchers::ExpectedsForMultipleDiffs]
+        def expected
+          return nil unless evaluator
+          ::RSpec::Matchers::ExpectedsForMultipleDiffs.for_many_matchers(diffable_matcher_list)
+        end
+
+      protected
+
+        def diffable_matcher_list
+          list = []
+          list.concat(diffable_matcher_list_for(matcher_1)) unless matcher_1_matches?
+          list.concat(diffable_matcher_list_for(matcher_2)) unless matcher_2_matches?
+          list
         end
 
       private
@@ -88,17 +110,29 @@ module RSpec
         end
 
         def matcher_1_matches?
-          @evaluator.matcher_matches?(matcher_1)
+          evaluator.matcher_matches?(matcher_1)
         end
 
         def matcher_2_matches?
-          @evaluator.matcher_matches?(matcher_2)
+          evaluator.matcher_matches?(matcher_2)
         end
 
         def matcher_supports_block_expectations?(matcher)
           matcher.supports_block_expectations?
         rescue NoMethodError
           false
+        end
+
+        def matcher_is_diffable?(matcher)
+          matcher.diffable?
+        rescue NoMethodError
+          false
+        end
+
+        def diffable_matcher_list_for(matcher)
+          return [] unless matcher_is_diffable?(matcher)
+          return matcher.diffable_matcher_list if Compound === matcher
+          [matcher]
         end
 
         # For value expectations, we can evaluate the matchers sequentially.
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -1,0 +1,79 @@
+module RSpec
+  module Matchers
+    # @api private
+    # Handles list of expected values when there is a need to render
+    # multiple diffs. Also can handle one value.
+    class ExpectedsForMultipleDiffs
+      # @private
+      # Default diff label when there is only one matcher in diff
+      # output
+      DEFAULT_DIFF_LABEL = "Diff:".freeze
+
+      # @private
+      # Maximum readable matcher description length
+      DESCRIPTION_MAX_LENGTH = 65
+
+      def initialize(expected_list)
+        @expected_list = expected_list
+      end
+
+      # @api private
+      # Wraps provided expected value in instance of
+      # ExpectedForMultipleDiffs. If provided value is already an
+      # ExpectedForMultipleDiffs then it just returns it.
+      # @param [Any] expected value to be wrapped
+      # @return [RSpec::Matchers::ExpectedsForMultipleDiffs]
+      def self.from(expected)
+        return expected if self === expected
+        new([[expected, DEFAULT_DIFF_LABEL]])
+      end
+
+      # @api private
+      # Wraps provided matcher list in instance of
+      # ExpectedForMultipleDiffs.
+      # @param [Array<Any>] matchers list of matchers to wrap
+      # @return [RSpec::Matchers::ExpectedsForMultipleDiffs]
+      def self.for_many_matchers(matchers)
+        new(matchers.map { |m| [m.expected, diff_label_for(m)] })
+      end
+
+      # @api private
+      # Returns message with diff(s) appended for provided differ
+      # factory and actual value if there are any
+      # @param [String] message original failure message
+      # @param [Proc] differ_factory brand new differ factory method (or proc, block)
+      # @param [Any] actual value
+      # @return [String]
+      def message_with_diff(message, differ_factory, actual)
+        diff = diffs(differ_factory, actual)
+        message = "#{message}\n#{diff}" unless diff.empty?
+        message
+      end
+
+    private
+
+      def self.diff_label_for(matcher)
+        "Diff for (#{truncated(description_for(matcher))}):"
+      end
+
+      def self.description_for(matcher)
+        matcher.description
+      rescue NoMethodError
+        matcher.inspect
+      end
+
+      def self.truncated(description)
+        return description if description.length <= DESCRIPTION_MAX_LENGTH
+        description[0...DESCRIPTION_MAX_LENGTH - 3] << "..."
+      end
+
+      def diffs(differ_factory, actual)
+        @expected_list.map do |(expected, diff_label)|
+          diff = differ_factory.call.diff(actual, expected)
+          next if diff.empty?
+          "#{diff_label}#{diff}"
+        end.compact.join("\n")
+      end
+    end
+  end
+end
--- a/spec/rspec/matchers/built_in/compound_spec.rb
+++ b/spec/rspec/matchers/built_in/compound_spec.rb
@@ -1,6 +1,13 @@
 module RSpec::Matchers::BuiltIn
   RSpec.describe Compound do
 
+    let(:matcher_without_diffable) { include("foo") }
+
+    before do
+      allow(RSpec::Matchers.configuration).to receive_messages(:color? => false)
+      allow(matcher_without_diffable).to receive(:diffable?).and_raise(NoMethodError)
+    end
+
     shared_examples "making a copy" do |compound_method, copy_method|
       context "when making a copy via `#{copy_method}`" do
         it "uses a copy of the base matchers" do
@@ -375,6 +382,171 @@ module RSpec::Matchers::BuiltIn
             EOS
           end
         end
+
+        context "when the first matcher is diffable" do
+          subject { include("foo").and be_a(String) }
+
+          it 'is diffable' do
+            expect(subject).to be_diffable
+          end
+
+          context "when only first matcher fails" do
+            it 'fails with a message containing a diff for first matcher' do
+              expected_failure = dedent(<<-EOS)
+                |Diff for (include "foo"):
+                |@@ -1,2 +1,3 @@
+                |-foo
+                |+baz
+                |+bar
+              EOS
+
+              expect {
+                expect(dedent(<<-EOS)).to subject
+                  |baz
+                  |bar
+                EOS
+              }.to fail_matching(expected_failure)
+            end
+          end
+
+          context "when only second matcher fails" do
+            subject { include("baz").and be_a(Fixnum) }
+
+            it 'fails with a message not containing a diff for first matcher' do
+              expect {
+                expect(dedent(<<-EOS)).to subject
+                  |baz
+                  |bar
+                EOS
+              }.to fail_matching(a_string_excluding "Diff")
+            end
+          end
+
+          context "when both matcher fail" do
+            subject { include("foo").and eq(35) }
+
+            it "fails with a message containing a diff with first matcher" do
+              expected_failure = dedent(<<-EOS)
+                |   expected "baz\\nbar" to include "foo"
+                |
+                |...and:
+                |
+                |   expected: 35
+                |        got: "baz\\nbar"
+                |
+                |   (compared using ==)
+                |
+                |Diff for (include "foo"):
+                |@@ -1,2 +1,3 @@
+                |-foo
+                |+baz
+                |+bar
+              EOS
+
+              expect {
+                expect(dedent(<<-EOS)).to subject
+                  |baz
+                  |bar
+                EOS
+              }.to fail_matching(expected_failure)
+            end
+          end
+        end
+
+        context "when the first matcher does not implement #diffable?" do
+          subject { matcher_without_diffable.and exist }
+
+          it 'is not diffable' do
+            expect(subject).not_to be_diffable
+          end
+        end
+
+        context "when the second matcher does not implement #diffable?" do
+          subject { exist.and matcher_without_diffable }
+
+          it 'is not diffable' do
+            expect(subject).not_to be_diffable
+          end
+        end
+
+        context "when the second matcher is diffable" do
+          subject { eq(35).and include("foo") }
+
+          it 'is diffable' do
+            expect(subject).to be_diffable
+          end
+
+          it 'fails with a message containing a diff for second matcher' do
+            expected_failure = dedent(<<-EOS)
+              |   expected: 35
+              |        got: "baz\\nbar"
+              |
+              |   (compared using ==)
+              |
+              |...and:
+              |
+              |   expected "baz\\nbar" to include "foo"
+              |Diff for (include "foo"):
+              |@@ -1,2 +1,3 @@
+              |-foo
+              |+baz
+              |+bar
+            EOS
+
+            expect {
+              expect(dedent(<<-EOS)).to subject
+                |baz
+                |bar
+              EOS
+            }.to fail_matching(expected_failure)
+          end
+        end
+
+        context "when both matchers are diffable" do
+          subject { include("bar").and include("foo") }
+
+          it 'is diffable' do
+            expect(subject).to be_diffable
+          end
+
+          it 'fails with a message containing diffs for both matcher' do
+            expected_failure = dedent(<<-EOS)
+              |expected "baz\\nbug" to include "bar" and expected "baz\\nbug" to include "foo"
+              |Diff for (include "bar"):
+              |@@ -1,2 +1,3 @@
+              |-bar
+              |+baz
+              |+bug
+              |
+              |Diff for (include "foo"):
+              |@@ -1,2 +1,3 @@
+              |-foo
+              |+baz
+              |+bug
+            EOS
+
+            expect {
+              expect(dedent(<<-EOS)).to subject
+                |baz
+                |bug
+              EOS
+            }.to fail_matching(expected_failure)
+          end
+        end
+
+        context "when both matchers are not diffable" do
+          subject { be_a(String).and be_truthy }
+
+          it 'is not diffable' do
+            expect(subject).not_to be_diffable
+          end
+
+          it 'fails with a message not containing any diff' do
+            expect {
+              expect(35).to subject
+            }.to fail_matching(a_string_excluding "Diff")
+          end
+        end
       end
     end
 
@@ -489,6 +661,119 @@ module RSpec::Matchers::BuiltIn
           end
         end
       end
+
+      context "when first matcher is diffable" do
+        subject { include("foo").or eq(35) }
+
+        it "is diffable" do
+          expect(subject).to be_diffable
+        end
+
+        it 'fails with a message containing diff for first matcher' do
+          expected_failure = dedent(<<-EOS)
+            |   expected "baz\\nbug" to include "foo"
+            |
+            |...or:
+            |
+            |   expected: 35
+            |        got: "baz\\nbug"
+            |
+            |   (compared using ==)
+            |
+            |Diff for (include "foo"):
+            |@@ -1,2 +1,3 @@
+            |-foo
+            |+baz
+            |+bug
+          EOS
+
+          expect {
+            expect(dedent(<<-EOS)).to subject
+              |baz
+              |bug
+            EOS
+          }.to fail_matching(expected_failure)
+        end
+      end
+
+      context "when second matcher is diffable" do
+        subject { eq(35).or include("foo") }
+
+        it "is diffable" do
+          expect(subject).to be_diffable
+        end
+
+        it 'fails with a message containing diff for second matcher' do
+          expected_failure = dedent(<<-EOS)
+            |   expected: 35
+            |        got: "baz\\nbug"
+            |
+            |   (compared using ==)
+            |
+            |...or:
+            |
+            |   expected "baz\\nbug" to include "foo"
+            |Diff for (include "foo"):
+            |@@ -1,2 +1,3 @@
+            |-foo
+            |+baz
+            |+bug
+          EOS
+
+          expect {
+            expect(dedent(<<-EOS)).to subject
+              |baz
+              |bug
+            EOS
+          }.to fail_matching(expected_failure)
+        end
+      end
+
+      context "when both matchers are diffable" do
+        subject { include("foo").or include("buzz") }
+
+        it "is diffable" do
+          expect(subject).to be_diffable
+        end
+
+        it 'fails with a message containing diffs for both matcher' do
+          expected_failure = dedent(<<-EOS)
+            |expected "baz\\nbug" to include "foo" or expected "baz\\nbug" to include "buzz"
+            |Diff for (include "foo"):
+            |@@ -1,2 +1,3 @@
+            |-foo
+            |+baz
+            |+bug
+            |
+            |Diff for (include "buzz"):
+            |@@ -1,2 +1,3 @@
+            |-buzz
+            |+baz
+            |+bug
+          EOS
+
+          expect {
+            expect(dedent(<<-EOS)).to subject
+              |baz
+              |bug
+            EOS
+          }.to fail_matching(expected_failure)
+        end
+      end
+
+      context "when both matchers are not diffable" do
+        subject { be_a(String).or be_a(Fixnum) }
+
+        it "is not diffable" do
+          expect(subject).not_to be_diffable
+        end
+
+        it 'fails with a message containing diffs for both matcher' do
+          expect {
+            expect(true).to subject
+          }.to fail_matching(a_string_excluding "Diff")
+        end
+      end
     end
 
     context "when chaining many matchers together" do
@@ -496,6 +781,38 @@ module RSpec::Matchers::BuiltIn
         matcher = start_with("f").and end_with("z").or end_with("o")
         expect("foo").to matcher
         expect(matcher.description).to eq('start with "f" and end with "z" or end with "o"')
+      end
+
+      it 'fails with complete diffs if its matchers are diffable' do
+        matcher = include("bar").and include("buzz").or include("foo")
+
+        expected_failure = dedent(<<-EOS)
+          |expected "bug\\nsquash" to include "bar" and expected "bug\\nsquash" to include "buzz" or expected "bug\\nsquash" to include "foo"
+          |Diff for (include "bar"):
+          |@@ -1,2 +1,3 @@
+          |-bar
+          |+bug
+          |+squash
+          |
+          |Diff for (include "buzz"):
+          |@@ -1,2 +1,3 @@
+          |-buzz
+          |+bug
+          |+squash
+          |
+          |Diff for (include "foo"):
+          |@@ -1,2 +1,3 @@
+          |-foo
+          |+bug
+          |+squash
+        EOS
+
+        expect {
+          expect(dedent(<<-EOS)).to matcher
+            |bug
+            |squash
+          EOS
+        }.to fail_matching(expected_failure)
       end
 
       it 'fails with a complete message' do
--- a/spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
+++ b/spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
@@ -1,0 +1,116 @@
+module RSpec
+  module Matchers
+    RSpec.describe ExpectedsForMultipleDiffs do
+
+      before do
+        stub_const("::RSpec::Matchers::ExpectedsForMultipleDiffs::DESCRIPTION_MAX_LENGTH", 30)
+      end
+
+      class FakeDiffer
+        FACTORY = Proc.new { FakeDiffer }
+
+        def self.diff(actual, expected)
+          [actual, expected].inspect
+        end
+      end
+
+      let(:differ_factory) { FakeDiffer::FACTORY }
+
+      let(:message) { "a message" }
+      let(:actual) { "actual value" }
+
+      let(:wrapped_value) { described_class.from("expected value") }
+
+      let(:matcher_1) { instance_double(BuiltIn::BaseMatcher, :description => "matcher 1 description", :expected => "expected 1") }
+      let(:matcher_2) { instance_double(BuiltIn::BaseMatcher, :description => "matcher 2 description", :expected => "expected 2") }
+      let(:matcher_3) { instance_double(BuiltIn::BaseMatcher, :description => "matcher 3 description", :expected => "expected 3") }
+
+      let(:long_description) { "a very very long description for my custom smart matcher, which can be used for everything" }
+      let(:truncated_description) { "a very very long descriptio..." }
+      let(:matcher_with_long_description) { instance_double(BuiltIn::BaseMatcher, :description => long_description, :expected => "expected value") }
+
+      let(:matcher_without_description_defined) { double("custom matcher", :expected => "expected value", :inspect => "#<CustomMatcher:0xf0c8561a55>") }
+
+      before { allow(matcher_without_description_defined).to receive(:description).and_raise(NoMethodError) }
+
+      describe ".from" do
+        it "wraps provided value in ExpectedsForMultipleDiffs" do
+          expect(wrapped_value).to be_a(described_class)
+        end
+
+        it "returns original value if it was already wrapped" do
+          expect(described_class.from(wrapped_value)).to be(wrapped_value)
+        end
+      end
+
+      describe ".for_many_matchers" do
+        let(:wrapped_value) { described_class.for_many_matchers([matcher_1, matcher_2, matcher_3]) }
+
+        it "has a diff for all matchers with their description" do
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
+            |Diff for (matcher 1 description):["actual value", "expected 1"]
+            |Diff for (matcher 2 description):["actual value", "expected 2"]
+            |Diff for (matcher 3 description):["actual value", "expected 3"]
+          EOS
+        end
+      end
+
+      describe "#message_with_diff" do
+        it "returns just provided message if diff is empty" do
+          allow(FakeDiffer).to receive(:diff) { "" }
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
+          EOS
+        end
+
+        it "returns regular message with diff when single expected" do
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
+            |Diff:["actual value", "expected value"]
+          EOS
+        end
+
+        it "returns message with diff and matcher description when single expected with matcher" do
+          wrapped_value = described_class.for_many_matchers([include("expected value")])
+
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
+            |Diff for (include "expected value"):["actual value", ["expected value"]]
+          EOS
+        end
+
+        it "returns message with diff and truncated matcher description if it is too long" do
+          wrapped_value = described_class.for_many_matchers([matcher_with_long_description])
+
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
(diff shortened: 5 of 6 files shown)


waterlink:
@myronmarston @JonRowe can you look into this one? Is design I described in PR description appropriate for you?

myronmarston:
> @myronmarston @JonRowe can you look into this one? Is design I described in PR description appropriate for you?

Sorry, the github notification for this go buried in a bunch of other notifications and I forgot about it.  I'll take a look today and leave feedback.

myronmarston:
Thanks for taking a stab at this, @waterlink.  Here's an assortment of my thoughts...


We need to decide if the diffs is going to be interleaved within the compound failure message (as you've done here) or come afterwords (as it normally is).  Which direction we go will dictate which solution we use.  I can see arguments both ways:

* Interleaving the diff in the failure message puts the diff right after the failure message it relates to.  This makes it easier to understand what it relates to.  However, doing this forces certain design choices on us as we can't keep the failure message and diff strictly separated as they are now.  I suspect a large part of the complexity you're dealing with here has to do with that.  There's also another issue: while the interleaved diff flows well in the compound multi-line failure messages, not all compound failures have multiple lines.  For example, see [this spec](https://github.com/rspec/rspec-expectations/blob/a88281bb96439d7e485b36f9c686a4ad9662c203/spec/rspec/matchers/built_in/compound_spec.rb#L336-L340).  You can't really put diffs in the middle of that failure message. Although...I suppose you could just convert single-line failures into the current multi-line form with the indentation and the `<conjunction>....` bit.  I'll need to think more about if we want that.
* Putting the diffs at the end feels _much_ simpler to me.  I suspect that the `ExpectedsForMultipleDiffs` solution (which we discussed in #712 as what I would conceptually prefer) would become more feasible if we put diffs at the end. It would also keep the strict separation between failure message and diffs which I conceptually like -- it feels like a better division of responsibilities.  Finally, it would allow single-line compound failure messages to remain as a single line.  The big downside to putting diffs at the end is that it's less obvious to the user which matcher it relates to.  In practice, I doubt that'll be much of an issue as it should be pretty obvious from the diff and the order.  Also, we could perhaps do something like `diff (for <matcher description>):` so that the diffs are labeled.

Overall, I think I lean towards keeping the diffs separate at the end rather than interleaved, since it will probably be a much simpler implementation, and as maintainer, that matters to me :).

As for your current implementation -- I find it quite confusing, honestly.  It seems very haskell-ish -- do you do much haskell?  I admit I haven't done much and really only started to understand monads a few months ago.  I did take the recent [edx.org FP101 class](https://courses.edx.org/courses/DelftX/FP101x/3T2014/info) and quite enjoyed learning haskell there.  

Regardless, I'm struggling to wrap my mind around the solution here, and it feels non-idiomatic in ruby.  One part in particular that's confusing me is `LazyFailureMessage#bind`.  There are 3 different blocks/procs at work here -- `value` (which I feel is misnamed -- it's a proc that does something, not a static value), the block you are passing to `value.call` and `blk`.  I don't understand what these 3 procs represent so it's hard to understand the code.  I think `LazyFailureMessage#lift` would be more appropriately named `#for` or `#from` as that feels more idiomatic to me and will be obvious that it returns a `LazyFailureMessage` instance for the provided argument.  `lift` is too haskell-ish in ruby code for my tastes.

I might be more amenable to the interleaved diff solution if the code for it was more easily understandable but right now that's how I lean.  I am curious what the other RSpec folks think about the interleaving vs at-the-end question: @rspec/rspec ?

waterlink:
Thanks for your feedback @myronmarston.

About naming, you are right it is more or less inspired by monads. You may think about `#lift` as `#coerce` or `#from` in term of ruby, `value` will be something like `action`, `#bind` => `#call`, `#escape` => `#fetch`.

If it is appropriate to show all diffs in the end of the failure message, then I would go for `ExpectedsForMultipleDiffs` it would be really pretty easy to implement this way.

About multiline and singleline diffs: in the current implementation, in the end of the day, having a diffable matcher on either side of compound matcher, makes failure_message of this matcher automatically multiline and as a result compound matcher becomes multiline. By the way current implementation handles the case you linked [example output](https://github.com/waterlink/rspec-compound-matcher-diff-example/blob/master/output.txt#L84-L88)

And in the end, if @rspec/rspec prefers to just put diffs after failure message for compound matchers, I would actually vote for that too, because it makes the implementation way way easier

waterlink:
Nobody is interested in this topic seems.. Then I will just implement simpler approach and update the PR

myronmarston:
I'm interested...br

myronmarston:
> Nobody is interested in this topic seems.. Then I will just implement simpler approach and update the PR

Sounds good.

alex-fedorov:
@myronmarston no offense, I mean nobody but you :)

xaviershay:
I agree with myron's assessment.

waterlink:
@myronmarston I updated PR, can you take a look?

waterlink:
BTW, I get this strange stuff locally on ruby 2.2 https://gist.github.com/waterlink/48c60f8bf129553bdece : warning inside of rubygems..

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -36,6 +36,30 @@ def expects_call_stack_jump?
...
+
+          expecteds = ::RSpec::Matchers::ExpectedsForMultipleDiffs.new
+
+          if !matcher_1_matches? && matcher_1.diffable?
+            expecteds.add(matcher_1.expected, :for_matcher => matcher_1)
+          end
+
+          if !matcher_2_matches? && matcher_2.diffable?
+            expecteds.add(matcher_2.expected, :for_matcher => matcher_2)
+          end
```
myronmarston: Personally, I prefer immutable objects when we can find a way to make that work, since it's easier to reason about.  Can we make `RSpec::Matchers::ExpectedsForMultipleDiffs` immutable?

myronmarston: Also, do you think it would make sense to pass the matcher description to `expecteds` rather than the matcher?  All it needs is the description, and passing a string feels like lower coupling to me.  If there's a concrete reason you think it's advantageous to pass the matcher itself I'm open to keeping it as is.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+          new(matcher)
+        end
+
+        # @api private
+        # Returns proper diff headline for wrapper matcher with
+        # matcher description included
+        # @return [String]
+        def diff_headline
+          "Diff for (#{matcher.description}):"
+        end
```
myronmarston: `matcher.description` can be arbitrarily long -- for example, `include(some_long_4K_length_string).description` will include all 4K characters of the string.  If it's a long string it's going to obscure the diff.  What do you think about truncating the description so that it merely includes the first N characters of the string (with an appropriate `...` if truncated)?  N could be something like 65 or 70 -- whatever ensures the `Diff` line won't be longer than 80 characters.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+      # Null object for ForMatcher, used in case for_matcher was not
+      # provided
+      class NoMatcher < ForMatcher
+        # @api private
+        # Returns default diff headline with no matcher description
+        # @return [String]
+        def diff_headline
+          "Diff:"
+        end
+      end
```
myronmarston: `NoMatcher` doesn't really need to inherit from `ForMatcher`, right?  Seems like unnecessary inheritance.  In fact, `NoMatcher` is stateless, so IMO we'd be better off with:

``` ruby
module NoMatcher
  def self.diff_headline
    "Diff:"
  end
end
```

...and then changing `NoMatcher.new` to `NoMatcher` above.  As a singleton, stateless object, we'll have fewer object allocations which translates into less GC, etc.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+      def diffs(differ_factory, actual)
+        expected_list.map do |expected|
+          expected.diffs(differ_factory, actual)
+        end.inject([], :+).join("\n")
+      end
+
+      def expected_from(value, for_matcher)
+        return value.expected_list if self.class === value
+        [Expected.new(value, for_matcher)]
+      end
```
myronmarston: Combining this definition with the call site, I see that this is happening:

``` ruby
wrapped_item = [item]
item_list.concat wrapped_item
```

...when we could just do:

``` ruby
item_list << item
```

...which, besides being less code, skips an unnecessary array allocation.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+    private
+
+      def diffs(differ_factory, actual)
+        expected_list.map do |expected|
+          expected.diffs(differ_factory, actual)
+        end.inject([], :+).join("\n")
+      end
+
+      def expected_from(value, for_matcher)
+        return value.expected_list if self.class === value
```
myronmarston: Having methods coerce their input is handy so you don't have to worry about what the argument is (or if an argument has already been coerced), but it feels like defensive programming to have this in a private method.  Given the only callers are in this class, won't it already know what the object is?  It feels like this conversion is here "just in case" and I'd prefer to see the conversions only done where they are truly needed (hopefully in one or maybe two spots?)

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+      def expected_list
+        @_expected_list ||= []
+      end
+
+    private
+
+      def diffs(differ_factory, actual)
+        expected_list.map do |expected|
+          expected.diffs(differ_factory, actual)
+        end.inject([], :+).join("\n")
```
myronmarston: This method iterates over the list twice -- once for `map`, once for `inject`.  I suspect it will perform better to iterate only once.  Can you use `flat_map` instead and only iterate once?

Note that 1.8.7 lacks `Enumerable#flat_map`; in rspec-core we defined our own wrapper that deals with that:

https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/flat_map.rb

(See also rspec/rspec-core#1854 -- there's a mistake in that file). While you can't directly use rspec-core from rspec-expectations for that, you could do something like:

``` ruby
module RSpec::Expectations
  if defined?(RSpec::Core::FlatMap
    FlatMap = RSpec::Core::FlatMap
  else
    # define it inline using an identical definition to what rspec-core has
  end
end
```

It's a bit non-DRY but I'm not really concerned since `flat_map` isn't a piece of knowledge that's likely to evolve.

BTW, we do have benchmarks in rspec-core showing that `flat_map` is faster than `inject` -- see rspec/rspec-core#1836.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,118 @@
...
+        # @param [Any] actual value
+        # @return [String]
+        def diffs(differ_factory, actual)
+          diff = differ_factory.call.diff(actual, value)
+          return [] if diff.empty?
+          ["#{for_matcher.diff_headline}#{diff}"]
+        end
+      end
+
+      ForMatcher = Struct.new(:matcher)
```
myronmarston: `ForMatcher` instances aren't really value object.  Using `Struct` is handy when you want the `==`/`eql?`/`hash` semantics it provides for value objects, but here they are unused, so I'd prefer to just use a simple class def.

All that said...as much as I like defining a polymorphic interface and multiple small implementations, as you've done here...it feels like a lot of code and ceremony for something that could be much simpler.  For example, `ExpectedsForMultipleDiffs` could have two factory methods:

``` ruby
class ExpectedsForMultipleDiffs
  DEFAULT_DIFF_LABEL = "Diff:".freeze

  def self.from(expected)
    return expected if self === expected
    new([expected, DEFAULT_DIFF_LABEL)
  end

  def self.for_many_matchers(matchers)
    new(matchers.map { |m| [m.expected, diff_label_for(m)] })
  end
end
```

A few things to note about this:

* I intend `for_many_matchers` be the factory method used from `CompoundMatcher`.  Notice that this makes the class immutable, and simplifies the calling interface -- rather than passing `expected` and `matcher` separately (and/or the description string), it just passes the matcher, and let's `for_many_matchers` worry about extracting `expected` and the diff label from it.
* No need for the polymorphic `ForMatcher` and `NoMatcher` classes -- the difference is immediately dealt with at the point of instantiation very simply.
* For something as simple as this, I think I prefer keeping the expected value and diff label together as a simple tuple (e.g. an array of 2 values) rather than making the `Expected` class.  The class feels like overkill when it only provides one method, and I suspect that using arrays will probably perform better since ruby has optimized arrays in C so much.  Plus, these sorts of tuples work really elegantly when later iterating over them since you can destructure into the two elements directly in the block args, e.g.:

``` ruby
diffs = expected_list.map do |(expected_value, diff_label)|
  diff_for(expected_value, diff_label)
end.join("\n")
```

Thoughts?

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
myronmarston: So I just realized that the `flat_map` is unnecessary.  You can just use `map` if `Expected#diffs` returned a single string rather than wrapping it an array unnecessarily.  (Or if you go with the refactoring I recommended in https://github.com/rspec/rspec-expectations/pull/713/files#r23470626)

myronmarston:
Thanks, @waterlink, this is looking good.  Some of my earlier comments are obseleted by my later ones ( I comment as I read the diff, and sometimes I understand things better later and that changes what I suggest).  My main suggestion is in this comment:

https://github.com/rspec/rspec-expectations/pull/713#discussion_r23470626

myronmarston:
> BTW, I get this strange stuff locally on ruby 2.2 https://gist.github.com/waterlink/48c60f8bf129553bdece : warning inside of rubygems..

I get that locally, too.  I suspect that travis is running a different version of rubygems.  The fix would be to change https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/spec/prevent_load_time_warnings.rb so that it filters out warnings coming from rubygems.  Actually...an even better option might be to make those specs shell out with `--disable=gems` since I think that will prevent rubygems from being loaded.  If you want to take a stab at fixing that, feel free but it hasn't been a priority since it's only failing locally and doesn't represent a failure we care about.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: Yes, truncating will do good here.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: It will not work this way for:

```ruby
expect(something).to include("foo").and exclude("bar").and include("test")
```

Because this compound matcher represents a tree:

```
Compound :and
  - include "foo"
  - Compound :and
    - exclude "bar"
    - include "test"
```

This `item` at some point will be another instance of `ExpectedsForMultipleDiffs`, that actually contains its own `expecteds_list`, so we need to extract it from there and append at the end of current list.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: I like this simple code, but I'm still concerned about complicated tree of compound matchers, so I will try to achieve something like that, probably will be a bit more complicated (+1 small method or something).

waterlink:
Thanks for your feedback, @myronmarston, I will look into it and will update a PR with much (or even much-much) simpler solution.

waterlink:
@myronmarston I have just update the PR, can you take a look?

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -36,6 +36,28 @@ def expects_call_stack_jump?
...
+        # @api private
+        # @return [RSpec::Matchers::ExpectedsForMultipleDiffs]
+        def expected
+          return nil unless evaluator
+          ::RSpec::Matchers::ExpectedsForMultipleDiffs.for_many_matchers(matcher_list)
+        end
+
+      protected
+
+        def matcher_list
```
myronmarston: I think `matcher_list` is too generic of a name -- outside of the context of diffing/`expected`, I'd expect `compound_matcher.matcher_list` to always return all matchers, not just diffable ones.  WDYT of `matcher_list_for_expected` to make the purpose more explicit?

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -101,6 +123,11 @@ def matcher_supports_block_expectations?(matcher)
           false
         end
 
+        def matcher_list_for(matcher)
```
myronmarston: Same here...it'd be nice to have this make clear that it's just diffable matchers.  Maybe `diffable_matcher_list_from(matcher)`?

lib/rspec/matchers/built_in/compound.rb
myronmarston: Actually, after seeing `matcher_list_for` below, I'm thinking `diffable_matcher_list` is an even better name.

lib/rspec/matchers/built_in/compound.rb
myronmarston: Also, if we do update the name to `diffable_matcher_list_from`, we should probably put `matcher.diffable?` check here to match the method name...which would also reduce duplication a bit.

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -36,6 +36,28 @@ def expects_call_stack_jump?
...
+        def expected
+          return nil unless evaluator
+          ::RSpec::Matchers::ExpectedsForMultipleDiffs.for_many_matchers(matcher_list)
+        end
+
+      protected
+
+        def matcher_list
+          list = []
+          list += matcher_list_for(matcher_1) if !matcher_1_matches? && matcher_1.diffable?
```
myronmarston: `concat` is more efficient than `+=` since it doesn't allocate a new array.

lib/rspec/expectations/fail_with.rb
```diff
@@ -24,8 +24,7 @@ def fail_with(message, expected=nil, actual=nil)
                                "appropriate failure_message[_when_negated] method to return a string?"
         end
 
-        diff = differ.diff(actual, expected)
-        message = "#{message}\nDiff:#{diff}" unless diff.empty?
+        message = ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(expected).message_with_diff(message, method(:differ), actual)
```
myronmarston: I think that `differ` can be passed directly as the 2nd argument -- no need to make it a factory.  It'll be a bit more straight forward and efficient that way.  Honestly, the only reason `differ` isn't memoized is because of cases where the user may re-configure `color` -- in that case, a memoized instance could work in properly.  But here that concern doesn't apply because the user won't have a chance to reconfigure `color` while these objects are in use.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    private
+
+      def self.diff_label_for(matcher)
+        "Diff for (#{matcher.description}):"
+      end
+
+      def diffs(differ_factory, actual)
+        diffs = []
+        @expected_list.each do |(value, diff_label)|
```
myronmarston: I think I'd slightly prefer `value` be `expected` here, to make it explicit what it is...but not a merge blocker.

lib/rspec/matchers/built_in/compound.rb
myronmarston: One other thing about `diffable?`: it's an optional part of the matcher protocol.  That means that some matchers may not implement it.  Thus, we need to check `respond_to?(:diffable?)` before sending that message, like it's done here:

https://github.com/rspec/rspec-expectations/blob/45ff9faa26ea912d6b13d4044e5691b894d9fcc5/lib/rspec/expectations/handler.rb#L37

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +375,165 @@ def expect_block
...
+        context "when the first matcher is diffable" do
+          subject { include("foo").and be_a(String) }
+
+          it 'is diffable' do
+            expect(subject).to be_diffable
+          end
+
+          context "when only first matcher fails" do
+            it 'fails with a message containing a diff for first matcher' do
+              allow(RSpec::Matchers.configuration).to receive_messages(:color? => false)
```
myronmarston: Given this is a super common stub, should we extract it into a `before` hook?

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +375,165 @@ def expect_block
...
+                  |bar
+                EOS
+              }.to fail_matching(a_string_excluding "Diff")
+            end
+          end
+
+          context "when both matcher fail" do
+            subject { include("foo").and eq(35) }
+
+            it 'fails with a message containing a diff with second matcher\'s failure' do
```
myronmarston: * I'd prefer to use double quotes if you are using `'` as an apostrophe so we don't have to escape it.
* This doc string doesn't seem accurate...the diff shown below is for the 1st matcher's failure, not the 2nd matcher.

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      let(:message) { "a message" }
+      let(:actual) { "actual value" }
+
+      let(:wrapped_value) { described_class.from("expected value") }
+
+      let(:matcher_1) { double("Matcher", :description => "matcher 1 description", :expected => "expected 1") }
+      let(:matcher_2) { double("Matcher", :description => "matcher 2 description", :expected => "expected 2") }
+      let(:matcher_3) { double("Matcher", :description => "matcher 3 description", :expected => "expected 3") }
```
myronmarston: I'd prefer to see `instance_double` used (maybe `instance_double(BaseMatcher)`) as that's safer.

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    RSpec.describe ExpectedsForMultipleDiffs do
+
+      # Setup fake differ
+      let(:differ_factory) { double("Differ factory method", :call => differ) }
+      let(:differ) { double("Differ") }
+
+      before do
+        allow(differ).to receive(:diff) do |_actual, expected|
+          [_actual, expected].inspect
+        end
```
myronmarston: Given that this isn't a canned response but instead has logic, do you think it would make more sense to just make a fake differ?

``` ruby
class FakeDiffer
  def self.diff(_actual, expected)
    [_actual, expected].inspect
  end
end
```

Then there's no need to re-stub it for each spec.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -498,6 +778,40 @@ def expect_block
         expect(matcher.description).to eq('start with "f" and end with "z" or end with "o"')
       end
 
+      it 'fails with complete diffs if it matchers are diffable' do
```
myronmarston: Should this be "if it's matchers are diffable"?

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,66 @@
...
+      def message_with_diff(message, differ_factory, actual)
+        diff = diffs(differ_factory, actual)
+        message = "#{message}\n#{diff}" unless diff.empty?
+        message
+      end
+
+    private
+
+      def self.diff_label_for(matcher)
+        "Diff for (#{matcher.description}):"
```
myronmarston: This still needs to be truncated, right?

myronmarston:
Thanks, @waterlink.  This is looking much better.   Nice to see how much simpler you made it!

lib/rspec/matchers/built_in/compound.rb
waterlink: +1 for `diffable_matcher_list`

lib/rspec/matchers/built_in/compound.rb
waterlink: But `BaseMatcher` defines defaul `diffable?` as `false`. I assume user custom matcher does not have to derive from `BaseMatcher`?

lib/rspec/matchers/built_in/compound.rb
waterlink: +1

lib/rspec/expectations/fail_with.rb
waterlink: I think I noticed some problems with that, it seemed like differ wasn't stateless, i.e. the more you use it, more weirder result you get. But could be I wrong, I will try and see if it possible.

waterlink: Yes, now I see. For example when you do a `diff` on the same `differ` twice you get:

```
       Diff for (include "bar"):
       @@ -1,2 +1,3 @@
       -bar
       +baz
       +bug
       
       Diff for (include "foo"):
       @@ -1,2 +1,3 @@
       --bar
       +baz
       +bu
```

waterlink: notice excess `-` (minus) in second diff.

waterlink: Something along these lines causes it to do so: https://github.com/rspec/rspec-support/blob/master/lib/rspec/support/differ.rb#L30

Actually it is pretty hard to reason about this class..

waterlink: I think, we can leave this `differ_factory` untouched, and a bit later I will just take a stab at refactoring `Differ` class.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: +1

spec/rspec/matchers/built_in/compound_spec.rb
waterlink: makes sense

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
waterlink: I actually thought about that after I've written this. Since you have suggested it too, I will go for that.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: Yes, I have totally forgot..

lib/rspec/expectations/fail_with.rb
myronmarston: > I think, we can leave this differ_factory untouched, and a bit later I will just take a stab at refactoring Differ class.

:+1: I didn't realize it had these problems.  It'd be great to fix the statefulness issues later on, as you said.

lib/rspec/matchers/built_in/compound.rb
myronmarston: > But BaseMatcher defines default diffable? as false. I assume user custom matcher does not have to derive from BaseMatcher?

We use `BaseMatcher` internally but have it labeled as private:

https://github.com/rspec/rspec-expectations/blob/45ff9faa26ea912d6b13d4044e5691b894d9fcc5/lib/rspec/matchers/built_in/base_matcher.rb#L9-L13

waterlink:
hmm, ruby 1.9.2 had a segfault suddenly: https://travis-ci.org/rspec/rspec-expectations/jobs/48559100 I will look into it

waterlink:
Everything looks fine locally though. @myronmarston Can you look into this failure, I don't see any connection even to `rspec-expectations` there in stacktraces. Have you seen something before on travis?

myronmarston:
> Have you seen something before on travis?

Yes, we've seen odd segfaults from 1.9.2 from time to time.  Here are some notes from the last time we saw it:

https://github.com/rspec/rspec-core/pull/1683#issuecomment-55854029

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -36,6 +36,28 @@ def expects_call_stack_jump?
           NestedEvaluator.matcher_expects_call_stack_jump?(matcher_2)
         end
 
+        # @api private
+        # @return [Boolean]
+        def diffable?
+          [matcher_1, matcher_2].any?(&method(:matcher_is_diffable?))
```
waterlink: Looks like this can fail for 1.9.2. So I will make it pass explicit block with method call inside..

myronmarston: Honestly, I'd rather it be this:

``` ruby
def diffable?
  matcher_is_diffable?(matcher_1) || matcher_is_diffable?(matcher_2)
end
```

I don't think that's enough duplication to put it in an array and use `any?`, particularly when the `Array#any?` way has an extra object allocation for the array.

waterlink:
Travis looks green now. Looks like I have addressed all CR and Travis problems. @myronmarston can you take another look?

lib/rspec/matchers/built_in/compound.rb
waterlink: Looks good to me too.

waterlink: Done.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+
+    private
+
+      def self.diff_label_for(matcher)
+        "Diff for (#{truncated(matcher.description)}):"
+      end
+
+      def self.truncated(description)
+        return description if description.length < DESCRIPTION_MAX_LENGTH
```
myronmarston: Should this be `<=` ?

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,75 @@
...
+        description[0..DESCRIPTION_MAX_LENGTH] + "..."
+      end
```
myronmarston: This isn't really a big deal (and certainly not a merge blocker), but if the description is too long it will actually wind up longer than `DESCRIPTION_MAX_LENGTH` (due to the addition of `...`).  What's slightly odd is that if a description is 66 or 67 characters, it'll get truncated to 65, and the `...` will get added...which will give us a _longer_ string than we had originally!

Seems like the string slice should stop 3 characters short to allow for `...`?

Also, I think we can use `<<` instead of `+` (one less string allocation).  The sliced string is safe to mutate since nothing else has a reference to it.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +379,155 @@ def expect_block
...
+              |        got: "baz\\nbar"
+              |
+              |   (compared using ==)
+              |
+              |Diff for (include "foo"):
+              |@@ -1,2 +1,3 @@
+              |-foo
+              |+baz
+              |+bar
+            EOS
```
myronmarston: I think this heredoc should be indented 2 spaces to be consistent.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +379,155 @@ def expect_block
...
+
+              expect {
+                expect(dedent(<<-EOS)).to subject
+                |baz
+                |bar
+              EOS
```
myronmarston: Same here.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: At first I chosen `<=`, but if I think about name of constant.. Yeah you're right, should be `<=`.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +379,155 @@ def expect_block
...
+            |   (compared using ==)
+            |
+            |...and:
+            |
+            |   expected "baz\\nbar" to include "foo"
+            |Diff for (include "foo"):
+            |@@ -1,2 +1,3 @@
+            |-foo
+            |+baz
+            |+bar
```
myronmarston: It'd be nice if the `|` lines were indented 2 spaces...

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -375,6 +379,155 @@ def expect_block
...
+            |-bar
+            |+baz
+            |+bug
+            |
+            |Diff for (include "foo"):
+            |@@ -1,2 +1,3 @@
+            |-foo
+            |+baz
+            |+bug
+            EOS
```
myronmarston: Here, too.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -489,6 +642,119 @@ def expect_block
...
+          |        got: "baz\\nbug"
+          |
+          |   (compared using ==)
+          |
+          |Diff for (include "foo"):
+          |@@ -1,2 +1,3 @@
+          |-foo
+          |+baz
+          |+bug
+          EOS
```
myronmarston: And here.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -489,6 +642,119 @@ def expect_block
...
+          |
+          |...or:
+          |
+          |   expected "baz\\nbug" to include "foo"
+          |Diff for (include "foo"):
+          |@@ -1,2 +1,3 @@
+          |-foo
+          |+baz
+          |+bug
+          EOS
```
myronmarston: and here.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -489,6 +642,119 @@ def expect_block
...
+          |-foo
+          |+baz
+          |+bug
+          |
+          |Diff for (include "buzz"):
+          |@@ -1,2 +1,3 @@
+          |-buzz
+          |+baz
+          |+bug
+          EOS
```
myronmarston: And here.

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -498,6 +764,38 @@ def expect_block
...
+        |-buzz
+        |+bug
+        |+squash
+        |
+        |Diff for (include "foo"):
+        |@@ -1,2 +1,3 @@
+        |-foo
+        |+bug
+        |+squash
+        EOS
```
myronmarston: And here.

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+          expect(wrapped_value.message_with_diff(
+            message, differ_factory, actual
+          )).to eq(dedent <<-EOS)
+            |a message
+            |Diff for (include "expected value"):["actual value", ["expected value"]]
+          EOS
+        end
+
+        it "returns message with diff and truncated matcher description if it is too long" do
```
myronmarston: Currently, this spec will fail if we change the value of `DESCRIPTION_MAX_LENGTH`.  However, the correctness of the behavior doesn't really depend on the actual value of that constant; it just depends on it doing the correct thing relative to that constant.  Thus, we can make this spec less brittle by stubbing that constant (potentially to a completely different value) -- then the spec will still demonstrate the behavior relative to that constant but will not care about wha the value of the constant is.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,75 @@
...
+      def message_with_diff(message, differ_factory, actual)
+        diff = diffs(differ_factory, actual)
+        message = "#{message}\n#{diff}" unless diff.empty?
+        message
+      end
+
+    private
+
+      def self.diff_label_for(matcher)
+        "Diff for (#{truncated(matcher.description)}):"
```
myronmarston: I forgot to mention this earlier...`description` is optional (but encouraged), too. See here for the protocol docs:

http://www.rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers/MatcherProtocol

Thus, we need to handle matchers which do not have a description.  Elsewhere we've handled it with a simple rescue:

https://github.com/rspec/rspec-expectations/blob/45ff9faa26ea912d6b13d4044e5691b894d9fcc5/lib/rspec/expectations/expectation_target.rb#L112-L116

The thinking is that in 99% of cases the matcher will have `description`, so while `respond_to?` is more performant that allowing an exception to be raised, that only comes into play 1% of the time and the rest of the time it's faster to not check `respond_to?`.

Honestly, I don't particularly care which route you go to deal with it (respond to or exception rescue).

It'd also be nice to have specs for matchers that don't support `diffable?` and `description` since it's an easy thing to forget.

myronmarston:
Left some more feedback.  Sorry for not catching this stuff sooner!

waterlink:
@myronmarston not a problem, very useful feedback

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
waterlink: I actually like the fact that don't checking `respond_to?` is actually faster 99% of the time.

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
waterlink: Stubbed out this constant, so spec actually checks that code does the right thing regardless of any future (hopefully numeric) value of this constant

waterlink:
@myronmarston WDYT about this commit https://github.com/rspec/rspec-expectations/commit/a1fdff877c751b0aff0c28bd05eec465be4b7f54 ?

waterlink:
Meh, it just goes crazy on 1.8.7.

waterlink:
So now I have a rubocop offense:

`lib/rspec/matchers/built_in/compound.rb:6:7: C: Class definition is too long. [102/100]`

How do we usually tackle these ones?
Of course I can start extracting some monadish small classes out of it :)

waterlink:
Actually, for me it feels like a some kind of `MatcherWrapper` is missing. Which will return default values on optional methods in matcher protocol if they are not defined on the matcher it wraps. Something along these lines: https://gist.github.com/waterlink/8afa40f0c869b6953f06

It reduces duplication present in couple of places. But probably will defeat the performance win over using `rescue NoMethodError` instead of `respond_to?`...

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
myronmarston: > the right thing regardless of any future (hopefully numeric) value of this constant

If you're concerned, you could always add a `expect(RSpec::Matchers::ExpectedsForMultipleDiffs::DESCRIPTION_MAX_LENGTH).to be_an(Integer)` expectation before stubbing it.

myronmarston:
> So now I have a rubocop offense:

> lib/rspec/matchers/built_in/compound.rb:6:7: C: Class definition is too long. [102/100]

> How do we usually tackle these ones?
> Of course I can start extracting some monadish small classes out of it :)

I like rubocop but I'm not sure that check has that much value, honestly.  Elsewhere we've just disabled it:

https://github.com/rspec/rspec-expectations/blob/45ff9faa26ea912d6b13d4044e5691b894d9fcc5/lib/rspec/matchers/built_in/raise_error.rb#L7

...which would work here. It might be worth disabling the check entirely for our projects, but that should be done in a separate PR with a separate discussion.

waterlink:
Ended up using rubocop directive, lets see if build is green

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
waterlink: Looks like probably too much, I can't imagine somebody making it non-numeric TBH :)

waterlink:
@myronmarston can you take another look? Looks like this one is ready

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      def diffs(differ_factory, actual)
+        diffs = []
+        @expected_list.each do |(expected, diff_label)|
+          diff = differ_factory.call.diff(actual, expected)
+          next if diff.empty?
+          diffs << "#{diff_label}#{diff}"
+        end
+        diffs.join("\n")
+      end
```
myronmarston: Could this function be simplified with `map` and `compact`?

``` ruby
def diffs(differ_factory, actual)
  @expected_list.map do |(expected, diff_label)|
    diff = differ_factory.call.diff(actual, expected)
    "#{diff_label}#{diff}" unless diff.empty?
  end.compact.join("\n")
end
```

(Not a merge blocker).

myronmarston:
LGTM.  I'd like to see this squashed into one commit (or a couple, if you see a few logically distinct steps) and then I'll be happy to merge.  Thanks for doing such a great job, @waterlink!

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
alex-fedorov: Done.

alex-fedorov:
@myronmarston I have squashed commits except the one that disables class length check on `Compound` matcher class.

myronmarston:
Thanks. Will merge when green.

waterlink:
Green

myronmarston:
Merged.  Thanks, @waterlink....your work here is really excellent!

waterlink:
@myronmarston Thanks!

PR #718, opened 2015-01-21
Fixup start and end with

myronmarston:


lib/rspec/matchers/built_in/start_or_end_with.rb
```diff
@@ -56,10 +52,16 @@ def subsets_comparable?
         end
       end
 
+      # For RSpec 3.1, the base class was named `StartAndEndWith`. For SemVer reasons,
+      # we still provide this constant until 4.0.
+      # @deprecated Use StartOrEndWith instead.
+      # @private
+      StartAndEndWith = StartOrEndWith
```
JonRowe: When we were building RSpec 3 we would define `const_missing` and define the lookup there to issue a deprecation, is it worth doing that here too?

myronmarston: I think it'd be worth doing in 3.99.  I've realized that deprecation warnings we added in 2.14, 2.13 and earlier wound up largely being a nuisance to users, given we had the whole 2.99 -> 3.0 process.  I'm planning on there being very few issued deprecations (if any) during 3.x releases before 3.99, unless there is a specific reason we feel users need to get a noisy deprecation before 3.99.

JonRowe: Fair enough :)

PR #722, opened 2015-01-29
followup for #713 :: Use differ directly instead of differ factory method

waterlink:
followup for #713 
related to rspec/rspec-support#160


myronmarston:
LGTM.  Will merge when green.

myronmarston:
Thanks!

PR #728, opened 2015-02-06
Refactor failure matcher definitions to use a recommended technique.

myronmarston:
I wouldn't encourage users to directly define methods on `RSpec::Matchers` so we probably shouldn't, either.


JonRowe:
LGTM

PR #730, opened 2015-02-08
Isolate a spec depending on set from the main process

JonRowe:
This is breaking my attempts to remove our usage of `Set` in `rspec-core` because we don't require it before use here, but I also don't want to poison our own env so isolate it in a sub process (we use this trick on rspec-core to test similar things).

### DIFF
--- a/spec/rspec/matchers/built_in/contain_exactly_spec.rb
+++ b/spec/rspec/matchers/built_in/contain_exactly_spec.rb
@@ -1,3 +1,4 @@
+require 'rspec/support/spec/in_sub_process'
 require 'timeout'
 
 class UnsortableObject
@@ -278,6 +279,8 @@ RSpec.describe "expect(...).not_to contain_exactly(:with, :multiple, :args)" do
 end
 
 RSpec.describe "matching against things that aren't arrays" do
+  include RSpec::Support::InSubProcess
+
   it "fails with nil and the expected error message is given" do
     expect {
       expect(nil).to contain_exactly(1, 2, 3)
@@ -297,10 +300,13 @@ RSpec.describe "matching against things that aren't arrays" do
   end
 
   it 'works with other collection objects' do
-    expect(Set.new([3, 2, 1])).to contain_exactly(1, 2, 3)
-    expect {
-      expect(Set.new([3, 2, 1])).to contain_exactly(1, 2)
-    }.to fail_including("expected collection contained:  [1, 2]")
+    in_sub_process do
+      require 'set'
+      expect(Set.new([3, 2, 1])).to contain_exactly(1, 2, 3)
+      expect {
+        expect(Set.new([3, 2, 1])).to contain_exactly(1, 2)
+      }.to fail_including("expected collection contained:  [1, 2]")
+    end
   end
 
   it 'works with non-enumerables that implement `to_ary`' do


JonRowe:
Merging

myronmarston:
I meant to comment on here before this got merged...but I think a better solution is to require `set` in that particular spec file.  The sub process thing is useful but has the side effect of creating a gap in our spec coverage on rubies that don't support forking (like JRuby and all rubies on windows).  IMO, set being required is really only a problem in our `lib` files.  Requiring it in a `spec` file has no detrimental effects I can think of so I think I'd prefer we go in that direction.

JonRowe:
> but I think a better solution is to require set in that particular spec file.

> Requiring it in a spec file has no detrimental effects I can think of so I think I'd prefer we go in that direction.

It enables us accidentally use Set elsewhere and not get caught out immediately.

myronmarston:
> It enables us accidentally use Set elsewhere and not get caught out immediately.

Given the extent of our cucumber coverage, I would imagine that would catch it.

JonRowe:
We use Set in our cucumber coverage of Core so I'd still rather not, I'd be ok with JRuby requiring this so the coverage isn't lost.

myronmarston:
> We use Set in our cucumber coverage of Core so I'd still rather not, I'd be ok with JRuby requiring this so the coverage isn't lost.

Are you referring to this?

https://github.com/rspec/rspec-core/blob/9d49bcb69d841d5d4258c2fb94cbb239e0868f9d/features/example_groups/shared_examples.feature#L51

That's the only use of set I can find in our rspec-core cukes.  Given that each cuke runs a fresh rspec process, that is 100% isolated (just like `in_sub_process`)...I think the cukes still work as coverage for this.

JonRowe:
Yeah but we run the specs against other repos, not the cukes, so I still think this should be isolated.

myronmarston:
> Yeah but we run the specs against other repos, not the cukes, so I still think this should be isolated.

Sure, but the individual repo builds run the cukes so it will still be caught.  There are a number of other places where a spec loads part of the stdlib to ensure parts of RSpec work well with the stdlib, for example:

https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/matchers/built_in/eq_spec.rb

We load `date`, `complex` and `bigdecimal` there.  I don't want to go down the path of always using the `in_sub_process` technique for these places, and if we're not going to do it in general why do it for this `set` case?  I think keeping coverage up against all supported rubies is a lot more important than the inconvenience of accidentally breaking one of the other repo builds because the spec use wasn't isolated in a sub process and the cuke didn't run during the other repo build.

Ultimately, I think we'd be better off adding a step to our build that enforces that only an allowed whitelist of libraries are loaded from the stdlib (I have some ideas how to do that).  I don't want to bother using `in_sub_process` for the isolation in the meantime.

JonRowe:
> inconvenience of accidentally breaking one of the other repo builds

That's not the issue, the issue is accidentally shipping broken code because we've forgotten to require something, or breaking our users applications because they don't realise they need to require something.

myronmarston:
> the issue is accidentally shipping broken code

Skipping specs that could otherwise run on rubies we support is a great way to ship broken code.  In my 4+ years of working on RSpec, I can't think of a single time we've shipped broken code because of something being required from a spec file.  I can think of many occasions when we expected something to "just work" on JRuby and it didn't, and we only knew about it because the spec ran against JRuby rather than being skipped on JRuby as your change here does.

Again, in our cukes, we repeatedly start new, fresh, RSpec processes with nothing extra loaded, and execute many example spec files covering most of RSpec's features.  If we use a library in `lib` but only have it required in a spec file (I think that's your concern, right?) a cuke is almost certainly going to fail.

More than that, if we're serious about minimizing what gets required outside of RSpec itself, I think we should write some shared examples in rspec-support to enforce that.  That's a much more robust way to deal with it.  I'm going to open an issue here in rspec-support about that in a bit.

Would it really bother you if I push a revert of this? (I think that's the best course of action and I hate vetoing things but this conversation doesn't seem to be going anywhere...)

JonRowe:
> Skipping specs that could otherwise run on rubies we support is a great way to ship broken code. In my 4+ years of working on RSpec, I can't think of a single time we've shipped broken code because of something being required from a spec file. I can think of many occasions when we expected something to "just work" on JRuby and it didn't, and we only knew about it because the spec ran against JRuby rather than being skipped on JRuby as your change here does.

I'm not suggesting skipping specs, I'm suggesting doing this on Rubies that support `in_sub_process` and supporting JRuby the old way.

>  If we use a library in lib but only have it required in a spec file (I think that's your concern, right?) a cuke is almost certainly going to fail.

That's only half my concern. The rest of it is external test suites.

> More than that, if we're serious about minimizing what gets required outside of RSpec itself, I think we should write some shared examples in rspec-support to enforce that. That's a much more robust way to deal with it. I'm going to open an issue here in rspec-support about that in a bit.

This I'm on board with.

> Would it really bother you if I push a revert of this?

Yes.

> (I think that's the best course of action and I hate vetoing things but this conversation doesn't seem to be going anywhere...)

I'm willing to do both things here, use `in_sub_process` where we can and use a vanilla require where we can't...

myronmarston:
> I'm suggesting doing this on Rubies that support in_sub_process and supporting JRuby the old way.

Why did you merge this before rigging up the alternate JRuby way, then?  And what would that look like?

> That's only half my concern. The rest of it is external test suites.

I have no idea what you mean.  Can you explain?

> Yes.

Well, it bothers me that you merged this before anyone had a chance to review it :(.  I'd prefer we reverted and kept discussing it.

> I'm willing to do both things here, use `in_sub_process` where we can and use a vanilla require where we can't...

Maybe `in_sub_process_if_possible`?

Also, I still don't understand what makes `set` special.  We have multiple other places where we require something from stdlib in a spec file to test how an RSpec feature interacts with something from stdlib, and we've not dealt with that in this fashion before.  If we're going to make this the new way we deal with it, I'd like it to be consistent across the board (e.g. in a PR that does that in all places rather than just this one).

JonRowe:
> Why did you merge this before rigging up the alternate JRuby way, then? And what would that look like?
> Well, it bothers me that you merged this before anyone had a chance to review it :(. I'd prefer we reverted and kept discussing it.

I'm sorry, I didn't think anyone would have an issue with it.

> Also, I still don't understand what makes set special. We have multiple other places where we require something from stdlib in a spec file to test how an RSpec feature interacts with something from stdlib, and we've not dealt with that in this fashion before. If we're going to make this the new way we deal with it, I'd like it to be consistent across the board (e.g. in a PR that does that in all places rather than just this one).

It isn't, I just came across this during work elsewhere I'd happily tackle those too.

JonRowe:
(I'm going to open a PR using the new support stuff for those later today)

PR #734, opened 2015-02-13
Isolate specs from std lib when possible

JonRowe:
Ping @myronmarston I went through and isolated out specs from the standard library.

### DIFF
--- a/spec/rspec/expectations/configuration_spec.rb
+++ b/spec/rspec/expectations/configuration_spec.rb
@@ -45,8 +45,6 @@ module RSpec
       end
 
       context 'on an interpreter that does not provide BasicObject', :uses_should, :unless => defined?(::BasicObject) do
-        include RSpec::Support::InSubProcess
-
         def with_delegate
           in_sub_process_if_possible do
             require 'delegate'
--- a/spec/rspec/expectations/configuration_spec.rb
+++ b/spec/rspec/expectations/configuration_spec.rb
@@ -1,5 +1,3 @@
-require 'delegate'
-
 module RSpec
   module Expectations
     RSpec.describe Configuration do
@@ -47,7 +45,15 @@ module RSpec
       end
 
       context 'on an interpreter that does not provide BasicObject', :uses_should, :unless => defined?(::BasicObject) do
-        before { RSpec::Expectations::Syntax.disable_should(Delegator) }
+        include RSpec::Support::InSubProcess
+
+        def with_delegate
+          in_sub_process_if_possible do
+            require 'delegate'
+            RSpec::Expectations::Syntax.disable_should(Delegator)
+            yield
+          end
+        end
 
         let(:klass) do
           Class.new(SimpleDelegator) do
@@ -58,9 +64,11 @@ module RSpec
         let(:instance) { klass.new(Object.new) }
 
         it 'provides a means to manually add it Delegator' do
-          instance.should_not respond_to(:delegated?) # because #should is being delegated...
-          config.add_should_and_should_not_to Delegator
-          instance.should respond_to(:delegated?) # now it should work!
+          with_delegate do
+            instance.should_not respond_to(:delegated?) # because #should is being delegated...
+            config.add_should_and_should_not_to Delegator
+            instance.should respond_to(:delegated?) # now it should work!
+          end
         end
       end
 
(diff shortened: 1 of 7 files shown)


spec/rspec/expectations/configuration_spec.rb
```diff
@@ -47,7 +45,15 @@ class << rspec_dup; undef configuration; end
       end
 
       context 'on an interpreter that does not provide BasicObject', :uses_should, :unless => defined?(::BasicObject) do
-        before { RSpec::Expectations::Syntax.disable_should(Delegator) }
+        include RSpec::Support::InSubProcess
+
+        around do |example|
+          RSpec::Expectations::Syntax.disable_should(Delegator)
```
myronmarston: Given `require 'delegate'` now comes after this line, and `Delegator` is defined in the `delegate` stdlib...won't this fail?

spec/rspec/expectations/configuration_spec.rb
```diff
@@ -47,7 +45,15 @@ class << rspec_dup; undef configuration; end
       end
 
       context 'on an interpreter that does not provide BasicObject', :uses_should, :unless => defined?(::BasicObject) do
-        before { RSpec::Expectations::Syntax.disable_should(Delegator) }
+        include RSpec::Support::InSubProcess
```
myronmarston: This line shows up again and again in this PR.  WDYT of `RSpec.configure { |c| c.include RSpec::Support::InSubProcess }` in `spec_helper`?

JonRowe: Fine by me, it's just we'd only included it where needed previously :D

spec/rspec/expectations/configuration_spec.rb
JonRowe: True... I guess it passed for me because I have BasicObject

spec/rspec/expectations/configuration_spec.rb
```diff
@@ -47,7 +45,13 @@ class << rspec_dup; undef configuration; end
       end
 
       context 'on an interpreter that does not provide BasicObject', :uses_should, :unless => defined?(::BasicObject) do
-        before { RSpec::Expectations::Syntax.disable_should(Delegator) }
+        around do |example|
+          in_sub_process_if_possible do
```
myronmarston: `in_sub_process`/`in_sub_process_if_possible` is not safe to use in an `around` hook -- it makes the example _always_ pass.   Try adding literally anything (even something like `adfkjsadfsa` to the first line of the example) and you'll see it passes.

That's because `in_sub_process` communicates pass/fail to the parent process by re-raising an exception, but in an `around` hook, the example failure exception has already been handled by RSpec, and the execution result state gets updated--but that happens in the short-lived child process, and the parent process is never made aware of it.

myronmarston: In this case I think you can do away with the `around` hook entirely and just move it all into the example itself.

JonRowe: Good point, I actually was meaning to circle back to this due to issues elsewhere meaning it all pended!

JonRowe: (Fixed)

myronmarston:
This LGTM.  I'd still like to understand what problem you feel this solves, though.  In #730 you said:

> That's only half my concern. The rest of it is external test suites.

I still don't understand what "the rest of it is external test suites" means.  Can you explain a problem external test suites would have that this will solve?

JonRowe:
Yeah I think I might have been wrong with the initial implications of using `require` in our specs but I'd still like to do this to be on the safe side, initially the missing `require` was highlighted by a build over on `rspec-core` but I'd still like us not to have these components loaded if we don't need to. I feel it reduces our cross section and makes us less likely to accidentally rely on the standard library.

 IMO this is important and I actually noticed Travis was far more sensitive to missing components than my own machine, I'm hypothesising that this is related to me using rvm and not bundler bin stubs...

myronmarston:
> IMO this is important and I actually noticed Travis was far more sensitive to missing components than my own machine, I'm hypothesising that this is related to me using rvm and not bundler bin stubs...

Do you use `--standalone` (as our travis builds do and I do)?  Besides speeding things up by avoiding the bundler tax at runtime, the act of avoiding bundler/rubygems at runtime means any stdlibs used by those aren't automatically loaded, which could affect here.

Anyhow, merge away.

JonRowe:
> Do you use --standalone (as our travis builds do and I do)? Besides speeding things up by avoiding the bundler tax at runtime

Nope, I have rvm gemsets instead, same performance enhancement.

PR #736, opened 2015-02-15
Allow custom matchers to use methods in Pretty

jaredbeck:
When writing a `failure_message` for a custom matcher, the methods in `Pretty` are useful.

Maybe we don't need all of them to be public, but some, like `to_sentence` would be nice.

This PR is just to get the conversation started.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,8 @@
 
 Enhancements:
 
+* Expose `RSpec::Matchers::EnglishPhrasing` to make it easier to write
+  nice failure messages in custom matchers. (Jared Beck, #736)
 * Make RSpecs fail matchers (for checking examples fail) publically available
   for use by extension/plugin authors. (Charlie Rudolph, #729)
 * Avoid loading `tempfile` (and its dependencies) unless
--- a/lib/rspec/matchers/built_in/base_matcher.rb
+++ b/lib/rspec/matchers/built_in/base_matcher.rb
@@ -53,11 +53,12 @@ module RSpec
         end
 
         # @api private
-        # Generates a "pretty" description using the logic in {Pretty}.
+        # Generates a description using {EnglishPhrasing}.
         # @return [String]
         def description
-          return name_to_sentence unless defined?(@expected)
-          "#{name_to_sentence}#{to_sentence @expected}"
+          desc = EnglishPhrasing.split_words(name)
+          desc << EnglishPhrasing.list(@expected) if defined?(@expected)
+          desc
         end
 
         # @api private
@@ -84,7 +85,8 @@ module RSpec
 
         def assert_ivars(*expected_ivars)
           return unless (expected_ivars - present_ivars).any?
-          raise "#{self.class.name} needs to supply#{to_sentence expected_ivars}"
+          ivar_list = EnglishPhrasing.list(expected_ivars)
+          raise "#{self.class.name} needs to supply#{ivar_list}"
         end
 
         if RUBY_VERSION.to_f < 1.9
--- a/lib/rspec/matchers/built_in/eq.rb
+++ b/lib/rspec/matchers/built_in/eq.rb
@@ -20,7 +20,8 @@ module RSpec
         # @api private
         # @return [String]
         def description
-          "#{name_to_sentence} #{@expected.inspect}"
+          english_name = EnglishPhrasing.split_words(name)
+          "#{english_name} #{@expected.inspect}"
         end
 
         # @api private
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -300,7 +300,9 @@ module RSpec
 
         # The default description.
         def description
-          "#{name_to_sentence}#{to_sentence expected}#{chained_method_clause_sentences}"
+          english_name = EnglishPhrasing.split_words(name)
+          expected_list = EnglishPhrasing.list(expected)
+          "#{english_name}#{expected_list}#{chained_method_clause_sentences}"
         end
 
         # Matchers do not support block expectations by default. You
@@ -320,7 +322,9 @@ module RSpec
           return '' unless Expectations.configuration.include_chain_clauses_in_custom_matcher_descriptions?
 
           @chained_method_clauses.map do |(method_name, method_args)|
-            " #{split_words(method_name)}#{to_sentence(method_args)}"
+            english_name = EnglishPhrasing.split_words(method_name)
+            arg_list = EnglishPhrasing.list(method_args)
+            " #{english_name}#{arg_list}"
           end.join
         end
       end
@@ -336,7 +340,7 @@ module RSpec
         # Allows expectation expressions to be used in the match block.
         include RSpec::Matchers
 
-        # Converts matcher name and expected args to an English expresion.
+        # Facilitates better descriptions and failure messages.
         include RSpec::Matchers::Pretty
 
         # Supports the matcher composability features of RSpec 3+.
--- a/spec/rspec/matchers/english_phrasing_spec.rb
+++ b/spec/rspec/matchers/english_phrasing_spec.rb
@@ -1,0 +1,81 @@
+module RSpec
+  module Matchers
+    RSpec.describe EnglishPhrasing do
+      describe ".split_words" do
+        it "replaces underscores with spaces" do
+          expect(
+            described_class.split_words(:banana_creme_pie)
+          ).to eq("banana creme pie")
+        end
+
+        it "first casts its argument to string" do
+          arg = double(:to_s => "banana")
+          expect(described_class.split_words(arg)).to eq("banana")
+        end
+      end
+
+      describe ".list" do
+        context "given nil" do
+          it "returns value from inspect, and a leading space" do
+            expect(described_class.list(nil)).to eq(" nil")
+          end
+        end
+
+        context "given a Struct" do
+          it "returns value from inspect, and a leading space" do
+            banana = Struct.new("Banana", :flavor).new
+            expect(
+              described_class.list(banana)
+            ).to eq(" #{banana.inspect}")
+          end
+        end
+
+        context "given an Enumerable" do
+          before do
+            allow(described_class).to(
+              receive(:item_description).and_return("Banana")
+            )
+          end
+
+          context "with zero items" do
+            it "returns the empty string" do
+              expect(described_class.list([])).to eq("")
+            end
+          end
+
+          context "with one item" do
+            let(:list) { [double] }
+            it "returns description, and a leading space" do
+              expect(described_class.list(list)).to eq(" Banana")
+              expect(described_class).to(
+                have_received(:item_description).once
+              )
+            end
+          end
+
+          context "with two items" do
+            let(:list) { [double, double] }
+            it "returns descriptions, and a leading space" do
+              expect(described_class.list(list)).to eq(" Banana and Banana")
+              expect(described_class).to(
+                have_received(:item_description).twice
+              )
+            end
+          end
+
+          context "with three items" do
+            let(:list) { [double, double, double] }
+            it "returns descriptions, and a leading space" do
+              expect(
+                described_class.list(list)
+              ).to eq(" Banana, Banana, and Banana")
+              expect(described_class).to(
+                have_received(:item_description).exactly(3).times
+              )
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -4,6 +4,7 @@ RSpec::Support.define_optimized_require_for_rspec(:matchers) { |f| require_relat
 
 %w[
   pretty
+  english_phrasing
   composable
   built_in
   generated_descriptions
@@ -243,7 +244,7 @@ module RSpec
     #     alias $1 $2
     def self.alias_matcher(new_name, old_name, options={}, &description_override)
       description_override ||= lambda do |old_desc|
-        old_desc.gsub(Pretty.split_words(old_name), Pretty.split_words(new_name))
+        old_desc.gsub(EnglishPhrasing.split_words(old_name), EnglishPhrasing.split_words(new_name))
       end
       klass = options.fetch(:klass) { AliasedMatcher }
 
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -87,11 +87,11 @@ module RSpec
         end
 
         def expected_to_sentence
-          split_words(@expected)
+          EnglishPhrasing.split_words(@expected)
         end
 
         def args_to_sentence
-          to_sentence(@args)
+          EnglishPhrasing.list(@args)
         end
       end
 
@@ -252,7 +252,7 @@ module RSpec
         end
 
         def prefix_to_sentence
-          split_words(@prefix)
+          EnglishPhrasing.split_words(@prefix)
         end
 
         def failure_message_expecting(value)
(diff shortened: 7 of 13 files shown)


myronmarston:
I'm open to exposing this, but have a few suggestions:

* I really don't want `name` or `name_to_sentence`, in their current forms publicly exposed.  Relying upon the `@name` instance variable from the host class has always made me nervous and I don't want to encode that as part of the public API.
* The more I think about it, the more convinced I am that this would work just as well, if not better, as a utility module that supports these methods being called directly on it rather than being a mix-in.  The only aspect that relies upon being mixed in is the `name` stuff I mentioned above, that I'd rather we refactor, anyway.
* `Pretty` is OK as an internal module name, but isn't very descriptive as a public module.  I'm thinking we should rename it.  Maybe `EnglishExpressions` or something is better?

JonRowe:
> * The more I think about it, the more convinced I am that this would work just as well, if not better, as a utility module that supports these methods being called directly on it rather than being a mix-in.  The only aspect that relies upon being mixed in is the `name` stuff I mentioned above, that I'd rather we refactor, anyway.
> * `Pretty` is OK as an internal module name, but isn't very descriptive as a public module.  I'm thinking we should rename it.  Maybe `EnglishExpressions` or something is better?

I like both these ideas, feel like tackling it @jaredbeck ?

jaredbeck:
> I like both these ideas, feel like tackling it @jaredbeck ?

Heck yeah.  Let me give it a try.

jaredbeck:
> Pretty is OK as an internal module name, but isn't very descriptive as a public module. I'm thinking we should rename it. Maybe EnglishExpressions or something is better?

`EnglishPhrases` is a bit easier to type, or how about simply `English`?

jaredbeck:
OK, after some heckling from rubocop and yard, I got the travis build passing, so I guess this is ready for review.  Please see the detailed commit message for an overview and rationale.

There's only one thing I'm not happy about.   It's easy to use `English` from within, e.g. `BaseMatcher`, but inside `Matchers.define` I had to use the fully-qualified `RSpec::Matchers::English`, as you can see in `define_matcher.feature`.  This feels awkward, and I'm hoping y'all can suggest a better way.

JonRowe:
I'd prefer `EnglishPhrases` over `English`

Changelog.md
```diff
@@ -2,6 +2,11 @@
 
 Enhancements:
 
+* Allow perfectionists who write custom matchers to improve their
+  `failure_message` by using methods from RSpec::Matchers::Pretty,
+  now renamed to RSpec::Matchers::English.  Before this, the module
```
JonRowe: This is a bit verbose, should wrap constant names in backticks and you have an excess space after a period. How about:

```
* Expose `RSpec::Matchers::EnglishPhrases` to make it easier to write nice failure messages in custom matchers.
```

lib/rspec/matchers/hash_formatting.rb
```diff
@@ -0,0 +1,20 @@
+module RSpec
+  module Matchers
+    # @api private
```
JonRowe: If these are still to be private they probably could have been left in `Pretty`...

jaredbeck:
I've changed the name to `EnglishPhrases` and replaced the changelog entry.

> If these are still to be private they probably could have been left in Pretty...

I think there's some value in separating `Naming` from `HashFormatting`.  For example, there are two tests that include `HashFormatting`, but they don't need to include `Naming`.

JonRowe:
> I think there's some value in separating Naming from HashFormatting. For example, there are two tests that include HashFormatting, but they don't need to include Naming.

We try to reduce the number of files we use (because these things actually do matter when it comes to library load times etc) and as it's a private API that we've had in one place for a while I think it's ok to leave them there.

jaredbeck:
> We try to reduce the number of files we use ..

Sounds good.  I've re-combined them.

lib/rspec/matchers/english_phrases.rb
```diff
@@ -0,0 +1,58 @@
...
+      class << self
+        # Converts a symbol into an English expression.
+        #
+        #     humanize(:banana_creme_pie) #=> "banana creme pie"
+        #
+        # The method name is intended to be familiar to
+        # ActiveSupport users.  It is similar, but has
+        # fewer features.
+        #
+        def humanize(sym)
```
myronmarston: Is `humanize` what AS calls this?

lib/rspec/matchers/english_phrases.rb
```diff
@@ -0,0 +1,58 @@
...
+      private
+
+        def is_matcher_with_description?(object)
+          RSpec::Matchers.is_a_matcher?(object) &&
+            object.respond_to?(:description)
+        end
+
+        # Converts the given item to string suitable for use in a
+        # list expression.
+        def list_item(obj)
```
myronmarston: I think I'd prefer this be called `item_description`.  The caller is `list `but there's nothing about this logic that is limited to lists.

lib/rspec/matchers/english_phrases.rb
```diff
@@ -0,0 +1,58 @@
+module RSpec
+  module Matchers
+    # Facilitates converting ruby objects to English phrases.
+    module EnglishPhrases
+      class << self
```
myronmarston: I have a preference for using `def self.method_name` over opening the singleton class, and that's what we've favored elsewhere. Mind switching it?

lib/rspec/matchers/english_phrases.rb
myronmarston: Honestly, I think I liked the old name of `split_words` better...

lib/rspec/matchers/english_phrases.rb
```diff
@@ -0,0 +1,58 @@
+module RSpec
+  module Matchers
+    # Facilitates converting ruby objects to English phrases.
+    module EnglishPhrases
```
myronmarston: I think I slightly prefer `EnlighsPhrasing` over `EnglishPhrases` (hard to explain why, though, and it's a very lightly helped opinion).

lib/rspec/matchers/built_in/eq.rb
```diff
@@ -20,7 +20,8 @@ def failure_message_when_negated
         # @api private
         # @return [String]
         def description
-          "#{name_to_sentence} #{@expected.inspect}"
+          english_name = EnglishPhrases.humanize(name)
+          "%s %s" % [english_name, @expected.inspect]
```
myronmarston: I much prefer using string interpolation to using format strings and `%` as you've done here and below.  Among other things, this form does an extra array allocation.

lib/rspec/matchers/english_phrases.rb
```diff
@@ -0,0 +1,58 @@
...
+          when 0
+            ""
+          when 1
+            " #{items[0]}"
+          when 2
+            " #{items[0]} and #{items[1]}"
+          else
+            " #{items[0...-1].join(', ')}, and #{items[-1]}"
+          end
+        end
```
myronmarston: I think the leading space is a little odd in this.  I know it's how it worked before but I think it will surprise users.  You mind removing it and updating the call sites?

lib/rspec/matchers/english_phrases.rb
myronmarston: I think I figured out what I dislike about `EnglishPhrases`...as a plural, it sounds like it's a list of phrases, whereas `EnglishPhrasing` reads more like it helps you create strings using English phrasing (which is more accurate to what it does).

jaredbeck: Makes sense to me.

lib/rspec/matchers/built_in/eq.rb
jaredbeck: OK, I'll switch back to interpolation.  If performance were not a concern, I like format strings because the non-interpolated content (in this case, whitespace) is easier to see.

lib/rspec/matchers/english_phrases.rb
jaredbeck: Good point.  It would definitely surprise people.

lib/rspec/matchers/english_phrases.rb
jaredbeck: Without the `class << self`, I don't think the `private` keyword will work.

```ruby
module Works
  class << self
    def z; end
    private
    def y; end
  end
end
Works.z #=> nil
Works.y # NoMethodError: private method `y' called for Z:Module

module DoesntWork
  def self.a; end
  private
  def self.b; end
end
DoesntWork.a #=> nil
DoesntWork.b #=> nil
```

Is there another way to do private module methods?

lib/rspec/matchers/english_phrases.rb
jaredbeck: Hmm.  There isn't always a leading space.  If we remove the leading space here, we could end up duplicating the "whitespace logic" in each of the call sites.

myronmarston:
> Is there another way to do private module methods?

Use `private_class_method :method_name`.

myronmarston:
> Hmm. There isn't always a leading space. If we remove the leading space here, we could end up duplicating the "whitespace logic" in each of the call sites.

Good point.  I think you should add a `@note` YARD doc tag mentioning that the returned string has a leading space except when given an empty list.  That way we're explicitly telling people about it.

jaredbeck:
> Use private_class_method ..
> 
> .. add a `@note` YARD doc tag ..

Done and done.  I also added unit tests for `.split_words` and `.list` (spec/rspec/matchers/english_phrasing_spec.rb) and I look forward to any feedback you might have there.

jaredbeck:
>  I also added unit tests ..

If you're satisfied with the unit tests, I'd suggest that they replace the cucumber test I wrote.  The unit tests are faster, and the cucumber test feels more like a test than like documentation.

JonRowe:
I feel we should keep the cuke as it helps expose this feature...

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -53,11 +53,12 @@ def match_unless_raises(*exceptions)
...
 
         # @api private
-        # Generates a "pretty" description using the logic in {Pretty}.
+        # Generates a description using {EnglishPhrasing}.
         # @return [String]
         def description
-          return name_to_sentence unless defined?(@expected)
-          "#{name_to_sentence}#{to_sentence @expected}"
+          desc = EnglishPhrasing.split_words(name)
+          desc += EnglishPhrasing.list(@expected) if defined?(@expected)
```
myronmarston: For perf reasons we can use `<<` here rather than `+=` and allocate one less string.

lib/rspec/matchers/dsl.rb
```diff
@@ -336,7 +340,7 @@ class Matcher
         # Allows expectation expressions to be used in the match block.
         include RSpec::Matchers
 
-        # Converts matcher name and expected args to an English expresion.
+        # Facilitates better descriptions and failure messages.
         include RSpec::Matchers::Pretty
```
myronmarston: Can we do away with this module include entirely?

spec/rspec/matchers/english_phrasing_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    RSpec.describe EnglishPhrasing do
+      describe ".split_words" do
+        it "replaces underscores with spaces" do
+          expect(described_class.split_words(:banana_creme_pie)).to \
+            eq("banana creme pie")
+        end
+
+        it "first casts its argument to string" do
+          x = double
+          allow(x).to receive(:to_s).and_return("banana")
```
myronmarston: These two lines can be simplified to one:

``` ruby
double(:to_s => "banana")
```

spec/rspec/matchers/english_phrasing_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+            expect(described_class.list(nil)).to eq(" nil")
+          end
+        end
+
+        context "given a Struct" do
+          it "returns value from inspect, and a leading space" do
+            banana = Struct.new("Banana", :flavor)
+            allow(banana).to receive(:inspect).and_return("Delicious")
+            expect(described_class.list(banana)).to eq(" Delicious")
+          end
```
myronmarston: `banana` is a struct class, not a struct instance, and it doesn't satisfy the `Struct ===` check of the `list` method:

``` ruby
irb(main):002:0> Person = Struct.new(:name)
=> Person
irb(main):003:0> Struct === Person
=> false
irb(main):004:0> Struct === Person.new
=> true
```

You'd have to do `banana.new` to get an instance that satisfies that check.

It also seems a bit smelly to stub a core method like `inspect`...you could just use `banana_instance.inspect` in the expectation instead.

spec/rspec/matchers/english_phrasing_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+            it "returns the empty string" do
+              expect(described_class.list([])).to eq("")
+            end
+          end
+
+          context "with one item" do
+            let(:list) { [double] }
+            it "returns description, and a leading space" do
+              expect(described_class.list(list)).to eq(" Banana")
+              expect(described_class).to \
```
myronmarston: I'm really not a fan of the use of `\` between `to` and the matcher.  It's not a style we use anywhere else in RSpec's specs. You mind putting this all on one line?  If you feel some of these are two long, you can use the styles we use elsewhere:

``` ruby
expect(
  some_long_expression.that_goes_on_its_own_line
).to something

# or

expect(some_expression).to matcher(
  matcher_args,
  can, go,
  on, other_lines
)
```

(Or some combination thereof).

myronmarston:
> I feel we should keep the cuke as it helps expose this feature...

I'm :-1: on this.  IMO, the cukes are not an effective way to document every API.  They are there to document the main use case but I don't expect this to be a commonly used API.  The YARD docs are better for that.  Plus having the cuke has a cost that the YARD docs have, in terms of longer build times, etc.

myronmarston:
@jaredbeck -- sorry about the delay.  I left some more comments but this is getting close to ready to merge.

jaredbeck:
> @jaredbeck -- sorry about the delay. I left some more comments but this is getting close to ready to merge.

No worries.  Thanks for the great comments, I'm learning from them.  I'll ping again when I've addressed them.  Let me know what you guys decide about the cuke.

lib/rspec/matchers/dsl.rb
jaredbeck: I don't know.  It doesn't look easy.  `Pretty#name` is used in at least one place: `Matcher#inspect`.  

However, I think we can make `Pretty#underscore` private, and I have done so.

jaredbeck: > However, I think we can make Pretty#underscore private, and I have done so.

On second thought, though `Pretty#underscore` was designated `@api private`, I don't know exactly what that means in this project, so making `#underscore` private may or may not constitute a breaking change.  Please advise.

jaredbeck:
> Thanks for the great comments .. I'll ping again when I've addressed them. 

Ready for another review, thanks!

features/custom_matchers/define_matcher.feature
```diff
@@ -335,3 +335,37 @@ Feature: define matcher
...
+      RSpec.describe Carrot do
+        it { is_expected.not_to descend_from([Fruit, Food, Object]) }
+      end
+      """
+    When I run `rspec ./matcher_english_spec.rb --format documentation`
+    Then the exit status should be 0
+    And the output should contain all of these:
+      | 2 examples, 0 failures                          |
+      | should descend from Fruit, Food, and Object     |
+      | should not descend from Fruit, Food, and Object |
```
myronmarston: I'm still in favor of this scenario being removed, given the spec coverage you added.

jaredbeck:
> I'm still in favor of this [cuke] scenario being removed, given the spec coverage you added.

Removed.

myronmarston:
Thanks, @jaredbeck!

jaredbeck:
> Thanks, @jaredbeck!

Thank *you*.  It was fun.

PR #737, opened 2015-02-16
Docs: Recommend the USE_GIT_REPOS var

jaredbeck:
As a first-time contributor to rspec, I was unfamiliar with the dynamic Gemfile, and I ran into trouble because I had an old clone sitting in just the right place.  The dynamic Gemfile thought I wanted to use that old clone, so I got a dependency conflict.

It seems to me that using dependencies from the filesystem is a less common use-case than simply using the latest from git, so I think the instructions should explicity USE_GIT_REPOS.

### DIFF
--- a/DEV-README.md
+++ b/DEV-README.md
@@ -3,6 +3,7 @@
     git clone git://github.com/rspec/rspec-expectations.git
     cd rspec-expectations
     gem install bundler
+    export USE_GIT_REPOS=1 # optional
     bundle install
 
 Now you should be able to run any of:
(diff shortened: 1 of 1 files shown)


JonRowe:
Heh, I suspect as a team we'll be on the fence about this, I introduced this because I tend to work in isolation on a repo, where as @samphippen and @myronmarston use the rspec/rspec-dev repo to manage all the dependencies at once. 

If we did make this change we'd want to recommend it across the board and not just in one projects readme.

myronmarston:
> It seems to me that using dependencies from the filesystem is a less common use-case

I have absolutely no idea if that's the more common or less common case.  The `USE_GIT_REPOS` ENV var is something @JonRowe added and, as far as I knew until now, he was the only one using it, and I thought of it as an uncommon way to manage things.

This is also not an rspec-expectations specific-issue.  Any discussion/recommendations should probably go in a central place.  I'm planning to work on updating the rsepc-dev README soon to provide better advice.  Maybe you can review that once I've updated it?

JonRowe:
Closing for now, I feel we should centralise the instructions for this, maybe someone should open a PR on rspec-dev

jaredbeck:
An alternative would be to mention `USE_GIT_REPOS` as a footnote, perhaps phrased as a troubleshooting step..  "oh, if you ran into trouble because you had an old clone sitting in just the right place, you might want to try USE_GIT_REPOS etc."

myronmarston:
I'm planning to update the rspec-dev docs soon, and will take this into account.

PR #739, opened 2015-02-18
Leverage new differ autoload.

myronmarston:
Builds on rspec/rspec-support#181. This avoids the
cost of loading diff/lcs, pp, etc for cases where
we don’t need to print any diffs.


JonRowe:
Appveyor has gone away for some reason but this LGTM so merging regardless.

PR #740, opened 2015-02-22
Use precision based formatting provided by rspec-support

gavingmiller:
Precision output was moved to rspec-support, this commit removes the code that was moved. As part of rspec/rspec-mocks#898

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -135,11 +135,11 @@ module RSpec
   #       end
   #
   #       def failure_message
-  #         "expected #{@target.inspect} to be in Zone #{@expected}"
+  #         "expected #{RSpec::Support::ObjectInspector.inspect(@target)} to be in Zone #{@expected}"
   #       end
   #
   #       def failure_message_when_negated
-  #         "expected #{@target.inspect} not to be in Zone #{@expected}"
+  #         "expected #{RSpec::Support::ObjectInspector.inspect(@target)} not to be in Zone #{@expected}"
   #       end
   #     end
   #
--- a/lib/rspec/matchers/built_in/base_matcher.rb
+++ b/lib/rspec/matchers/built_in/base_matcher.rb
@@ -80,6 +80,16 @@ module RSpec
           false
         end
 
+        # @api private
+        def expected_formatted
+          RSpec::Support::ObjectInspector.inspect(@expected)
+        end
+
+        # @api private
+        def actual_formatted
+          RSpec::Support::ObjectInspector.inspect(@actual)
+        end
+
         # @private
         def self.matcher_name
           @matcher_name ||= underscore(name.split("::").last)
@@ -141,7 +151,7 @@ module RSpec
           # you often only need to override `description`.
           # @return [String]
           def failure_message
-            "expected #{actual.inspect} to #{description}"
+            "expected #{RSpec::Support::ObjectInspector.inspect(@actual)} to #{description}"
           end
 
           # @api private
@@ -150,7 +160,7 @@ module RSpec
           # you often only need to override `description`.
           # @return [String]
           def failure_message_when_negated
-            "expected #{actual.inspect} not to #{description}"
+            "expected #{RSpec::Support::ObjectInspector.inspect(@actual)} not to #{description}"
           end
 
           # @private
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -136,11 +136,11 @@ module RSpec
         def expected_error
           case @expected_message
           when nil
-            description_of(@expected_error)
+            RSpec::Support::ObjectInspector.inspect(@expected_error)
           when Regexp
-            "#{@expected_error} with message matching #{@expected_message.inspect}"
+            "#{@expected_error} with message matching #{RSpec::Support::ObjectInspector.inspect(@expected_message)}"
           else
-            "#{@expected_error} with #{description_of @expected_message}"
+            "#{@expected_error} with #{RSpec::Support::ObjectInspector.inspect(@expected_message)}"
           end
         end
 
@@ -155,7 +155,7 @@ module RSpec
 
           backtrace = format_backtrace(@actual_error.backtrace)
           [
-            ", got #{@actual_error.inspect} with backtrace:",
+            ", got #{RSpec::Support::ObjectInspector.inspect(@actual_error)} with backtrace:",
             *backtrace
           ].join("\n  # ")
         end
--- a/spec/rspec/expectations/expectation_target_spec.rb
+++ b/spec/rspec/expectations/expectation_target_spec.rb
@@ -117,25 +117,27 @@ module RSpec
           end
 
           it "uses the matcher's `description` in the error message" do
-            custom_matcher = Module.new do
-              def self.supports_block_expectations?; false; end
-              def self.description; "matcher-description"; end
+            RSpec::Matchers.define :custom_matcher_with_description do
+              def supports_block_expectations?; false; end
+
+              match { true }
+              description { "matcher-description" }
             end
 
             expect {
-              expect { }.to custom_matcher
+              expect { }.to custom_matcher_with_description
             }.to fail_with(/\(matcher-description\)/)
           end
 
           context 'when the matcher does not define `description` (since it is an optional part of the protocol)' do
             it 'uses `inspect` in the error message instead' do
-              custom_matcher = Module.new do
-                def self.supports_block_expectations?; false; end
-                def self.inspect; "matcher-inspect"; end
+              RSpec::Matchers.define :custom_matcher_without_description do
+                def supports_block_expectations?; false; end
+                def inspect; "matcher-inspect"; end
               end
 
               expect {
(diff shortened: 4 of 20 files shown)


myronmarston:
Thanks for taking a crack at this.  While you're at it, do you mind doing an audit of the other places we use `inspect` to format an object for display and considering if it should be updated to use the new rspec-support method?

gavingmiller:
@myronmarston absolutely. I've got a partial set of locations, and will continue work on that.

gavingmiller:
@myronmarston started to move the ObjectInspector work into rspec-expectations. Two questions for you:

1. Is it sufficient to do replacement of `inspect` for `ObjectInspector` without writing new tests as long as the old tests don't break?
2. It appears that the build machine is using an outdated version of `rspec-support`. Is there something I need to do to resolve that?

JonRowe:
> 1. Is it sufficient to do replacement of inspect for ObjectInspector without writing new tests as long as the old tests don't break?

It's preferable for me!

> 2. It appears that the build machine is using an outdated version of `rspec-support`. Is there something I need to do to resolve that?

I think you just haven't added a require for it anywhere?

spec/spec_helper.rb
```diff
@@ -1,3 +1,4 @@
+require 'rspec/support/object_inspector'
```
myronmarston: This require makes `RSpec::Support::ObjectInspector` available in our spec suite, but does not make it available for users of rspec-expectations.  The failing cucumber features demonstrate that; they are failing because it hasn't been loaded in the environment.  It's not the responsibility of each test environment to load this file. Instead, it should be moved into `lib` so that it's always available to anyone who loads rspec-expectations.  I think it should go in `lib/rspec/matchers.rb`.

Also, I'd prefer you use `RSpec::Support.require_rspec_support "object_inspector"`.  That uses an optimized form of require that is more performant on rubies that support `require_relative`.

spec/rspec/expectations/expectation_target_spec.rb
```diff
@@ -117,25 +117,27 @@ def self.description; "foo"; end
...
 
           context 'when the matcher does not define `description` (since it is an optional part of the protocol)' do
             it 'uses `inspect` in the error message instead' do
-              custom_matcher = Module.new do
-                def self.supports_block_expectations?; false; end
-                def self.inspect; "matcher-inspect"; end
+              RSpec::Matchers.define :custom_matcher_without_description do
+                def supports_block_expectations?; false; end
+                def inspect; "matcher-inspect"; end
               end
```
myronmarston: Why did you change this from a bare `Module` to matcher defined via `RSpec::Matchers.define`?

As the doc string states, this spec is meant to cover matchers that do not define `description`.  Matchers defined via `RSpec::Matchers.define` automatically define `description`.  You could undefine the method but it's simpler just to use a from-scratch object or module that avoids defining it to begin with.

spec/rspec/expectations/expectation_target_spec.rb
```diff
@@ -117,25 +117,27 @@ def self.description; "foo"; end
...
           it "uses the matcher's `description` in the error message" do
-            custom_matcher = Module.new do
-              def self.supports_block_expectations?; false; end
-              def self.description; "matcher-description"; end
+            RSpec::Matchers.define :custom_matcher_with_description do
+              def supports_block_expectations?; false; end
+
+              match { true }
+              description { "matcher-description" }
             end
```
myronmarston: Why did you switch from a module def to `RSpec::Matchers.define` here?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -155,7 +155,7 @@ def given_error
 
           backtrace = format_backtrace(@actual_error.backtrace)
           [
-            ", got #{@actual_error.inspect} with backtrace:",
+            ", got #{RSpec::Support::ObjectInspector.inspect(@actual_error)} with backtrace:",
```
myronmarston: Seeing the extra verbosity here makes me think we should change this matcher (and others where you had to do this) to subclass `RSpec::Matchers::BaseMatcher`.  The matchers that do not subclass that base class aren't intended to specifically avoid subclassing it; it's just that they didn't have a reason to subclass it before, but I think they do now.

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -141,7 +151,7 @@ module DefaultFailureMessages
           # you often only need to override `description`.
           # @return [String]
           def failure_message
-            "expected #{actual.inspect} to #{description}"
+            "expected #{RSpec::Support::ObjectInspector.inspect(@actual)} to #{description}"
```
myronmarston: Any reason not to use `actual_formatted` here?

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -150,7 +160,7 @@ def failure_message
           # you often only need to override `description`.
           # @return [String]
           def failure_message_when_negated
-            "expected #{actual.inspect} not to #{description}"
+            "expected #{RSpec::Support::ObjectInspector.inspect(@actual)} not to #{description}"
```
myronmarston: Any reason not to use `actual_formatted` here?

lib/rspec/matchers.rb
```diff
@@ -135,11 +135,11 @@ module RSpec
   #       end
   #
   #       def failure_message
-  #         "expected #{@target.inspect} to be in Zone #{@expected}"
+  #         "expected #{RSpec::Support::ObjectInspector.inspect(@target)} to be in Zone #{@expected}"
```
myronmarston: We've officially said that `RSpec::Support` is not part of RSpec's public API.  That may change at some point, but we're not yet ready to commit to maintaining its APIs according to SemVer policies.  Given that, I'd rather our docs not show it's use.

myronmarston:
Thanks for following up on this, @gavingmiller!  Your work here is very thorough, which is great.

> Is it sufficient to do replacement of inspect for ObjectInspector without writing new tests as long as the old tests don't break?

It's certainly over kill to write new specs for each matcher for this.  After all, if you're going to remember to write a spec for a particular matcher you're going to remember to update it to use the `ObjectInspector` and if you're going to forget to update the matcher, you're going to forget to write the spec...so it doesn't really have any benefit that I can see.

That said, I think there _is_ a real benefit to adding a spec to the [shared example group](https://github.com/rspec/rspec-expectations/blob/master/spec/support/shared_examples.rb) that we apply to all built-in matchers.  It enforces a bunch of things to ensure our matchers behave consistently, and we want our matchers to behave consistently with regard to the `ObjectInspector`.  Adding a spec there will help ensure all the built-in matchers use `ObjectInspector` as they should, and will ensure all future built-in matchers will likewise get the same treatment.

I played around with this a bit based on your branch and came up with this spec that can be added to that shared example group (probably at the bottom):

``` ruby
  it 'uses the `ObjectInspector` for `failure_message`' do
    allow(RSpec::Support::ObjectInspector).to receive(:inspect).and_return("detailed inspect")
    matcher.matches?(invalid_value)
    message = matcher.failure_message

    # Undo our stub so it doesn't affect the `include` matcher below.
    allow(RSpec::Support::ObjectInspector).to receive(:inspect).and_call_original
    expect(message).to include("detailed inspect")
  end
end
```

It's certainly not perfect -- stubbing `RSpec::Support::ObjectInspector` is a bit smelly compared to constructing the matchers with a `Time` object or something similar -- but given that there's no common instantiation interface for matchers (e.g. some take no args, some take 1, some take many...), this is the best I could come up with.  It basically just ensures that `failure_message` uses the inspector some how.  So it doesn't really verify the `ObjectInspector` is used _properly_, but it does verify that it is used at all, and I think there's a benefit in that.  Trying it out locally, there are a number of failures, and some of those (e.g. the ones for `respond_to` and `output`) were legitimate cases where `ObjectInspector` is not being used but should be.  So I think it serves a purpose in terms of highlighting those cases and forcing us to deal with them.  That said, there are some of the failures that I noticed that look like static (or nearly static) failure messages that aren't inspecting anything, and thus should properly not be using the `ObjectInspector`.  For those cases, we'll probably want to specifically opt-out from this spec using an option passed at the shared group inclusion site.

It's probably also worth looking into adding a similar spec for `failure_message_when_negated` and `description` as I'd expect those to commonly need to use `ObjectInspector` as well, although that might be more trouble than it's worth.  The `failure_message` check will at least enforce that every matcher uses the `ObjectInspector` somewhere.

Let me know if you need some more guidance than that (there are definitely some open questions here!) but that should be enough to move this along, hopefully.

myronmarston:
You mentioned in rspec/rspec-mocks#899 that you don't have the time to take that PR further.  Do you plan to take this one further?  I'm happy to pick it up from here if you don't have the time for this one, either.  Either way, just let us know.

Thanks!

gavingmiller:
Hey Myron - I won't be able to get this one over the finish line. My apologies for leaving these dangling on you.

myronmarston:
> My apologies for leaving these dangling on you.

No need to apologize!  Thanks for getting it this far.  RSpec is better because of the work you've done!

myronmarston:
Closing in favor of #765.

Issue #741, opened 2015-02-23
When `be_true` (via `true?`) or `be_false` (via `false?`) fail, mention their removal


myronmarston:
`be_true` and `be_false` were removed from RSpec 3, but they are still caught by the dynamic predicate matchers and translated into `true?` and `false?` calls, which fail and confuse users.  When users wrongly use them in RSpec 3 I think we should include a note that instructs them to use `be true` or `be truthy` similar to the deprecation we added in 2.99.

myronmarston:
Fixed by #744.

PR #744, opened 2015-02-26
Suggest `be_truthy`/`be true` when using `be_true`

imtayadeway:
Applies the same for `be_false`. Still fails as expected, but adds an
additional message to suggest an alternative if still using the
deprecated `be_true`/`be_false`.

Little unsure of the wording here - this most likely needs revision.

https://github.com/rspec/rspec-expectations/issues/741

### DIFF
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -264,7 +264,15 @@ module RSpec
           return nil if predicate_accessible?
 
           msg = "expected #{@actual} to respond to `#{predicate}`"
-          msg << " but `#{predicate}` is a private method" if private_predicate?
+
+          if private_predicate?
+            msg << " but `#{predicate}` is a private method"
+          elsif predicate == :true?
+            msg << " or perhaps you meant `be true` or `be_truthy`"
+          elsif predicate == :false?
+            msg << " or perhaps you meant `be false` or `be_falsey`"
+          end
+
           msg
         end
       end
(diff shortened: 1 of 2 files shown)


lib/rspec/matchers/built_in/be.rb
```diff
@@ -265,6 +265,8 @@ def validity_message
 
           msg = "expected #{@actual} to respond to `#{predicate}`"
           msg << " but `#{predicate}` is a private method" if private_predicate?
+          msg << ". Use `be true` or `be_truthy`" if predicate == :true?
+          msg << ". Use `be false` or `be_falsey`" if predicate == :false?
```
JonRowe: How about: 
> or perhaps you meant `be true` or `be truthy`

imtayadeway: Thanks! Just updated.

lib/rspec/matchers/built_in/be.rb
```diff
@@ -265,6 +265,8 @@ def validity_message
 
           msg = "expected #{@actual} to respond to `#{predicate}`"
           msg << " but `#{predicate}` is a private method" if private_predicate?
+          msg << " or perhaps you meant `be true` or `be_truthy`" if predicate == :true?
+          msg << " or perhaps you meant `be false` or `be_falsey`" if predicate == :false?
```
myronmarston: If `true?` or `false` is a private predicate would we want both the `but #true? is a private method` and the `or perhaps you meant "be true" or "be truthy"` parts?  I think thinking that we wouldn't.  Maybe these 3 conditionals should be made exclusive, like so?

``` ruby
if private_predicate?
  msg << ...
elsif predicate == :true?
  msg << ....
elsif predicate == :false?
  msg << ...
end
```

JonRowe: That works for me :)

imtayadeway: Just updated :)

JonRowe:
Thanks @imtayadeway

Issue #745, opened 2015-02-27
respond_to_missing? isn't defined in RSpec::Matchers


abotalov:
Currently `method_missing` is defined in RSpec::Matchers (https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers.rb#L919-L928) but `respond_to_missing?` isn't defined. It looks like because of it checking for `respond_to?` doesn't work.

myronmarston:
Huh, you're right.  I'm surprised we forgot to do that!

Want to work up a PR to add it?  The one suggestion I have is to be sure to define `respond_to?` on 1.8.7 rather than `respond_to_missing?` since `respond_to_missing?` wasn't around in 1.8.

abotalov:
Where should I put tests for it?

myronmarston:
Since it's missing from the `RSpec::Matchers` module, how about it's spec?

https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/matchers_spec.rb

PR #751, opened 2015-03-14
Make a context where RSpec::Matchers is included respond to missing be_* or have_*

abotalov:
Fixes https://github.com/rspec/rspec-expectations/issues/745

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -915,6 +915,7 @@ module RSpec
 
     BE_PREDICATE_REGEX = /^(be_(?:an?_)?)(.*)/
     HAS_REGEX = /^(?:have_)(.*)/
+    DYNAMIC_MATCHER_REGEX = Regexp.union(BE_PREDICATE_REGEX, HAS_REGEX)
 
     def method_missing(method, *args, &block)
       case method.to_s
@@ -925,6 +926,18 @@ module RSpec
       else
         super
       end
+    end
+
+    if RUBY_VERSION.to_f >= 1.9
+      def respond_to_missing?(method, *)
+        method =~ DYNAMIC_MATCHER_REGEX || super
+      end
+    else # for 1.8.7
+      def respond_to?(method, *)
+        method = method.to_s
+        method =~ DYNAMIC_MATCHER_REGEX || super
+      end
+      public :respond_to?
     end
 
     # @api private
--- a/spec/rspec/matchers_spec.rb
+++ b/spec/rspec/matchers_spec.rb
@@ -24,6 +24,22 @@ RSpec.describe RSpec::Matchers do
       end
     end
   end
+
+  describe "#respond_to?" do
+    it "handles dynamic matcher methods" do
+      expect(self).to respond_to(:be_happy, :have_eyes_closed)
+    end
+
+    it "supports the optional `include_private` arg" do
+      expect(self.respond_to?(:puts, true)).to eq true
+      expect(self.respond_to?(:puts, false)).to eq false
+      expect(self.respond_to?(:puts)).to eq false
+    end
+
+    it "allows `method` to get dynamic matcher methods", :if => RUBY_VERSION.to_f >= 1.9 do
+      expect(method(:be_happy).call).to be_a(be_happy.class)
+    end
+  end
 end
 
 module RSpec


lib/rspec/matchers.rb
```diff
@@ -915,6 +915,7 @@ def self.configuration
 
     BE_PREDICATE_REGEX = /^(be_(?:an?_)?)(.*)/
     HAS_REGEX = /^(?:have_)(.*)/
+    BE_OR_HAS_PREDICATE_REGEX = /^(?:be_(?:an?_)?|have_).*/
```
myronmarston: This could be simplified to `Regexp.union(BE_PREDICATE_REGEX, HAS_REGEX)`, which has the benefit of always being in sync with `BE_PREDICATE_REGEX` and `HAS_REGEX` when they change.

Also, I think `DYNAMIC_MATCHER_REGEX` is a better name for it, since in the context of `respond_to?` we really don't care which matchers will handle the message, just that one of the dynamic matchers will.

lib/rspec/matchers.rb
```diff
@@ -927,6 +928,18 @@ def method_missing(method, *args, &block)
       end
     end
 
+    if RUBY_VERSION.to_f >= 1.9
+      def respond_to_missing?(method, include_private=false)
+        method = method.to_s
```
myronmarston: The `to_s` here is unnecessary -- on 1.9+, symbols can be compared to regexes with `=~`.

spec/rspec/matchers_spec.rb
```diff
@@ -24,6 +24,16 @@
...
     end
   end
+
+  describe "#respond_to?" do
+    it "returns true for methods in example scope" do
+      expect(self.respond_to?(:be_happy)).to eq(true)
+    end
+
+    it "returns false for methods not defined in matcher or example scope" do
+      expect(self.respond_to?(:have_eyes_closed)).to eq(true)
```
myronmarston: There's a discrepancy between the doc string ("returns false....") and the expectation (`to eq(true)`).  Seems like you should move this expectation into the other example, and come up with a new message to check here.

spec/rspec/matchers_spec.rb
```diff
@@ -24,6 +24,16 @@
       end
     end
   end
+
+  describe "#respond_to?" do
+    it "returns true for methods in example scope" do
+      expect(self.respond_to?(:be_happy)).to eq(true)
```
myronmarston: This can use the `respond_to` matcher, which is particiularly useful when checking multiple messages (as I suggest below that you do here):

``` ruby
expect(self).to respond_to(:be_happy, :have_eyes_closed)
```

spec/rspec/matchers_spec.rb
```diff
@@ -24,6 +24,16 @@
       end
     end
   end
+
+  describe "#respond_to?" do
+    it "returns true for methods in example scope" do
```
myronmarston: This doc string isn't very communicative, IMO.  I think something like this would read better:

``` ruby
it "handles dynamic matcher methods"
```

spec/rspec/matchers_spec.rb
```diff
@@ -24,6 +24,16 @@
...
+      expect(self.respond_to?(:be_happy)).to eq(true)
+    end
+
+    it "returns false for methods not defined in matcher or example scope" do
+      expect(self.respond_to?(:have_eyes_closed)).to eq(true)
+    end
+  end
```
myronmarston: There are a couple pitfalls with `respond_to?` that are easy to fall into that I think we should have specs for:

* The 2nd arg (`include_private`) is easy to forget to support.  Would be nice to have a spec like:

``` ruby
it "supports the optional `include_private` arg" do
  expect(self.respond_to?(:puts, true)).to eq true
  expect(self.respond_to?(:puts, false)).to eq false
  expect(self.respond_to?(:puts)).to eq false
end
```

* If you define `respond_to?` on 1.9 instead of `respond_to_missing?`, the respond to checks you've done here would all pass.  The benefit of `respond_to_missing?` over `respond_to?` is it allows `method` to work, so a spec that requires that `respond_to_missing?` is used would be nice:

``` ruby
it "allows `method` to get dynamic matcher methods", :if => RUBY_VERSION.to_f >= 1.9 do
  expect(method(:be_happy).call).to be_a(be_happy.class)
end
```

(Of course, it's good to confirm that these specs fail as expected when you break the implementation).

lib/rspec/matchers.rb
```diff
@@ -927,6 +928,18 @@ def method_missing(method, *args, &block)
...
+        method =~ BE_OR_HAS_PREDICATE_REGEX || super
+      end
+    else # for 1.8.7
+      def respond_to?(method, include_private=false)
```
myronmarston: `respond_to?` must be public but you're in the section of this module under `private`.  OTOH, `respond_to_missing?` should be private (it's a callback rather than something meant to be called from outside the object, like `respond_to?` is), so my suggestion is just to add a `public :respond_to?` line after this method def to flip it back to public.

myronmarston:
Thanks, @abotalov!  Left a few comments but nothing major.

abotalov:
@myronmarston Thanks a lot about your comments! I attempted to fix them now.

myronmarston:
Thanks, @abotalov!

PR #755, opened 2015-03-17
Prevent infinite ranges blowing up matcher descriptions

### LINKED ISSUE #754: TypeError: "can't iterate from Float" when matching Hashes with Ranges
Failure/Error: expect({a: -Float::INFINITY..Float::INFINITY}).to eq({b: -Float::INFINITY..Float::INFINITY})
     TypeError:
       can't iterate from Float

JonRowe:
Prevent infinite ranges blowing up matcher descriptions, like IO it seems they're not always enumerable. Fixes #754.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,11 @@ Enhancements:
   `RSpec::Matchers#respond_to?` and `RSpec::Matchers#method` handle
   dynamic predicate matchers. (Andrei Botalov, #751)
 
+Bug Fixes:
+
+* Prevent `Range`s from being enumerated when generating matcher
+  descriptions. (Jon Rowe, #755)
+
 ### 3.2.0 / 2015-02-03
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.1.2...v3.2.0)
 
--- a/lib/rspec/matchers/composable.rb
+++ b/lib/rspec/matchers/composable.rb
@@ -103,7 +103,7 @@ module RSpec
           Hash[surface_descriptions_in(item.to_a)]
         elsif Struct === item
           item.inspect
-        elsif enumerable?(item)
+        elsif should_enumerate?(item)
           begin
             item.map { |subitem| surface_descriptions_in(subitem) }
           rescue IOError # STDOUT is enumerable but `map` raises an error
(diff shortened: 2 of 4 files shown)


lib/rspec/matchers/composable.rb
```diff
@@ -106,7 +106,9 @@ def surface_descriptions_in(item)
         elsif enumerable?(item)
           begin
             item.map { |subitem| surface_descriptions_in(subitem) }
-          rescue IOError # STDOUT is enumerable but `map` raises an error
+          # STDOUT is enumerable but `map` raises an error
+          # Range is enumerable except when infinite
+          rescue IOError, TypeError
```
myronmarston: Instead of rescueing `TypeError`, should we consider changing the definition of `enumerable?` so that ranges aren't enumerated?  After all, even for an enumerable range (e.g. all ints: `1..10`), I don't think we want it enumerated in the description.  We'd rather just have it inspected and displayed as `1..10`, right?

If we go that route, we should perhaps rename `enumerable?` to `should_enumerate?` or something.

JonRowe: Sure

JonRowe:
Updated

spec/rspec/matchers/composable_spec.rb
```diff
@@ -12,6 +12,19 @@ module Matchers
...
       end
 
+      it "does not blow up when surfacing descriptions from an unreadable Range object" do
+        infinite_range = -Float::INFINITY..Float::INFINITY
+        expect {
+          expect(1).to matcher_using_surface_descriptions_in(infinite_range)
+        }.to fail_with(infinite_range.inspect)
+      end
+
+      it "doesnt enumerate normal ranges" do
```
myronmarston: s/doesnt/doesn't/

Changelog.md
```diff
@@ -14,6 +14,11 @@ Enhancements:
   `RSpec::Matchers#respond_to?` and `RSpec::Matchers#method` handle
   dynamic predicate matchers. (Andrei Botalov, #751)
 
+Bug Fixes:
+
+* Exclude Ranges from being enumerable when generating matcher
```
myronmarston: Ranges are still `Enumerable` but we are choosing not to enumerate the.  Maybe "Exclude Ranges from being enumerated when generating matcher descriptions and failure messages" is better?

myronmarston:
LGTM.  Do you plan to backport this to 3.2?

JonRowe:
Travis doesnt like whatever I did, reopening PR

PR #756, opened 2015-03-17
Prevent infinite ranges blowing up matcher descriptions

JonRowe:
See #755, this is just for Travis.

PR #757, opened 2015-03-18
Fix descriptions with ranges kick travis

myronmarston:
Supersedes #756 (and #755).  I forced the commit SHA to be different so that hopefully travis will build it.

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -115,7 +115,7 @@ module RSpec
 
         def verify_message
           return true if @expected_message.nil?
-          values_match?(@expected_message, @actual_error.message)
+          values_match?(@expected_message, @actual_error.message.to_s)
         end
 
         def prevent_invalid_expectations
--- a/spec/rspec/matchers/composable_spec.rb
+++ b/spec/rspec/matchers/composable_spec.rb
@@ -3,13 +3,28 @@ module RSpec
     RSpec.describe Composable do
       RSpec::Matchers.define :matcher_using_surface_descriptions_in do |expected|
         match { false }
-        failure_message { surface_descriptions_in(expected) }
+        failure_message { surface_descriptions_in(expected).to_s }
       end
 
       it "does not blow up when surfacing descriptions from an unreadable IO object" do
         expect {
           expect(3).to matcher_using_surface_descriptions_in(STDOUT)
         }.to fail_with(STDOUT.inspect)
+      end
+
+      it "does not blow up when surfacing descriptions from an unreadable Range object" do
+        infinity = (1.0/0.0)
+        infinite_range = -infinity..infinity
+        expect {
+          expect(1).to matcher_using_surface_descriptions_in(infinite_range)
+        }.to fail_with(infinite_range.inspect)
+      end
+
+      it "does not enumerate normal ranges" do
+        range = 1..3
+        expect {
+          expect(1).to matcher_using_surface_descriptions_in(range)
+        }.to fail_with(range.inspect)
       end
 
       it "doesn't mangle struct descriptions" do
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,10 @@ Bug Fixes:
 
 * Make `contain_exactly` / `match_array` work with strict test doubles
   that have not defined `<=>`. (Myron Marston, #758)
+* Prevent `Range`s from being enumerated when generating matcher
+  descriptions. (Jon Rowe, #755)
+* Ensure exception messages are compared as strings in the `raise_error`
+  matcher. (Jon Rowe, #755)
 
 ### 3.2.0 / 2015-02-03
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.1.2...v3.2.0)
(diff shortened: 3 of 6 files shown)


myronmarston:
Now it's building.  Somehow this repo go turned off on travis.  I had to turn it back on to make it work and push once more.

myronmarston:
Hmm, looks like this fails on 1.9.2 and 1.8.7 :cry:

JonRowe:
So for some reason `.message` on error wasn't returning a string on 1.8.7. I don't even lol.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -46,6 +46,7 @@ def matches?(given_proc, negative_expectation=false, &block)
               @raised_expected_error = true
               @with_expected_message = verify_message
             end
+            true
```
myronmarston: I don't see what this line accomplishes.  This isn't a return statement, so this doesn't cause `true` to be returned.  There's no variable assignment on the `begin/end` expression that would get the value `true`.  As far as I can see, this does nothing...am I missing something?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -115,7 +116,7 @@ def eval_block
 
         def verify_message
           return true if @expected_message.nil?
-          values_match?(@expected_message, @actual_error.message)
+          values_match?(@expected_message, @actual_error.message.to_s)
```
myronmarston: So this change seems (to me at least) to be completely unrelated to the change of this PR.  I guess your change surfaced a bug in the `raise_error` matcher that this fixes?  If that's the case, can you add a spec to `raise_error_spec.rb` to specifically cover this behavior with an example rather than implicitly relying upon the composable specs?

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: Pass, wasn't intended to be comitted

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: I've added a spec, but FWIW it's `Composable` thats causing the problem which this surfaces, not setting the exception message to a string basically.

myronmarston: I see the problem now -- in `composable_spec.rb`, we have this:

``` ruby
RSpec::Matchers.define :matcher_using_surface_descriptions_in do |expected|
  match { false }
  failure_message { surface_descriptions_in(expected) }
end
```

`surface_descriptions_in` isn't necessarily meant to return a string (it's meant to return a data structure of the same shape), but the matcher protocol says `failure_message` should be a string.  So our custom matcher is written incorrectly.

That said, I wonder if we should add a `to_s` somewhere for situations like this...maybe [here](https://github.com/rspec/rspec-expectations/blob/f69cc918be1bdaccdb8f85877178c62d89439bdc/lib/rspec/expectations/fail_with.rb#L27)?

I'm fine with keeping the `raise_error` fix and spec as well -- it makes it a bit more robust, I think.

myronmarston:
This needs to be rebased before appveyor will build it again.

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: I've fixed the custom matcher, and let my raise_error changes, I'm not sure if adding a to_s where you suggest would also fix the issue, as the nested exception is a bit weird anyway, so I'm happy with this as is now

myronmarston: > I'm not sure if adding a to_s where you suggest would also fix the issue

It would...I tried it locally before suggesting it :).

JonRowe: Still seems like the wrong place, as it should already be a string before it gets there, if it was an object that wasn't an exception (and thus has a to_s of it's message) it wouldn't fix it.

myronmarston: It handles cases where the custom matcher's `failure_message` returns something besides a string.  That seems worthwhile to me, but it's not a big deal to me either way.

spec/rspec/matchers/composable_spec.rb
```diff
@@ -12,6 +12,21 @@ module Matchers
         }.to fail_with(STDOUT.inspect)
       end
 
+      it "does not blow up when surfacing descriptions from an unreadable Range object" do
+        infinity ||= (1.0/0.0)
```
myronmarston: in what circumstance would `infinity` already be set?  Isn't this equivalent to `infinity = (1.0/0.0)`?  Why use `||=`?

myronmarston:
LGTM, merge when the build is done.

spec/rspec/matchers/composable_spec.rb
JonRowe: Sorry another thing I didn't mean to commit, I was originally attempting to set the constant if it didn't exist (1.8.7 basically)

myronmarston:
Thoughts on backporting to 3.2?

JonRowe:
Yeah I think thats acceptable, done

PR #758, opened 2015-03-19
Fix `contain_exactly` to work with test doubles.

myronmarston:
Test doubles do not define `<=>` by default.

### DIFF
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -60,7 +60,9 @@ module RSpec
         end
 
         def safe_sort(array)
-          array.sort rescue array
+          array.sort
+        rescue Exception
+          array
         end
 
         def missing_items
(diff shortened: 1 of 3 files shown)


lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -60,7 +60,9 @@ def convert_actual_to_an_array
         end
 
         def safe_sort(array)
-          array.sort rescue array
+          array.sort
+        rescue Exception
```
JonRowe: Why does this fix it?

myronmarston: Because RSpec::Mocks::ExpectationNotMetError subclasses Exception, not StandardError.

JonRowe: But if the issue is that `<=>` isn't defined why is that causing that error and not `NoMethodError`?

myronmarston: Because that's what an rspec-mocks test double raises.

JonRowe:
LGTM

Issue #746, opened 2015-03-04
`include` diff can wrongly indicate some substrings are missing that are not


myronmarston:
Consider this spec:

``` ruby
specify "the diff indicates the substrings that are included are not" do
  words = "  foo\nbar\nbazz"
  expect(words).to include("foo", "bar") # passes
  expect(words).to include("foo", "bar", "gaz")
end
```

The 2nd expectation fails with:

```
     Failure/Error: expect(words).to include("foo", "bar", "gaz")
       expected "  foo\nbar\nbazz" to include "foo", "bar", and "gaz"
       Diff:

       @@ -1,4 +1,4 @@
       -foo
       +  foo
        bar
       -gaz
       +bazz
```

Notice that `foo` is shown in the diff to not match, and yet the string clearly includes `foo`, and as the first expectation shows, it is considered to match.  It's just the last `include` argument (`gaz`) that does not match. The diff is confusing here.

imtayadeway:
started to look into this last night. in rspec-support, i can verify the following:
```ruby
# spec/rspec/support/differ_spec.rb
it "outputs unified diff of a string and a collection of strings" do
  expected = ["foo", "bar", "gaz"]
  actual = " foo\nbar\nbazz"
  expected_diff = <<-'EOD'


@@ -1,4 +1,4 @@
-foo
+ foo
 bar
-gaz
+bazz
EOD

  diff = differ.diff(actual, expected)
  expect(diff).to be_diffed_as(expected_diff)
end
```
are we saying that the differ is behaving as expected here, given the above input, and that this is definitely an issue in rspec-expectations? if we can just clarify this, i'm happy to start tackling this :)

myronmarston:
@imtayadeway -- First off, I apologize for taking so long to get back to you.  The notification email has been sitting in my inbox for a while and I only now got around to responding.  Sorry about that.

Anyhow, I would say that it's not clear where the fault with this lies.  A line of `"foo"` is different than a line of `" foo"`, so it's correct that the differ highlights it as a difference.  And with a matcher like `eq`, it's proper that it be considered a difference.  The problem is that the differ doesn't understand the semantics of the matcher's fuzzy matching logic.  I have no ideas of what a good solution for this would be.

imtayadeway:
@myronmarston no apologies necessary! thanks for your thoughts - i'll try to dig into this this weekend.

imtayadeway:
@myronmarston having given this some more thought.... I'm coming back with more problems :)

[...]

PR #763, opened 2015-04-11
Avoid diffing for problematic `include` cases

imtayadeway:
Avoid diffing for problematic include cases.
    
Include#diffable? will return true unless it detects a multiline string with at least one "expected" that is included in "actual", but is not an exact match for any single line.

This was the most conservative strategy I could think of on this pass, and to my amazement it didn't cause any other tests to fail. (!)

I can squash these commits once this PR is ready, but didn't want to lose my first commit in case I needed that test again. It was by far the hardest part of the task so far!

Addresses https://github.com/rspec/rspec-expectations/issues/746

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -46,7 +46,7 @@ module RSpec
         # @api private
         # @return [Boolean]
         def diffable?
-          true
+          !diff_would_wrongly_highlight_matched_item?
         end
 
       private
@@ -99,6 +99,15 @@ module RSpec
           return false unless actual.respond_to?(:any?)
 
           actual.any? { |value| values_match?(expected_item, value) }
+        end
+
+        def diff_would_wrongly_highlight_matched_item?
+          return false unless actual.is_a?(String) && expected.is_a?(Array)
+
+          lines = actual.split("\n")
+          expected.any? do |str|
+            actual.include?(str) && lines.none? { |line| line == str }
+          end
         end
       end
     end
--- a/spec/rspec/matchers/built_in/include_spec.rb
+++ b/spec/rspec/matchers/built_in/include_spec.rb
@@ -51,6 +51,12 @@ RSpec.describe "#include matcher" do
           expect("abc\ndef").to include("g", "h")
         }.to fail_matching("expected \"abc\\ndef\" to include \"g\" and \"h\"\nDiff")
       end
+
+      it "does not diff when lines match but are not an exact match" do
+        expect {
+          expect(" foo\nbar\nbazz").to include("foo", "bar", "gaz")
+        }.to fail_with(a_string_not_matching(/Diff/i))
+      end
     end
 
     context "for an array target" do
@@ -578,4 +584,3 @@ RSpec.describe "#include matcher" do
     raise_error(RSpec::Expectations::ExpectationNotMetError, /#{Regexp.escape(message)}/)
   end
 end
-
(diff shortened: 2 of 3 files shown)


spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -51,6 +51,14 @@
...
         }.to fail_matching("expected \"abc\\ndef\" to include \"g\" and \"h\"\nDiff")
       end
+
+      it "does not diff when lines match but are not an exact match" do
+        expect {
+          expect(" foo\nbar\nbazz").to include("foo", "bar", "gaz")
+        }.to raise_error(RSpec::Expectations::ExpectationNotMetError) { |error|
+          expect(error.message).not_to match(/Diff/)
+        }
+      end
```
myronmarston: This expectation can be simplified to:

``` ruby
expect {
  expect(" foo\nbar\nbazz").to include("foo", "bar", "gaz")
}.to fail_with(a_string_excluding("Diff"))
```

* `fail_with` is defined in [fail_matchers](https://github.com/rspec/rspec-expectations/blob/f0b79d1e58bcc209191bfc99a04211dd38718955/lib/rspec/matchers/fail_matchers.rb#L21-L28)
* `a_string_excluding` is defined as the negation of `a_string_including`  (which is itself an alias of `include`) -- see [here](https://github.com/rspec/rspec-expectations/blob/f0b79d1e58bcc209191bfc99a04211dd38718955/spec/support/matchers.rb#L42)

lib/rspec/matchers/built_in/include.rb
```diff
@@ -46,7 +46,12 @@ def failure_message_when_negated
         # @api private
         # @return [Boolean]
         def diffable?
-          true
+          return true unless actual.is_a?(String) && expected.is_a?(Array)
+
+          lines = actual.split("\n")
+          expected.none? do |str|
+            actual.include?(str) && lines.none? { |line| line == str }
+          end
```
myronmarston: :+1: I love how simple this is.

 I think it could be worth extracting this logic into a method that describes the what this logic does -- something like `diff_would_wrongly_highlight_matched_item?` -- and then `diffable?` could be defined as `!diff_would_wrongly_highlight_matched_item?`.

Thoughts?

myronmarston:
Well done!  I'll be happy to merge this once the feedback above is addressed.  Please squash it as well after addressing that.

spec/rspec/matchers/built_in/include_spec.rb
imtayadeway: Definitely agree this could (should) be simpler. It's a remnant from when I was trying to examine the diff and the strange characters it contains (I"ve since seen that there are better ways of doing that, too). Do you think it's OK to use `a_string_excluding` here, for the same reason as not using `fail_including` (as indicated at the bottom of this file)?

lib/rspec/matchers/built_in/include.rb
imtayadeway: Great suggestion!

spec/rspec/matchers/built_in/include_spec.rb
myronmarston: Good point!  I think maybe this will work:

``` ruby
RSpec::Matchers.define_negated_matcher :a_string_not_matching, :match

expect {
  expect(" foo\nbar\nbazz").to include("foo", "bar", "gaz")
}.to fail_with(a_string_not_matching(/Diff/i))
```

(The negated matcher def should probably go in `spec/support/matchers.rb`)

imtayadeway:
@myronmarston should be good to go now. thanks for the great suggestions! always a pleasure :)

myronmarston:
Thanks @imtayadeway!

PR #768, opened 2015-04-12
Warn about bare `raise_error`

### LINKED ISSUE #655: Suggestion: raise_error matcher should have no default exception type or default to StandardError
Using the `raise_error` matcher without specifying a particular error type is quite dangerous, as it  swallows any sort of exception, including syntax errors and rspec expectation errors. It's hard for me to imagine a reasonable justification for the author of a test not caring at all what type of error is thrown; in my opinion it would be more helpful to users of rspec to be forced to think at least a little bit about what type of error they are expecting.

If there's no agreement on that, then at least perhaps it might be worth considering changing the default from `Exception` to `StandardError`.

For quick reference:
https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/raise_error.rb#L11
https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/raise_error.rb#L42-L49

JonRowe:
Issues a warning (which can be supressed) when using a bare `raise_error` call. Fixes #655.

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -8,10 +8,16 @@ module RSpec
       class RaiseError
         include Composable
 
-        def initialize(expected_error_or_message=Exception, expected_message=nil, &block)
+        def initialize(expected_error_or_message=nil, expected_message=nil, &block)
           @block = block
           @actual_error = nil
+          @warn_about_bare_error =
+            RSpec::Expectations.configuration.warn_about_false_positives? &&
+            expected_error_or_message.nil?
+
           case expected_error_or_message
+          when nil
+            @expected_error, @expected_message = Exception, expected_message
           when String, Regexp
             @expected_error, @expected_message = Exception, expected_error_or_message
           else
@@ -23,6 +29,7 @@ module RSpec
         # Specifies the expected error message.
         def with_message(expected_message)
           raise_message_already_set if @expected_message
+          @warn_about_bare_error = false
           @expected_message = expected_message
           self
         end
@@ -36,6 +43,17 @@ module RSpec
           @with_expected_message = false
           @eval_block = false
           @eval_block_passed = false
+
+          if @warn_about_bare_error && !negative_expectation && @block.nil?
+            RSpec.warning("Using the `raise_error` matcher without providing a specific " \
+                          "error or message risks false positives, since `raise_error` " \
+                          "will match when Ruby raises a `NoMethodError`, `NameError` or " \
+                          "`ArgumentError`, potentially allowing the expectation to pass " \
+                          "without even executing the method you are intending to call. " \
+                          "Instead consider providing a specific error class or message. " \
+                          "This message can be supressed by setting: " \
+                          "`RSpec::Expectations.configuration.warn_about_false_positives = false`")
+          end
 
           return false unless Proc === given_proc
 
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -44,17 +44,7 @@ module RSpec
           @eval_block = false
           @eval_block_passed = false
 
-          if warning_about_bare_error && !negative_expectation
-            RSpec.warning("Using the `raise_error` matcher without providing a specific " \
-                          "error or message risks false positives, since `raise_error` " \
-                          "will match when Ruby raises a `NoMethodError`, `NameError` or " \
-                          "`ArgumentError`, potentially allowing the expectation to pass " \
-                          "without even executing the method you are intending to call. " \
-                          "Instead consider providing a specific error class or message. " \
-                          "This message can be supressed by setting: " \
-                          "`RSpec::Expectations.configuration.warn_about_false_positives = false`")
-          end
-
+          warn_about_bare_error if warning_about_bare_error && !negative_expectation
           return false unless Proc === given_proc
 
           begin
@@ -155,6 +145,17 @@ module RSpec
 
         def warning_about_bare_error
           @warn_about_bare_error && @block.nil?
+        end
+
+        def warn_about_bare_error
+          RSpec.warning("Using the `raise_error` matcher without providing a specific " \
+                        "error or message risks false positives, since `raise_error` " \
+                        "will match when Ruby raises a `NoMethodError`, `NameError` or " \
+                        "`ArgumentError`, potentially allowing the expectation to pass " \
+                        "without even executing the method you are intending to call. " \
+                        "Instead consider providing a specific error class or message. " \
+                        "This message can be supressed by setting: " \
+                        "`RSpec::Expectations.configuration.warn_about_false_positives = false`")
         end
 
         def expected_error
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -44,7 +44,7 @@ module RSpec
           @eval_block = false
           @eval_block_passed = false
 
-          if @warn_about_bare_error && !negative_expectation && @block.nil?
+          if warning_about_bare_error && !negative_expectation
             RSpec.warning("Using the `raise_error` matcher without providing a specific " \
                           "error or message risks false positives, since `raise_error` " \
                           "will match when Ruby raises a `NoMethodError`, `NameError` or " \
@@ -66,9 +66,7 @@ module RSpec
             end
           end
 
-          unless negative_expectation
-            eval_block if @raised_expected_error && @with_expected_message && @block
-          end
+          eval_block if !negative_expectation && ready_to_eval_block?
 
           expectation_matched?
         end
@@ -121,6 +119,10 @@ module RSpec
           @eval_block ? @eval_block_passed : true
         end
 
+        def ready_to_eval_block?
+          @raised_expected_error && @with_expected_message && @block
+        end
+
         def eval_block
           @eval_block = true
           begin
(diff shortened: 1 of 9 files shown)


lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -8,10 +8,25 @@ module BuiltIn
...
+        # Configures the bare `raise_error` warning that would otherwise be
+        # issued when using the matcher without a specific error or message.
+        def self.allow_bare_raise_error=(value)
+          @allow_bare_raise_error = value
+        end
+
+        # @private
+        def self.allow_bare_raise_error?
+          @allow_bare_raise_error ||= false
+        end
```
myronmarston: We have all our configuration exposed off of `RSpec::Expectations.configuration`, so it seems odd to have this other config here.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -37,6 +53,13 @@ def matches?(given_proc, negative_expectation=false, &block)
           @eval_block = false
           @eval_block_passed = false
 
+          if @warn_about_bare_error && !negative_expectation && @block.nil?
+            RSpec.warning("Using the `raise_error` matcher without providing a specific " \
+                          "error or message risks false positives, consider providing " \
+                          "a specific error instead. This message can be suppressed using " \
+                          "`RSpec::Matchers::BuiltIn::RaiseError.allow_bare_raise_error = true`")
+          end
```
myronmarston: It's clear to us why there's a risk of false positives, but I'm not sure it'll be clear to the user.  Might it be worth fleshing that out a bit?  Something like:

> ...since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call.

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -1,11 +1,22 @@
...
+
+  it "issues a warning when used without an error class or message" do
+    expect_warning_with_call_site __FILE__, __LINE__+1, /without providing a specific error/
+    expect { raise }.to raise_error
+  end
+
+  it "can supresses the warning when configured to do so" do
+    RSpec::Matchers::BuiltIn::RaiseError.allow_bare_raise_error = true
+    expect { raise }.to raise_error
+    RSpec::Matchers::BuiltIn::RaiseError.allow_bare_raise_error = false
```
myronmarston: This could leak the mutated config if the example fails before it hits the last line that sets the value back to `true`.

myronmarston: Is it worth having some examples specifically for the cases that do not issue a warning?  e.g.:

``` ruby
it 'does not issue a warning when an exception class is specified (even if it is just `Exception`)'
it 'does not issue a warning when a message is specified'
it 'does not issue a warning when a block is passed'
```

myronmarston:
So I'm realizing that this is going a different direction then we did for the issue of `expect { }.not_to raise_error(something_specific)` being prone to false positives -- there we raise an error.  It seems odd to raise for one form that's prone to false positives but warn for another.

I'm also realizing that we may have other places in the future where we realize that a particular use of matcher is prone to false positives and should be discouraged.

Given that, I'm thinking maybe we do the following:

* Move the config for this off of `RaiseError` and onto `RSpec::Expectations.configuration`, as I suggested below.
* Rename it to `warn_about_possible_false_positives` and use it for this and for any other future cases where there's a potential false positive situation.
* Update the `raise_error` logic for the negative case to align with this -- it should either warn (if the config is set) or allow the potential false positive.

Thoughts?

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: I was on the fence about where this should live, given it's very specific it seemed slightly odd to have a config option for it, but I'm happy to move the config there.

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: Yeah I was hoping someone would chip in some better copy :)

JonRowe:
Yup, agree

JonRowe:
This is changed as discussed, I'll tackle changing `raise_error` for the negative case in a separate PR

JonRowe:
Woop fixed Rubocop

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -37,6 +44,17 @@ def matches?(given_proc, negative_expectation=false, &block)
...
+          if warning_about_bare_error && !negative_expectation
+            RSpec.warning("Using the `raise_error` matcher without providing a specific " \
+                          "error or message risks false positives, since `raise_error` " \
+                          "will match when Ruby raises a `NoMethodError`, `NameError` or " \
+                          "`ArgumentError`, potentially allowing the expectation to pass " \
+                          "without even executing the method you are intending to call. " \
+                          "Instead consider providing a specific error class or message. " \
+                          "This message can be supressed by setting: " \
+                          "`RSpec::Expectations.configuration.warn_about_false_positives = false`")
+          end
```
myronmarston: I think these 10 lines could benefit from being extracted into a helper method.  It's a lot to have in-lined into `matches?`.

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -1,11 +1,38 @@
...
+
+  it "can supresses the warning when configured to do so" do
+    original_value = RSpec::Expectations.configuration.warn_about_false_positives?
+    begin
+      RSpec::Expectations.configuration.warn_about_false_positives = false
+      expect { raise }.to raise_error
+    ensure
+      RSpec::Expectations.configuration.warn_about_false_positives = original_value
+    end
+  end
```
myronmarston: Should the config toggling be moved into a shared context that handles that detail, like these?

https://github.com/rspec/rspec-expectations/blob/a75fe7bfc48788c724e51fb6e1b4581b76376e19/spec/spec_helper.rb#L53-L99

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -1,11 +1,38 @@
...
+  it 'does not issue a warning when an exception class is specified (even if it is just `Exception`)' do
+    expect { raise "error" }.to raise_error Exception
+  end
+
+  it 'does not issue a warning when a message is specified' do
+    expect { raise "error" }.to raise_error "error"
+  end
+
+  it 'does not issue a warning when a block is passed' do
+    expect { raise "error" }.to raise_error { |_| }
```
myronmarston: These examples all state that no warning is issued but there are no expectations to that effect.  I understand that our rspec-support warning tracking will fail the example if there is one but I think these examples would benefit from a more explicit expectation.  I expect us to always keep the warning prevention logic but there's no guarantee it will always work the way it is -- for example, maybe we'll refactor it to do a single end-of-run check or something rather than checking in each example...and if we change it, these examples, as they currently exist, would not verify what they claim to verify.

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: Done.

spec/rspec/matchers/built_in/raise_error_spec.rb
JonRowe: Yeah my first instinct was to add one, but we don't have the helper! So I got lazy ;)

spec/rspec/matchers/built_in/raise_error_spec.rb
JonRowe: Done

JonRowe:
This should be good to go now?

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -1,11 +1,36 @@
...
-    expect {raise}.to raise_error
+    expect { raise "error" }.to raise_error "error"
+  end
+
+  it "issues a warning when used without an error class or message" do
+    expect_warning_with_call_site __FILE__, __LINE__+1, /without providing a specific error/
+    expect { raise }.to raise_error
+  end
+
+  it "can supresses the warning when configured to do so", :warn_about_false_positives do
```
myronmarston: The suppression of the warning via the config is the point of this example, so it seems like it should be front and center in the example rather then hidden in a shared context.  I was expecting the shared context to be used to guarantee the config value is reset back to what it started at, but not to actually make the state change itself.

spec/spec_helper.rb
```diff
@@ -94,6 +94,13 @@ def hash_inspect(hash)
   end
 end
 
+RSpec.shared_context "with warn_about_false_positives set to false", :warn_about_false_positives do
+  original_value = RSpec::Expectations.configuration.warn_about_false_positives?
+
+  before(:context) { RSpec::Expectations.configuration.warn_about_false_positives = false }
+  after(:context)  { RSpec::Expectations.configuration.warn_about_false_positives = original_value }
+end
```
myronmarston: As mentioned above, I would expect this to sandbox the change rather than actually making the change...

myronmarston:
Left a couple comments but they aren't super important.  Feel free to merge w/o addressing them.

spec/spec_helper.rb
JonRowe: Changed :P

myronmarston:
LGTM, merge when green.

PR #769, opened 2015-04-14
Emphasize expect syntax in generated descriptions

ChrisArcand:
Especially as the new expect syntax is recommended over the old should syntax,
changes the ExpectationHandlers to use 'to expect to' rather than
'should' as their auto-generated description's verb when using one-liner syntax and expect syntax is enabled. When expect syntax is disabled, uses the old 'should' descriptions.

I think this more appropriately reflects the expect syntax in the output - regardless of the distinction that this is indeed at least `RSpec::Core::ExampleGroup#should` and not the very unrecommended `Object#should`. 

To be paired with https://github.com/rspec/rspec-core/pull/1936.

Looks like you have the rspec-core tests (which are fixed in the other PR) running here in Travis, too - so the 7 failures here match the other PR as intended.

### DIFF
--- a/features/custom_matchers/define_matcher.feature
+++ b/features/custom_matchers/define_matcher.feature
@@ -35,8 +35,8 @@ Feature: define matcher
     When I run `rspec ./matcher_with_default_message_spec.rb --format documentation`
     Then the exit status should not be 0
 
-    And the output should contain "should be a multiple of 3"
-    And the output should contain "should not be a multiple of 4"
+    And the output should contain "is expected to be a multiple of 3"
+    And the output should contain "is expected to not be a multiple of 4"
     And the output should contain "Failure/Error: it { is_expected.to be_a_multiple_of(4) }"
     And the output should contain "Failure/Error: it { is_expected.not_to be_a_multiple_of(3) }"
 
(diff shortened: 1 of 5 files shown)


myronmarston:
Thanks for the PR, @ChrisArcand.  You did a great job on this PR.

I'm on the fence about if we want this change, though.  We added `is_expected` in 3.0 to offer an API that aligns more closely with the `expect` syntax and to hopefully reduce confusion around the one-liner `should` syntax.  Aligning the generated descriptions with that seems like a generally good idea.  OTOH, I'm not sure that these generated descriptions are actually any better than the `should`-based ones.  Compare these:

```
User
  should require a value for `email`
```

vs.

```
User
  is expected to require a value for `email`
```

"is expected" seems more noisy than "should" and I'm not sure it's actually an improvement in the doc output.  But I can see the value in aligning with `is_expected`.

What do others think?

/cc @rspec/rspec

samphippen:
Thanks for the pr @ChrisArcand. The code changes are great.

I agree that is expected to is more noisy. I can see the value of making the documentation aligned, but I think practically speaking it won't change how people approach the expect syntax. I also think there might be people parsing the output in it's current form, so changing it might break stuff. I'm very very mildly against this change.

ChrisArcand:
I pondered a bit on the readability difference too, but figured I'd see if the aligning of the generated descriptions to the matchers was of value anyway; I also dug around a bit looking for a way to customize the message based on the actual matcher used (`should` vs `expect.to`) instead of just generically using `is expected to` if the syntax is at all enabled, but I couldn't find a way to do it without larger, unreasonable changes. 

TBH I prefer formatting descriptions without using _either_ 'should' or 'is expected to', (and it looks like you agree, @myronmarston)[https://github.com/andreareginato/betterspecs/issues/15], but I don't know if that's really possible in auto-generated descriptions :)

In any case, thanks! I appreciate both of your input and understand the mild hesitation. :+1: I'll close it up.

JonRowe:
I was mildly in favour of this FWIW ;)

myronmarston:
If enough people vote in favor of this we may revisit :).

PR #772, opened 2015-04-15
Fix match bug

myronmarston:


lib/rspec/matchers/built_in/match.rb
```diff
@@ -21,7 +21,14 @@ def diffable?
 
         def match(expected, actual)
           return true if values_match?(expected, actual)
-          actual.match(expected) if actual.respond_to?(:match)
+          actual.match(expected) if can_safely_call_match?(expected, actual)
```
JonRowe: If the intent here is to be falsey if we can't safely call match I think this would read better as a guard condition, e.g.

```Ruby
return false unless can_safely_call_match?(expected, actual)
actual.match(expected)
```

JonRowe:
LGTM bar a quibble over form

lib/rspec/matchers/built_in/match.rb
myronmarston: Done.

JonRowe:
:+1:

PR #773, opened 2015-04-16
Generating diffs of large objects consumes enough CPU to freeze the average computer

JonRowe:
When certain expectations fail RSpec will try to generate diff of the expected and received values. Usually this works fine, except when those values are 6000 line long HTML documents parsed using Nokogiri (or any other operation using big objects). In such a case generating a diff will consume enough CPU usage to at least freeze my Macbook Pro, requiring a hard reboot to make it usable again.

Repro: https://gist.github.com/YorickPeterse/ff2919edb411d9b2689e

Save the script somewhere and run it using either `ruby script.rb` or `rspec script.rb`. If you're using a laptop you'll want to take it off your lap and plug in the charger first.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ platforms :rbx do
   gem 'rubysl'
 end
 
+gem 'nokogiri'
 gem 'rubocop', "~> 0.23.0", :platform => [:ruby_19, :ruby_20, :ruby_21]
 
 eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
--- a/spec/boom_spec.rb
+++ b/spec/boom_spec.rb
@@ -1,0 +1,7213 @@
+require 'rspec'
+require 'rspec/autorun'
+require 'nokogiri'
+
+class SomeClass
+  def example(doc); end
+end
+
+# Warning: running this _will_ freeze your computer or at the very least make it
+# very slow, consume a lot of CPU and burn your lap (in case you're using a
+# laptop).
+RSpec.describe 'Diffing large documents' do
+  before :all do
+    html  = DATA.read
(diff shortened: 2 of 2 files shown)


JonRowe:
Let's see what Travis does..

YorickPeterse:
I'm not sure how that would be related. A colleague of mine ran the script on OS X and it produced the same pattern: it hangs while sucking up 100% CPU. They're also using Nokogiri 1.6.6.2.

JonRowe:
> I'm not sure how that would be related. A colleague of mine ran the script on OS X and it produced the same pattern: it hangs while sucking up 100% CPU. They're also using Nokogiri 1.6.6.2.

From memory, nokogiri bundles libxml2 in certain situations where it would be unavailable (e.g. OSX) but on linux it uses the platform library. This is the one key difference I can think of between OSX and linux that is causing this to fail for you.

JonRowe:
Hmm, no theres more to it. The autorun example actually doesn't blow up for me, but inserting it into RSpecs own spec suite does.

JonRowe:
Converting the nokogiri objects to string manually actually causes the spec to pass, converting them to their inspect output causes the spec to fail with a diff correctly.

JonRowe:
So I can recreate this with `PP.pp @doc_1`, so it's actually not strictly an RSpec issue (its also code in rspec-support not rspec-expectations FWIW @rspec/rspec) pondering what to do about this, if an object can't be safely coerced into a string using pretty print...

JonRowe:
So the nokogiri document when pretty printed comes out closer to 50000 lines, and it's possible to replicate this standalong, closing in favour of rspec/rspec-support#196

PR #774, opened 2015-04-17
Stop assuming the format of `test_double.inspect`.

myronmarston:
This is necessary for rspec/rspec-mocks#932.


myronmarston:
Merging so the other PR can go green.

Issue #733, opened 2015-02-12
Feature idea: `aggregate_failures`


myronmarston:
When you've got multiple independent expectations to make about a particular result, there are generally two routes people take:

``` ruby
RSpec.describe MyThing do
  it "does the thing and makes many expectations about it" do
    result = MyThing.do_it

    expect(result).to blah
    expect(result).to blah_blah
    expect(result).to have_other_interesting_properties
  end
end

# or

RSpec.describe MyThing do
  let(:result) { MyThing.do_it }
  specify { expect(result).to blah }
  specify { expect(result).to blah_blah }
  specify { expect(result).to have_other_interesting_properties }
end
```

Nether is perfectly ideal -- the latter gives you failure output for each expectation that failed, whereas the former runs faster, particularly if `MyThing.do_it` is slow.  People ask if RSpec supports a better solution for this from time to time (the most recent example is [on the mailing list](https://groups.google.com/forum/#!topic/rspec/RFRA5ckSols)).

One solution is available in rspec-given: it's [And API](https://github.com/rspec-given/rspec-given#and).  That's not going to work in vanilla RSpec, of course, but it got me thinking about this, and I've come up with an idea that I think would work quite well: a new block API like `aggregate_failures`.  Here's how it would work:

``` ruby
RSpec.describe MyThing do
  it "does the thing and makes many expectations about it" do
    result = MyThing.do_it

    aggregate_failures do
      expect(result).to blah
      expect(result).to blah_blah
      expect(result).to have_other_interesting_properties
    end
  end
end
```

The idea is that you can wrap any set of expectations with it, and it aggregates all the expectation failures that occur in the block, and, if any occur, raises a single exception at the end of the block containing all the failure messages.

Implementation wise, I've thought of a super simple way to do this: change [fail_with](https://github.com/rspec/rspec-expectations/blob/v3.2.0/lib/rspec/expectations/fail_with.rb#L29) so that rather than calling `raise` directly, it instead invokes `RSpec::Expectations.configuration.failure_notifier.call(failure)`.  We'd implement `failure_notifier` so that it uses a thread local for storage and defaults to `Kernel.raise`.  `aggregate_failures` would then simply change the `failure_notifier` for the current thread to an alternate implementation that aggregates the failures rather than raising immediately, yield, change it back, and then raise with all the failure if there are any.

I'm also thinking that we could automatically make this available via metadata as well:

``` ruby
RSpec.describe MyThing do
[...]

PR #776, opened 2015-04-19
Aggregate failures

myronmarston:
This is the beginning of my solution for #733.

TODOs:

- [x] Finish cuke.
- [x] Ensure this works well with our minitest integration (we reassign the `ExpectationNotMetError` constant there which could cause problems for the new exception class since it is a subclass).
- [x] Update docs to mention how this interacts with threading.
- [x] Update rspec-core to integrate with this.
- [x] Update rspec-mocks to use the rspec-support failure notifier so that mock failures can participate in this.
- [x] Figure out how this interacts with using an expectation from within a DSL-defined custom matcher.

### DIFF
--- a/features/aggregating_failures.feature
+++ b/features/aggregating_failures.feature
@@ -4,6 +4,8 @@ Feature: Aggregating Failures
 
   `aggregate_failures` takes an optional string argument that will be used in the aggregated failure message as a label.
 
+  RSpec::Core expands this feature a bit; see [the rspec-core docs](../../rspec-core/docs/expectation-framework-integration/aggregating-failures) for more detail.
+
   Note: The implementation of `aggregate_failures` uses a thread-local variable, which means that if you have an expectation failure in another thread, it'll abort like normal.
 
   Scenario: Multiple expectation failures within `aggregate_failures` are all reported
--- a/features/aggregating_failures.feature
+++ b/features/aggregating_failures.feature
@@ -1,97 +1,48 @@
-@wip @announce
 Feature: Aggregating Failures
 
-  Normally, an expectation failure causes the example to immediately bail.  Sometimes, however, you have multiple, independent expectations, and you'd like to be able to see all of the failures rather than just the first.  One solution is to split off a separate example for each expectation, but if the setup for the examples is slow, that's going to take extra time and slow things down. `aggregate_failures` provides an alternate solution. Within the block, expectation failures will not abort the example like normal; instead, the failures will be aggregated into a single exception that is raised at the end of the block, allowing you to see all expectations that failed.
+  Normally, an expectation failure causes the example to immediately abort.  When you have multiple independent expectations, it's nice to be able to see all of the failures rather than just the first.  One solution is to split off a separate example for each expectation, but if the setup for the examples is slow, that's going to take extra time and slow things down.  `aggregate_failures` provides an alternate solution.  It wraps a set of expectations with a block.  Within the block, expectation failures will not immediatly abort like normal; instead, the failures will be aggregated into a single exception that is raised at the end of the block, allowing you to see all expectations that failed.
 
   `aggregate_failures` takes an optional string argument that will be used in the aggregated failure message as a label.
 
-  Scenario: Multiple Expectation Failures in the same example are all reported
+  Note: The implementation of `aggregate_failures` uses a thread-local variable, which means that if you have an expectation failure in another thread, it'll abort like normal.
+
+  Scenario: Multiple expectation failures within `aggregate_failures` are all reported
     Given a file named "spec/aggregated_failure_spec.rb" with:
       """ruby
-      RSpec.describe "An aggregated failure" do
-        it "lists each of the individual failures in the failure output" do
-          aggregate_failures "testing equality" do
-            expect(1).to eq(2)
-            expect(2).to eq(3)
-            expect(3).to eq(4)
-          end
+      require 'rspec/expectations'
+      include RSpec::Matchers
+
+      Response = Struct.new(:status, :headers, :body)
+      response = Response.new(404, { "Content-Type" => "text/plain" }, "Not Found")
+
+      begin
+        aggregate_failures "testing response" do
+          expect(response.status).to eq(200)
+          expect(response.headers["Content-Type"]).to eq("application/json")
+          expect(response.body).to eq('{"message":"Success"}')
         end
+      rescue RSpec::Expectations::MultipleExpectationsNotMetError => e
+        puts e.message
+        exit(1)
       end
       """
-    When I run `rspec spec/aggregated_failure_spec.rb`
-    Then it should fail listing all the failures:
+    When I run `ruby spec/aggregated_failure_spec.rb`
+    Then it should fail with:
       """
-      Failures:
+      Got 3 failures from failure aggregation block "testing response":
 
-        1) An aggregated failure lists each of the individual failures in the failure output
-           Got 3 failures from failure aggregation block "testing equality".
-           # ./spec/aggregated_failure_spec.rb:3:in `block (2 levels) in <top (required)>'
+        1) expected: 200
+                got: 404
 
-           1.1) Failure/Error: expect(1).to eq(2)
+           (compared using ==)
 
-                  expected: 2
-                       got: 1
+        2) expected: "application/json"
+                got: "text/plain"
 
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:4:in `block (3 levels) in <top (required)>'
+           (compared using ==)
 
-           1.2) Failure/Error: expect(2).to eq(3)
+        3) expected: "{\"message\":\"Success\"}"
+                got: "Not Found"
 
-                  expected: 3
-                       got: 2
-
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:5:in `block (3 levels) in <top (required)>'
-
-           1.3) Failure/Error: expect(3).to eq(4)
-
-                  expected: 4
-                       got: 3
-
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:6:in `block (3 levels) in <top (required)>'
-      """
-
-  Scenario: Use `:aggregated_failures` metadata
-    Given a file named "spec/aggregated_failure_spec.rb" with:
-      """ruby
-      RSpec.describe "An aggregated failure" do
-        it "lists each of the individual failures in the failure output", :aggregate_failures do
-          expect(1).to eq(2)
-          expect(2).to eq(3)
-          expect(3).to eq(4)
-        end
-      end
-      """
-    When I run `rspec spec/aggregated_failure_spec.rb`
-    Then it should fail listing all the failures:
-      """
-      Failures:
-
-        1) An aggregated failure lists each of the individual failures in the failure output
-           Got 3 failures:
-
-           1.1) Failure/Error: expect(1).to eq(2)
-
-                  expected: 2
-                       got: 1
-
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:3:in `block (3 levels) in <top (required)>'
-
-           1.2) Failure/Error: expect(2).to eq(3)
-
-                  expected: 3
-                       got: 2
-
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:4:in `block (3 levels) in <top (required)>'
-
-           1.3) Failure/Error: expect(3).to eq(4)
-
-                  expected: 4
-                       got: 3
-
-                  (compared using ==)
-                # ./spec/aggregated_failure_spec.rb:5:in `block (3 levels) in <top (required)>'
+           (compared using ==)
       """
--- a/features/aggregating_failures.feature
+++ b/features/aggregating_failures.feature
@@ -1,0 +1,97 @@
+@wip @announce
+Feature: Aggregating Failures
+
+  Normally, an expectation failure causes the example to immediately bail.  Sometimes, however, you have multiple, independent expectations, and you'd like to be able to see all of the failures rather than just the first.  One solution is to split off a separate example for each expectation, but if the setup for the examples is slow, that's going to take extra time and slow things down. `aggregate_failures` provides an alternate solution. Within the block, expectation failures will not abort the example like normal; instead, the failures will be aggregated into a single exception that is raised at the end of the block, allowing you to see all expectations that failed.
+
+  `aggregate_failures` takes an optional string argument that will be used in the aggregated failure message as a label.
+
+  Scenario: Multiple Expectation Failures in the same example are all reported
+    Given a file named "spec/aggregated_failure_spec.rb" with:
+      """ruby
+      RSpec.describe "An aggregated failure" do
+        it "lists each of the individual failures in the failure output" do
+          aggregate_failures "testing equality" do
+            expect(1).to eq(2)
+            expect(2).to eq(3)
+            expect(3).to eq(4)
+          end
+        end
+      end
+      """
+    When I run `rspec spec/aggregated_failure_spec.rb`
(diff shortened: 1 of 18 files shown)


lib/rspec/matchers/fail_matchers.rb
```diff
@@ -14,8 +14,8 @@ module FailMatchers
       #
       # @example
       #   expect { some_expectation }.to fail
-      def fail
-        raise_error(RSpec::Expectations::ExpectationNotMetError)
+      def fail(&block)
+        raise_error(RSpec::Expectations::ExpectationNotMetError, &block)
```
JonRowe: Isn't this in support too?

myronmarston: No.  Why would it be?  "fail" in rspec-expectations refers to a `RSpec::Expectations::ExpectationNotMetError` whereas in rspec-mocks it refers to a `RSpec::Mocks::MockExpectationError` so there's no common meaning of "fail" across libs.

JonRowe: Ah of course :)

myronmarston:
OK, this PR is ready for review.  There are a couple open questions:

* What should `aggregate_failures` do when there is a single expectation failure, or not expectation failures but some other error?  At the moment it allows the exception in either case to propogate it as-is rather than wrapping it in a `MultipleExpectationsNotMetError`.  That means that the additional failure output for aggregation failure blocks (e.g. `Got x failures....`) only display when there are in fact multiple expectation failures (or a single expectation failure and some other exception).  I an see an argument for keeping the interface more uniform and having it always raise a `MultipleExpectationsNotMetError`, though.

* I'm struggling to figure out the best way to handle the cuke.  Currently it's labeled `@wip` because the formatting logic (which it relies upon, because it shows how it is formatted nicely in the output) is in rspec/rspec-core#1946.  That worries me, because if we ever decide to change that formatting in rspec-core, it would break the cuke here which could be annoying to deal with.  OTOH, I'm not really sure how to write a documentation-worthy cuke scenario without including the nicely formatted rspec-core failure message, unless we write the cuke so it's example code runs outside the context of rspec-core...but that would be kinda weird since all the other cuke scenarios use rspec-core (except the specific minitest one...)  I'm not sure what a good solution is here.  There's also the `:aggregate_failures` metadata, which is covered by the cuke here as well but is really an rspec-core thing that leverages this feature.

/cc @rspec/rspec (especially @xaviershay since you said you could do some code reviews for this stuff)

features/aggregating_failures.feature
```diff
@@ -0,0 +1,97 @@
+@wip @announce
+Feature: Aggregating Failures
+
+  Normally, an expectation failure causes the example to immediately bail.  Sometimes, however, you have multiple, independent expectations, and you'd like to be able to see all of the failures rather than just the first.  One solution is to split off a separate example for each expectation, but if the setup for the examples is slow, that's going to take extra time and slow things down. `aggregate_failures` provides an alternate solution. Within the block, expectation failures will not abort the example like normal; instead, the failures will be aggregated into a single exception that is raised at the end of the block, allowing you to see all expectations that failed.
```
xaviershay: first reaction to "bail" was it's too informal ... I wonder if non-native speakers know what it means. Maybe "return" is better? "exit"?

xaviershay: Too many commas in close proximity in second sentence.

features/aggregating_failures.feature
```diff
@@ -0,0 +1,97 @@
...
+
+  `aggregate_failures` takes an optional string argument that will be used in the aggregated failure message as a label.
+
+  Scenario: Multiple Expectation Failures in the same example are all reported
```
xaviershay: do we anticipate one form being "better" than the other? Should we recommend one? And if so, that should be presented first.

Feels weird (particularly for beginners) to have two basically equivalent versions of the same thing. Alternatively, the examples should demonstrate when you would use one versus the other.

Do we need both forms?

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -0,0 +1,159 @@
...
+    class FailureAggregator
+      attr_reader :block_label, :metadata
+
+      def aggregate
+        RSpec::Support.with_failure_notifier(self) do
+          begin
+            yield
+          rescue ExpectationNotMetError => e
+            failures << e
+          rescue Exception => e
```
xaviershay: I would expect this to be flagged by rubocop. Should comment why explicit `Exception` is the right class here, since it feels pretty "deep" in the stack.

If I have N long running aggregated failures, am I still able to ctrl+c effectively?

xaviershay: Or if not, how would I ever get more than one `other_error`?

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -0,0 +1,159 @@
...
+        @block_label = block_label
+        @metadata    = metadata
+      end
+
+      def raise_aggregated_failures
+        all_errors = failures + other_errors
+
+        case all_errors.size
+        when 0 then return nil
+        when 1 then raise all_errors.first
```
xaviershay: I'm ok with this behaviour of only raise multiple if there are indeed more than one.

lib/rspec/expectations/minitest_integration.rb
```diff
@@ -7,6 +7,19 @@ def expect(*a, &b)
     assert(true) # so each expectation gets counted in minitest's assertion stats
     super
   end
+
+  # Convert a `MultipleExpectationsNotMetError` to an `Minitest::Assertion` error so
```
xaviershay: `s/an/a/`

lib/rspec/matchers.rb
```diff
@@ -277,6 +277,37 @@ def self.define_negated_matcher(negated_name, base_name, &description_override)
       alias_matcher(negated_name, base_name, :klass => AliasedNegatedMatcher, &description_override)
     end
 
+    # Allows multiple expectations in the provided block to fail, and then
+    # aggregates them into a single exception, rather than bailing on the
```
xaviershay: if you change "bail" above, also change it here.

lib/rspec/matchers.rb
```diff
@@ -277,6 +277,37 @@ def self.define_negated_matcher(negated_name, base_name, &description_override)
...
+    # @raise [Expectations::ExpectationNotMetError] raised when a single
+    #   expectation fails.
+    # @raise [Exception] other sorts of exceptions will be raised as normal.
+    #
+    # @example
+    #   aggregate_failures("verifying response") do
+    #     expect(response.status).to eq(200)
+    #     expect(response.headers).to include("Content-Type" => "text/plain")
+    #     expect(response.body).to include("Success")
+    #   end
```
xaviershay: this strikes me as a much better example than the equality one in the cukes. I know it's not as runnable, but maybe we could do something similar. Maybe "testing equality" checks `==`/`eql?`/`hash` or something?

spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -0,0 +1,323 @@
...
+          )],
+          :other_errors => [error]
+        ))
+      end
+
+      it "includes the error in the raised aggregated failure when an expectation failed as well" do
+        expect_error_included_in_aggregated_failure StandardError.new("boom")
+      end
+
+      it "handles direct `Exceptions` and not just `StandardError` and descendents" do
```
xaviershay: this could be an alternate place to document the "why" of this.

lib/rspec/expectations/failure_aggregator.rb
xaviershay: hrmm, I actually don't understand yet how even the normal `ExpectationNotMetError` "resumes" after the first failure. Gonna go read the rspec-core PR, maybe there is something in there.

xaviershay:
Missing changelog.

lib/rspec/expectations/failure_aggregator.rb
xaviershay: nope, still don't get it. I guess this is supported by already-written code somewhere.

myronmarston: > Should comment why explicit Exception is the right class here, since it feels pretty "deep" in the stack.

Good idea; will do.

> If I have N long running aggregated failures, am I still able to ctrl+c effectively?

Assuming you're using rspec-core or some other test framework that handles that signal gracefully, yes.  rspec-core traps that interrupt and gracefully shuts down.  Since it traps the interrupt it doesn't raise an exception anymore.  I believe cucumber, minitest, etc handle ctrl-c in similar fashion.

> Or if not, how would I ever get more than one `other_error`?

You can't ever get more than one in `other_errors`.  I thought about exposing it as a scalar instead of an array, but I thought that it was better to have a uniform interface and provide both types of errors as arrays.  Plus who knows...maybe someone day we'll figure out a way to support multiple.  It's nice to make the interface an array so that we don't have to change the API if/when we later support that.

> hrmm, I actually don't understand yet how even the normal ExpectationNotMetError "resumes" after the first failure. 

It doesn't resume.  Expectation failures use `RSpec::Support.notify_failure`, and get added to the array via `call`, defined below.  Problem is, that uses a thread-local variable, which means that if you spawn a thread and use an expectation in it, it'll raise instead of invoking `call`, below.  So this is here just to handle expectation failures in threads.    The `ExpectationNotMetError` rescue is necessary for these specs to pass:

https://github.com/rspec/rspec-expectations/blob/9ad61a35b059893ff84b158fe56897e2cbb34be1/spec/rspec/expectations/failure_aggregator_spec.rb#L149-L171

Your questions here are good, and made me realize a potential improvement we could make.  Users will probably assume that the entire `aggregate_failures` block ran to completion but if an expectation failure happened in a thread or some other error was raised, it will have aborted early.  Given that...I'm thinking we should perhaps surface that to the user somehow.  Although, I want to get this wrapped up, so maybe we'll save that for a future enhancement unless you think it's essential to the feature.

features/aggregating_failures.feature
myronmarston: Thanks, I'll rewrite this.  Do you think that "abort" is better than "bail"?

features/aggregating_failures.feature
myronmarston: I think we need both.  In most tests that do one action and than may want to assert multiple things, wrapping the entire test works well and is really convenient.  However, in an end-to-end acceptance test, you can have multiple actions (e.g. simulating a user clicking through your app and doing stuff) and you won't want later steps to run if something failed at an earlier step...but _within_ a step it's nice to be able to assert multiple things.

You're right that this isn't explained well; I'll improve it...although, I'm still not sure where the cukes belong given that it sits at the intersection of core and expectations.

On a side note, I'm thinking about recommending to the cucumber team that in their rspec-expectations integration they may want to wrap every step in an `aggregation_failures` block.

lib/rspec/matchers.rb
myronmarston: I could perhaps make it runnable by making `response` a dumb struct, that isn't actually a real HTTP response.

myronmarston:
Thanks for the detailed review, @xaviershay.  One bit of context that may help is rspec/rspec-support#194, which this builds on top of.

lib/rspec/expectations/failure_aggregator.rb
xaviershay: ah, that makes sense. Maybe include some of that in a comment? Seems non-obvious.

Don't think any code changes are needed here though - seems like an edge case. If an exception is raise I think it's reasonable to expect we didn't run all the expectations. And the thread case is pretty weird, I can't imagine people hitting that often.

features/aggregating_failures.feature
xaviershay: abort is good

features/aggregating_failures.feature
xaviershay: No opinion about where cukes should live :/ It's a weird split.

features/aggregating_failures.feature
JonRowe: :+1: for abort although I like bail I think it's too informal too

lib/rspec/expectations/failure_aggregator.rb
JonRowe: :+1: for a comment for this

features/aggregating_failures.feature
abotalov: > On a side note, I'm thinking about recommending to the cucumber team that in their rspec-expectations integration they may want to wrap every step in an aggregation_failures block.

If a cuke scenario is very end-to-end, it can do so many things in a step that aggregating failures at step level won't make sense.

myronmarston: > If a cuke scenario is very end-to-end, it can do so many things in a step that aggregating failures at step level won't make sense.

Maybe.  After this is released I'll probably reach out to the cucumber team and let them know the feature is available and could be used to wrap each step and it's up to them to decide if that wold help meet user needs or not.

lib/rspec/expectations/failure_aggregator.rb
myronmarston: I added comments -- let me know what you think.

lib/rspec/expectations/minitest_integration.rb
myronmarston: Thanks, addressed.

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -0,0 +1,171 @@
...
+            # `other_errors`.
+            failures << e
+          rescue Exception => e
+            # Normally we don't want to rescue `Exception` but here it's important.
+            # If we don't, it would cause expectation failures that happened before now to
+            # be silenced...but a failing precondition expressed as an expectation could point
+            # to the root cause of the exception we're rescuing. So it's important to include
+            # both in the raised error. `notify_aggregated_failures` will ensure that either
+            # just this exception (if there are no others) or a `MultipleExpectationsNotMetError`
+            # is raised as appropriate, so this exception will still be surfaced to the user.
```
JonRowe: :+1:

xaviershay: I don't think this covers why `Exception` and not `StandardError`? i.e. the same comment would apply if the code read `rescue => e`. That was the original thing that set my spidey sense off.

myronmarston: > I don't think this covers why Exception and not StandardError? i.e. the same comment would apply if the code read rescue => e. That was the original thing that set my spidey sense off.

I reworded it....want to take a look again?

myronmarston:
OK, I pushed some more updates.  I think my changes here are done with the exception of a link I'm planning to add from the cuke here to the cuke I'm going to write in rspec-core.

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -0,0 +1,172 @@
...
+          rescue Exception => e
+            # While it is normally a bad practice to rescue `Exception`, it's important we do
+            # so here. It's low risk (`notify_aggregated_failures` below will re-raise the exception,
+            # or raise a `MultipleExpectationsNotMetError` that includes the exception), and it's
+            # essential that the user is notified of expectation failures that may have already
+            # occurred in the `aggregate_failures` block. Those expectation failures may provide
+            # important diagnostics for understanding why this exception occurred, and if simply
+            # allowed this exception to be raised as-is, it would (wrongly) suggest to the user
+            # that the expectation passed when it did not, which would be quite confusing.
+            other_errors << e
```
xaviershay: very good. Ship it!

myronmarston:
OK, I think I've addressed all the feedback.  The updated cuke is [viewable here](https://relishapp.com/rspec-staging/rspec-expectations/docs/aggregating-failures).

Are you satisfied with the cuke now, @xaviershay?  Anything else needed before I merge?

xaviershay:
ship it!

PR #775, opened 2015-04-21
Change error for `expect { }.not_to raise_error(something_specific)` to warn and respect new config

JonRowe:
Follow up to #768.  @JonRowe said he planned to do this but I wanted to open an issue so we don't forget.

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -11,9 +11,7 @@ module RSpec
         def initialize(expected_error_or_message=nil, expected_message=nil, &block)
           @block = block
           @actual_error = nil
-          @warn_about_bare_error =
-            RSpec::Expectations.configuration.warn_about_potential_false_positives? &&
-            expected_error_or_message.nil?
+          @warn_about_bare_error = warn_about_potential_false_positives? && expected_error_or_message.nil?
 
           case expected_error_or_message
           when nil
@@ -64,7 +62,7 @@ module RSpec
 
         # @private
         def does_not_match?(given_proc)
-          prevent_invalid_expectations
+          warn_for_false_positives
           !matches?(given_proc, :negative_expectation) && Proc === given_proc
         end
 
@@ -128,19 +126,23 @@ module RSpec
           values_match?(@expected_message, @actual_error.message.to_s)
         end
 
-        def prevent_invalid_expectations
-          what_to_raise = if expecting_specific_exception? && @expected_message
-                            "`expect { }.not_to raise_error(SpecificErrorClass, message)`"
-                          elsif expecting_specific_exception?
-                            "`expect { }.not_to raise_error(SpecificErrorClass)`"
-                          elsif @expected_message
-                            "`expect { }.not_to raise_error(message)`"
-                          end
+        def warn_for_false_positives
+          return unless warn_about_potential_false_positives?
+          expression = if expecting_specific_exception? && @expected_message
+                         "`expect { }.not_to raise_error(SpecificErrorClass, message)`"
+                       elsif expecting_specific_exception?
+                         "`expect { }.not_to raise_error(SpecificErrorClass)`"
+                       elsif @expected_message
+                         "`expect { }.not_to raise_error(message)`"
+                       end
 
-          return unless what_to_raise
+          return unless expression
 
-          specific_class_error = ArgumentError.new("#{what_to_raise} is not valid, use `expect { }.not_to raise_error` (with no args) instead")
-          raise specific_class_error
+          warn_about_negative_false_positive expression
+        end
+
+        def warn_about_potential_false_positives?
+          RSpec::Expectations.configuration.warn_about_potential_false_positives?
         end
 
         def warning_about_bare_error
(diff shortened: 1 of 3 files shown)


JonRowe:
Ping @myronmarston

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -148,14 +152,22 @@ def warning_about_bare_error
...
+        end
+
+        def warn_about expression, alternative
+          warning = "This will potentially allowing the expectation to pass without" \
+                    "without even executing the method you are intending to call. "
+
+          instructions = "This message can be supressed by setting: " \
+                         "`RSpec::Expectations.configuration.warn_about_potential_false_positives = false`"
+
+          RSpec.warning(expression + warning + alternative + instructions)
```
JonRowe: I felt this was about the right level of repetition for the messages whilst still not having two blocks of text, but I'm willing to be countered on this.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -128,19 +128,23 @@ def verify_message
...
+          return unless type
 
-          specific_class_error = ArgumentError.new("#{what_to_raise} is not valid, use `expect { }.not_to raise_error` (with no args) instead")
-          raise specific_class_error
+          expression = "Using `#{type}` risks false positives, since `#{type}` " \
+                       "won't match when Ruby raises a different error (e.g. " \
+                       "`NoMethodError`, `NameError` or `ArgumentError`). "
+          alternative = "Instead consider using `expect {}.not_to raise_error`. "
+
+          warn_about expression, alternative if RSpec::Expectations.configuration.warn_about_potential_false_positives?
```
myronmarston: Seems like the `RSpec::Expectations.configuration.warn_about_potential_false_positives?` check could go on the first line of the method and avoid any work if it's `false`, right?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -148,14 +152,22 @@ def warning_about_bare_error
...
+            "Using the `raise_error` matcher without providing a specific " \
+            "error or message risks false positives, since `raise_error` " \
+            "will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`. ",
+            "Instead consider providing a specific error class or message. "
+          )
+        end
+
+        def warn_about(expression, alternative)
+          warning = "This will potentially allowing the expectation to pass without" \
+                    "without even executing the method you are intending to call. "
```
myronmarston: I think this warning would be better without two withouts :).

lib/rspec/matchers/built_in/raise_error.rb
myronmarston: Personally, I think it's incidental duplication: just because there is some common text now, doesn't mean there will always be.  We may want to reword one message and the common helper method would probably get in the way.  Plus it's harder to get the sense for what the entire message will be when its broken up like this.

Do you mind defactoring it inspite of the textual duplication?

JonRowe:
Ok I defactored it

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -128,19 +126,23 @@ def verify_message
...
-          what_to_raise = if expecting_specific_exception? && @expected_message
-                            "`expect { }.not_to raise_error(SpecificErrorClass, message)`"
-                          elsif expecting_specific_exception?
-                            "`expect { }.not_to raise_error(SpecificErrorClass)`"
-                          elsif @expected_message
-                            "`expect { }.not_to raise_error(message)`"
-                          end
+        def warn_about_invalid_expectations
+          return unless warn_about_potential_false_positives?
+          type = if expecting_specific_exception? && @expected_message
```
myronmarston: I think `type` is a bad name for this.  WDTY of `expression` instead?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -158,6 +160,16 @@ def warn_about_bare_error
                         "`RSpec::Expectations.configuration.warn_about_potential_false_positives = false`")
         end
 
+        def warn_about_negative_false_positive(type)
+          RSpec.warning("Using `#{type}` risks false positives, since `#{type}` " \
+                       "won't match when Ruby raises a different error (e.g. " \
+                       "`NoMethodError`, `NameError` or `ArgumentError`), potentially " \
+                        "allowing the expectation to pass without even executing the " \
+                        "method you are intending to call. Instead consider using " \
```
myronmarston: I think the language about "won't match when Ruby raises a different error" is a little confusing.  How about this instead?

> Using `#{expression}` risks false positives, since literally _any other error_ would cause the
> expectation to pass, including those raised by Ruby (e.g. `NoMethodError`, `NameError` and `ArgumentError`),
> meaning the code you are intending to test may not even get reached.  Instead consider... (the rest of your message)

JonRowe: Updated

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -64,7 +62,7 @@ def matches?(given_proc, negative_expectation=false, &block)
 
         # @private
         def does_not_match?(given_proc)
-          prevent_invalid_expectations
+          warn_about_invalid_expectations
```
myronmarston: I'm thinking "invalid" is the right term to use here anymore since we allow it and it has to do with false positives...

myronmarston:
The comment I left above isn't a merge blocker.

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: Appveyor needed a kick anyway and I didn't feel that strongly about it so changed it ;)

myronmarston:
:+1:

PR #780, opened 2015-04-21
Show only differing results in #include

### LINKED ISSUE #771: `include` matcher failure message does not specify which items were not included
``` ruby
     Failure/Error: expect("abc").to include("a", "d", "c")
       expected "abc" to include "a", "d", and "c"
```

In this example, "a" and "c" were included but not "d".  The failure message doesn't specify which were included and which weren't, though.  In this case, that's not a huge deal since you can see the letters in the string, but if the subject's `inspect` output didn't indicate what it includes, you would have no way of knowing.

We should improve the failure message.

ChrisArcand:
Fixes #771 

Reimplements the `include` matcher to account for all divergent items (excluded when expected to be included in actual or included when expected to be excluded in actual) to be used in example failure output. Now when something is missing from an expectation of inclusion, the failure output will more helpfully tell you what exactly was missing rather than a full description of the expected items.

Most basic example:
```plaintext
# Current output
Failure/Error: expect("abc").to include("a", "d", "c")
       expected "abc" to include "a", "d", and "c"

Failure/Error: expect({ :a => 7, :b => 5 }).not_to include(:a => 7, :d => 3)
       expected {:a => 7, :b => 5} not to include {:a => 7, :d => 3}

# Revised output with this PR
Failure/Error: expect("abc").to include("a", "d", "c")
       expected "abc" to include "d"
       # "a" and "c" were indeed included, so only "d" is mentioned

Failure/Error: expect({ :a => 7, :b => 5 }).not_to include(:a => 7, :d => 3)
       expected {:a => 7, :b => 5} not to include {:a => 7}
       # :d => 3 was indeed not included, so only :a => 7 is mentioned
```

~~My ramblings on performance in the associated issue are addressed in this implementation; my first [poorly written] attempt at this prompted my confused comment but I believe this one works quite well and I benchmarked it extensively. It has little to no impact~~

Includes benchmarks; a [potentially] large hit in failures for includes is expected; instead of fulfilling a failed match and stopping execution, we now continue comparing the rest of `expected` to `actual` for the new failure output.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -12,35 +12,31 @@ module RSpec
         # @api private
         # @return [Boolean]
         def matches?(actual)
-          @actual = actual
-          perform_match(:all?, :all?)
+          perform_match(actual) { |v| v }
         end
 
         # @api private
         # @return [Boolean]
         def does_not_match?(actual)
-          @actual = actual
-          perform_match(:none?, :any?)
+          perform_match(actual) { |v| !v }
         end
 
         # @api private
         # @return [String]
         def description
-          described_items = surface_descriptions_in(expected)
-          item_list = EnglishPhrasing.list(described_items)
-          improve_hash_formatting "include#{item_list}"
+          improve_hash_formatting("include#{readable_list_of(expected)}")
         end
 
         # @api private
         # @return [String]
         def failure_message
-          improve_hash_formatting(super) + invalid_object_message
+          format_failure_message("to") { super }
         end
 
         # @api private
         # @return [String]
         def failure_message_when_negated
-          improve_hash_formatting(super) + invalid_object_message
+          format_failure_message("not to") { super }
         end
 
         # @api private
@@ -51,24 +47,43 @@ module RSpec
 
       private
 
-        def invalid_object_message
-          return '' if actual.respond_to?(:include?)
-          ", but it does not respond to `include?`"
+        def format_failure_message(preposition)
+          if actual.respond_to?(:include?)
+            improve_hash_formatting("expected #{description_of @actual} #{preposition} include#{readable_list_of @divergent_items}")
+          else
+            improve_hash_formatting(yield) + ", but it does not respond to `include?`"
+          end
         end
 
-        def perform_match(predicate, hash_subset_predicate)
-          return false unless actual.respond_to?(:include?)
+        def readable_list_of(items)
+          described_items = surface_descriptions_in(items)
+          if described_items.all? { |item| item.is_a?(Hash) }
+            " #{described_items.inject(:merge).inspect}"
+          else
+            EnglishPhrasing.list(described_items)
+          end
+        end
 
-          expected.__send__(predicate) do |expected_item|
+        def perform_match(actual, &block)
+          @actual = actual
+          @divergent_items = excluded_from_actual(&block)
+          actual.respond_to?(:include?) && @divergent_items.empty?
+        end
+
+        def excluded_from_actual
+          return [] unless @actual.respond_to?(:include?)
+
+          expected.inject([]) do |memo, expected_item|
             if comparing_hash_to_a_subset?(expected_item)
-              expected_item.__send__(hash_subset_predicate) do |(key, value)|
-                actual_hash_includes?(key, value)
+              expected_item.each do |(key, value)|
+                memo << { key => value } unless yield actual_hash_includes?(key, value)
               end
             elsif comparing_hash_keys?(expected_item)
-              actual_hash_has_key?(expected_item)
+              memo << expected_item unless yield actual_hash_has_key?(expected_item)
             else
-              actual_collection_includes?(expected_item)
+              memo << expected_item unless yield actual_collection_includes?(expected_item)
             end
+            memo
           end
         end
 
--- a/benchmarks/include_matcher.rb
+++ b/benchmarks/include_matcher.rb
@@ -1,0 +1,250 @@
+require 'benchmark/ips'
+require 'rspec/expectations'
+
+include RSpec::Matchers
+
+module RSpec
+  module Matchers
+    module BuiltIn
+      class OldInclude < BaseMatcher
+        def initialize(*expected)
+          @expected = expected
+        end
+
+        def matches?(actual)
+          @actual = actual
+          perform_match(:all?, :all?)
+        end
+
+        def does_not_match?(actual)
+          @actual = actual
+          perform_match(:none?, :any?)
+        end
+
+        def description
+          described_items = surface_descriptions_in(expected)
+          item_list = EnglishPhrasing.list(described_items)
+          improve_hash_formatting "include#{item_list}"
+        end
+
+        def failure_message
+          improve_hash_formatting(super) + invalid_object_message
+        end
+
+        def failure_message_when_negated
+          improve_hash_formatting(super) + invalid_object_message
+        end
+
+        def diffable?
+          !diff_would_wrongly_highlight_matched_item?
+        end
+
+      private
+
+        def invalid_object_message
+          return '' if actual.respond_to?(:include?)
+          ", but it does not respond to `include?`"
+        end
+
+        def perform_match(predicate, hash_subset_predicate)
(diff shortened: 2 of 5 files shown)


myronmarston:
> I believe this one works quite well and I benchmarked it extensively. It has little to no impact :+1:

Do you mind including your benchmarks?  We like to keep benchmarks around in `benchmarks`.  If you do that, a couple suggestions over the old benchmark you had:

* Use `benchmark-ips` over `benchmark` -- the former provides better output and makes comparisons easier.  (You'll have to add it to `Gemfile-custom`).
* Your benchmark before was using the matcher in an rspec-core example.  This causes rspec-core's overhead to be included in the benchmark.  IMO it's better to benchmark just the matcher w/o rspec-core.  You can do that by including `RSpec::Matchers` in a custom context (e.g. a class you define for this purpose) and then write an expectation in that context (you'll have to rescue `RSpec::Expectations::ExpectationNotMetError` for the failing case).
* When you use `benchmark-ips` you'll need to have both implementations available in the same process (rather than running a benchmark against 2 different commits).  I tend to do that by copying the original implementation into the file and renaming it so that you can directly compare it against the new implementation.

With the benchmark in the repo it'll give us the context of the benchmark results and allow anyone to run them.

myronmarston:
BTW, ignore the rubocop failures for now.  I have some feedback that will likely address class/method length.  We can revisit the rubocop failures later.

lib/rspec/matchers/built_in/include.rb
```diff
@@ -13,34 +13,50 @@ def initialize(*expected)
         # @return [Boolean]
         def matches?(actual)
           @actual = actual
-          perform_match(:all?, :all?)
+          if actual.respond_to?(:include?)
+            @divergent_items = excluded_from_actual
```
myronmarston: The fact that you only assign this _sometimes_ means this matcher is potentially prone to a bug:

``` ruby
include_one = include(1)
expect([1, 2, 3]).to include_one
expect(Object.new).to include_one
```

On the 2nd use of the matcher, `@divergent_items` will still have the value from the first use of the matcher.  It looks like there isn't actually a bug from this yet since your use of `@divergent_items` in the failure message methods is guarded by a `respond_to` check, but it's a bug waiting to happen if/when that ivar got used for something else.  I'd prefer to see it always assigned.

In addition, the `if actual.respond_to?(:include)` here and for `does_not_match?` is duplicated.  I think we can kill two birds with one stone:

``` ruby
def matches?(actual)
  @actual = actual
  @divergent_items = excluded_from_actual
  @divergent_items.empty?
end

def does_not_match?(actual)
  @actual = actual
  @divergent_items = included_from_actual
  @divergent_items.empty?
end

def excluded_from_actual(...)
  return [] unless actual.respond_to?(:include?)
  # ...
end
```

Thoughts?

ChrisArcand:
> Do you mind including your benchmarks?
Sure, I'll rewrite/add some in the way you describe. 

> BTW, ignore the rubocop failures for now.
Sounds good, feedback welcome. There's obviously some duplication in my implementation but I couldn't think of a good way to eliminate that effectively without making things seemingly too complex :/

Looks like I might have something else somewhere causing different Ruby versions to fail, too.

WIP :)

lib/rspec/matchers/built_in/include.rb
```diff
@@ -13,34 +13,50 @@ def initialize(*expected)
...
+          end
         end
 
         # @api private
         # @return [Boolean]
         def does_not_match?(actual)
           @actual = actual
-          perform_match(:none?, :any?)
+          if actual.respond_to?(:include?)
+            @divergent_items = included_from_actual
```
myronmarston: I think `included_in_actual` reads better than `included_from_actual`.

lib/rspec/matchers/built_in/include.rb
myronmarston: Woops, realized my suggestion above would break things.  Updated:

``` ruby
def matches?(actual)
  @actual = actual
  @divergent_items = excluded_from_actual
  actual.respond_to?(:include) && @divergent_items.empty?
end

def does_not_match?(actual)
  @actual = actual
  @divergent_items = included_from_actual
  actual.respond_to?(:include) && @divergent_items.empty?
end

def excluded_from_actual(...)
  return [] unless actual.respond_to?(:include?)
  # ...
end

Comment (ChrisArcand) [inline lib/rspec/matchers/built_in/include.rb]: I like it :+1:

File: lib/rspec/matchers/built_in/include.rb [hunk 3]
```diff
@@ -13,34 +13,50 @@ def initialize(*expected)
...
 
         # @api private
         # @return [String]
         def failure_message
-          improve_hash_formatting(super) + invalid_object_message
+          if actual.respond_to?(:include?)
+            improve_hash_formatting("expected #{description_of @actual} to include#{readable_list_of @divergent_items}")
+          else
+            improve_hash_formatting(super) + invalid_object_message
+          end
```
ChrisArcand: TBH I am not a fan of this (as well as it's duplication with the `when_negated` version). Any thoughts on how to clean/combine this up?

lib/rspec/matchers/built_in/include.rb
```diff
@@ -51,27 +67,50 @@ def diffable?
...
+              end
             else
-              actual_collection_includes?(expected_item)
+              if options[:invert] == true
+                memo << expected_item if actual_collection_includes?(expected_item)
+              else
+                memo << expected_item unless actual_collection_includes?(expected_item)
+              end
             end
+            memo
```
myronmarston: This method checks the same conditional 3 times and, IMO, is crying out for a more polymorphic approach.  Here's how we can do that, I think:

``` ruby
class Include
  def matches?(actual)
    perform_match(actual) { |v| v }
  end

  def does_not_match?(actual)
    perform_match(actual) { |v| !v }
  end

private

  def perform_match(actual, &block)
    @actual = actual
    @divergent_items = excluded_from_actual(&block)
    actual.respond_to?(:include?) && @divergent_items.include?
  end

  def excluded_from_actual
    return [] unless @actual.respond_to?(:include?)

    expected.inject([]) do |memo, expected_item|
      if comparing_hash_to_a_subset?(expected_item)
        expected_item.each do |(key, value)|
          memo << { key => value } unless (yield actual_hash_includes?(key, value))
        end
      elsif comparing_hash_keys?(expected_item)
        memo << expected_item unless (yield actual_hash_has_key?(expected_item))
      else
        memo << expected_item unless (yield actual_collection_includes?(expected_item))
      end
      memo
    end
  end
end
```

Thoughts? (This also further improves the `matches?` and `does_not_match?` suggestions from above).

ChrisArcand: Much better; I knew I could do something with yielding a block but couldn't put it together :/ I'll give that a go.

lib/rspec/matchers/built_in/include.rb
myronmarston: How's this?

``` ruby
def failure_message
  format_failure_message("to") { super }
end

  format_failure_message("not to") { super }
end

def format_failure_message(preposition)
  if actual.respond_to?(:include?)
    improve_hash_formatting("expected #{description_of @actual} #{preposition} include#{readable_list_of @divergent_items}")
  else
    improve_hash_formatting(yield) + invalid_object_message
  end
end
```

ChrisArcand: Well, at least I added/edited all the tests for this PR of yours :wink: Looks good; we can eliminate `invalid_object_message` with that, too.

ChrisArcand:
@myronmarston This should be good to look at again, now.

- Fixed the last few issues related to hash output with Ruby 1.8.7
- Looks like the build is still failing because there maybe be a corrupted version of nokogiri in the gem cache for Ruby 2.0.0 in Travis.
- Benchmarks added; I've done little benchmarking so let me know if something doesn't look right.

As far as the benchmarks go I believe they make sense; The hit to performance in `to include (failures)`, as we mentioned earlier, is expected; where before the `#any?` determines the match and returns on the first item that failed, the other items in expected will still be checked against actual for determining the failure output.

myronmarston:
Thanks!

Issue #762, opened 2015-04-10
`raise_error` does not support a matcher as first argument being matched against the message


myronmarston:
`raise_error` supports composable matchers:

* The first arg is matched against the exception object, supporting things like [this](https://gist.github.com/myronmarston/457b08f4de08a59e3eba).
* The 2nd arg is matched against the exception message, supporting things like `a_string_including(...)`.

Sometimes you only care about matching on the string, though.  `raise_error` partially supports this by allowing you to pass a regex as the first argument.  I think we should similarly support a matcher matching against the message as the first argument.  Basically, we would change `raise_error` so that when it receives a single argument, it passes if that argument matches against either the exception object itself or the message.

myronmarston:
Addressed by #782.

PR #782, opened 2015-04-27
RaiseError can support first arg matcher on message.

imtayadeway:
addresses: https://github.com/rspec/rspec-expectations/issues/762

I'm not totally confident this is the right strategy to go for here. @myronmarston - could there be other matchers that match on `String`/`Regexp` whose intended target would be the exception, not the message?

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -16,7 +16,7 @@ module RSpec
           case expected_error_or_message
           when nil
             @expected_error, @expected_message = Exception, expected_message
-          when String, Regexp
+          when String
             @expected_error, @expected_message = Exception, expected_error_or_message
           else
             @expected_error, @expected_message = expected_error_or_message, expected_message
@@ -48,7 +48,8 @@ module RSpec
           begin
             given_proc.call
           rescue Exception => @actual_error
-            if values_match?(@expected_error, @actual_error)
+            if values_match?(@expected_error, @actual_error) ||
+               values_match?(@expected_error, @actual_error.message)
               @raised_expected_error = true
               @with_expected_message = verify_message
             end
@@ -173,7 +174,11 @@ module RSpec
         def expected_error
           case @expected_message
           when nil
-            description_of(@expected_error)
+            if RSpec::Support.is_a_matcher?(@expected_error)
+              "Exception with #{description_of(@expected_error)}"
+            else
+              description_of(@expected_error)
+            end
           when Regexp
             "#{@expected_error} with message matching #{description_of(@expected_message)}"
           else
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -454,12 +454,13 @@ RSpec.describe "expect { ... }.not_to raise_error(NamedError, error_message) { |
 end
 
 RSpec.describe "Composing matchers with `raise_error`" do
-  matcher :an_error_with_attribute do |attr|
+  matcher :an_attribute do |attr|
     chain :equal_to do |value|
       @expected_value = value
     end
 
     match do |error|
+      return false unless error.respond_to?(attr)
       error.__send__(attr) == @expected_value
     end
   end
@@ -470,18 +471,28 @@ RSpec.describe "Composing matchers with `raise_error`" do
 
   describe "expect { }.to raise_error(matcher)" do
     it 'passes when the matcher matches the raised error' do
-      expect { raise FooError }.to raise_error(an_error_with_attribute(:foo).equal_to(:bar))
+      expect { raise FooError }.to raise_error(an_attribute(:foo).equal_to(:bar))
+    end
+
+    it 'passes when the matcher matches the exception message' do
+      expect { raise FooError, "food" }.to raise_error(a_string_including("foo"))
     end
 
     it 'fails with a clear message when the matcher does not match the raised error' do
       expect {
-        expect { raise FooError }.to raise_error(an_error_with_attribute(:foo).equal_to(3))
-      }.to fail_including("expected an error with attribute :foo equal to 3, got #<FooError: FooError>")
+        expect { raise FooError }.to raise_error(an_attribute(:foo).equal_to(3))
+      }.to fail_including("expected Exception with an attribute :foo equal to 3, got #<FooError: FooError>")
+    end
+
+    it 'fails with a clear message when the matcher does not match the exception message' do
+      expect {
+        expect { raise FooError, "food" }.to raise_error(a_string_including("bar"))
+      }.to fail_including('expected Exception with a string including "bar", got #<FooError: food')
     end
 
     it 'provides a description' do
-      description = raise_error(an_error_with_attribute(:foo).equal_to(3)).description
-      expect(description).to eq("raise an error with attribute :foo equal to 3")
+      description = raise_error(an_attribute(:foo).equal_to(3)).description
+      expect(description).to eq("raise Exception with an attribute :foo equal to 3")
     end
   end
 


lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -204,6 +208,13 @@ def expecting_specific_exception?
         def raise_message_already_set
           raise "`expect { }.to raise_error(message).with_message(message)` is not valid. The matcher only allows the expected message to be specified once"
         end
+
+        def matching_on_message?(expected_error_or_message)
+          RSpec::Support.is_a_matcher?(expected_error_or_message) &&
+            Array(expected_error_or_message.expected).all? do |expected|
+            [String, Regexp].include?(expected.class)
+          end
+        end
```
myronmarston: There's a couple problems with this...

* `expected` is an optional part of the matcher protocol some matchers may not have it.
* A matcher like `start_with('Message start').and end_with('Message end')` isn't going to have a `String` or `Regexp` value for `expected`.
* More generally, this is a decent heuristic but misses many cases where the user may be intending to match on the string.

myronmarston:
Thanks for taking a stab at this!

I think that instead of trying to figure out if the provided argument is intended to match the message rather than the exception, we should just do something like:

``` ruby
values_match?(expected, exception) || values_match?(expected, exception.message)
```

IOW, try to match against both the exception and the exception's message, and if either are true, consider it to have matched.  That way we don't have to guess intent or use any kind of heuristic.  

It also gives us the possibility of cleaning up some of the special case code in place now.  For example, `Regexp` is called out as a special case right now but `Regexp` supports the same `===` protocol a matcher argument does so we should be able to treat them the same.  (`values_match?` takes care of this for you by using `===` internally).

Does that make sense?

imtayadeway:
@myronmarston thanks for the feedback! that makes sense, yes. i'll try to take care of this tomorrow.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -48,7 +48,8 @@ def matches?(given_proc, negative_expectation=false, &block)
           begin
             given_proc.call
           rescue Exception => @actual_error
-            if values_match?(@expected_error, @actual_error)
+            if values_match?(@expected_error, @actual_error) ||
+               (values_match?(@expected_error, @actual_error.message) rescue false)
```
imtayadeway: i have one test blowing up because of this line (on `spec/rspec/matchers/built_in/raise_error_spec.rb:480`) which i've temporarily dealt with by adding this `rescue` here ( :scream: ). Is this a case that the fuzzy matcher should try to handle? It fails with: `NoMethodError: undefined method `foo' for "FooError":String`

myronmarston: It's a bug in the custom matcher used in that spec. A matchers `matches?` should return `true` or `false` to indicate a `match`, and if it's nonsensical to match an object against it can return `false` from `matches?` and return a message explaining the nonsensicality in `failure_message`.

In this case, the `an_error_with_attribute` matcher should do a `error.respond_to?(attr)` check in it's `match` logic.

Does that make sense?

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -473,12 +473,22 @@ def foo; :bar; end
...
     it 'fails with a clear message when the matcher does not match the raised error' do
       expect {
         expect { raise FooError }.to raise_error(an_error_with_attribute(:foo).equal_to(3))
       }.to fail_including("expected an error with attribute :foo equal to 3, got #<FooError: FooError>")
     end
 
+    it 'fails with a clear message when the matcher does not match the exception message' do
+      expect {
+        expect { raise FooError, "food" }.to raise_error(a_string_including("bar"))
+      }.to fail_including('expected a string including "bar", got #<FooError: food')
```
myronmarston: We're not expecting a string including "bar" -- we're expecting an exception with a message that matches that.  It would be nice if the failure message said `expected Exception with message matching (a string including "bar")` (or something like that -- basically whatever `raise_error(Exception, a_string_including("bar"))` fails with now).  It looks like there's another special case for `Regexp` for that kind of phrasing in place that we should expand to work for matchers, too.

imtayadeway:
apologies for the delay and the going back and forth, @myronmarston. if you feel you need to take this over, i won't be offended :smile: this has been a lot more complicated than i first thought!

i've changed the way the custom matcher on https://github.com/imtayadeway/rspec-expectations/blob/tw/raise-error-first-arg-matchers/spec/rspec/matchers/built_in/raise_error_spec.rb#L457 works. if this is not right, i'll happily revert to what was before. to explain - i did this to be more like `a_string_including` - since both are matchers, it became hard to reason what `#expected_error` should do, when one's target was the exception and the other's was the message on the exception. i settled on 'Exception with <description>'.

thanks!

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -173,7 +174,11 @@ def warn_about_negative_false_positive(expression)
         def expected_error
           case @expected_message
           when nil
-            description_of(@expected_error)
+            if RSpec::Support.is_a_matcher?(@expected_error)
+              "Exception with #{description_of(@expected_error)}"
```
myronmarston: I guess we don't differentiate here between a matcher matching on the exception object vs message?  Seems like we'd want a different message for those cases.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -173,7 +174,11 @@ def warn_about_negative_false_positive(expression)
...
+            else
+              description_of(@expected_error)
+            end
           when Regexp
```
myronmarston: Can the regexp special handling be removed like we did for the match logic?

myronmarston:
No need to apologize, @imtayadeway -- we all do this in our free time and appreciate you're helping us out here.  I left a couple comments but it'd also be nice to have this squashed before merging.

imtayadeway:
@myronmarston just squashed my commits. disappointed to say that i played around with this a fair bit and couldn't come up with a good solution to your comments. for one, i couldn't find a way of determining if a matcher's target was the message of the exception, or anything else (e.g. the as with custom matcher here https://github.com/rspec/rspec-expectations/pull/782/files#diff-4247e8ed6201bb33735ad5faf3f96b86R457). i was also unable to remove the second special case for Regexps =(. let me know if i can do anything to mitigate these!

myronmarston:
Thanks, @imtayadeway -- this is fine how you have it.

bdunne:
@imtayadeway This seems to have introduced a bug. When upgrading from 3.2.1 to 3.3.0 I see this now passes:
```ruby
expect { raise "error" }.to raise_error(String)
```

To verify I wrote an rspec spec and ran it on both 3.2.1 (passes) and 3.3.0 (fails):
```ruby
it 'raise_error with a class should match the error class, not the message class' do
  expect { expect { raise "error" }.to raise_error String }.to raise_error RSpec::Expectations::ExpectationNotMetError
end
```

cc @carbonin

myronmarston:
@bdunne: Given that we know support matching a single `raise_error` arg against the message, I'd expect `expect { raise "error" }.to raise_error(String)` to pass -- `String === any_string` returns true. In affect, you are specifying that the exception's `message` must be a string.

In 3.2 and before, when you passed a single arg to `raise_error` it tried to match against the exception object itself.  `String === exception_object` returned `false` (which is why it failed on 3.2) but that's a bit of a nonsensical expectation.  I'd go far as to say that while that expectation behaved a certain way, we didn't document how it would behave in that situation and it was essentially undefined behavior.

Thus I don't consider the behavior in 3.3 to be a bug.

What is it you were trying to express with that expectation?   Maybe we can help you come up with an alternate way to write that.

carbonin:
@myronmarston: We actually came across this by accident and, after understanding what was going on, realized why it was passing, but mentioned it because of how strange a user experience it was.  We have a way to implement what we needed to do, but thought that failing that test would make more sense from a user's point of view because by passing a class the intent would be for `raise_error` to match against the only possible option for which the class varies (i.e. the exception's class rather than the message's class which will always be `String`).

So definitely not a bug, but maybe an enhancement.

myronmarston:
I get what you're saying, but treating `String` special in this way would be inconsistent to anyone who understands that we use `===` to support composable matchers (automatically gaining the ability to match against classes, regexes, ranges, etc), so I think we're going to keep it the way we have it now.

JonRowe:
:+1:

PR #783, opened 2015-04-29
`satisfy` matcher custom descriptions

### LINKED ISSUE #719: Allow a custom message for satisfy expectations
I noticed that "satisfy block" description is hardcoded into the `Satisfy` matcher class. It should be possible to override the default message with a more meaningful one.

    satisfy("are even numbers") { |e| e % 2 == 0 }

I currently modified `Satisfy` and `#satisfy` to allow this.

### satisfy.rb

      class Satisfy < RSpec::Matchers::BuiltIn::Satisfy

        # @api private
        # @return [String]
        attr_reader :description

        def initialize(description='satify block',&block)
          super(&block)

          @description = description
        end

        # @api private
        # @return [String]
        def failure_message
          "expected #{@actual} to #{@description}"
        end

        # @api private
        # @return [String]
          "expected #{@actual} not to #{@description}"
        end

      end

ChrisArcand:
Closes #719

...So, awkwardly I _just_ stumbled upon https://github.com/rspec/rspec-expectations/pull/259 while searching for the issue this PR is supposed to resolve. 

@myronmarston This PR resolves the open issue as requested, but the closed issue contradicts it. Were you deciding that now this is a good idea?

TBH I had no idea you could pass a description to any matcher like that, and if that's the case I'd tend to agree with your past self :) If you still agree with that, I'll just leave it as is and instead change the documentation for `satisfy` to suggest giving a better description using the existing way.

### DIFF
--- a/features/built_in_matchers/satisfy.feature
+++ b/features/built_in_matchers/satisfy.feature
@@ -8,9 +8,14 @@ Feature: `satisfy` matcher
     expect(7).not_to satisfy { |v| v % 5 == 0 }
     ```
 
-  This flexibility comes at a cost, however: the failure message ("expected [actual] to satisfy
-  block") is not very descriptive or helpful.  You will usually be better served by using one of
-  the other built-in matchers, or writing a custom matcher.
+  The default failure message ("expected [actual] to satisfy block") is not very descriptive or helpful.
+  To add clarification, you can provide your own description as an argument:
+
+    ```ruby
+    expect(10).to satisfy("be a multiple of 5") do |v|
+      v % 5 == 0
+    end
+    ```
 
   Scenario: basic usage
     Given a file named "satisfy_matcher_spec.rb" with:
@@ -22,11 +27,15 @@ Feature: `satisfy` matcher
         # deliberate failures
         it { is_expected.not_to satisfy { |v| v > 5 } }
         it { is_expected.to satisfy { |v| v > 15 } }
+        it { is_expected.to_not satisfy("be greater than 5") { |v| v > 5 } }
+        it { is_expected.to satisfy("be greater than 15") { |v| v > 15 } }
       end
       """
     When I run `rspec satisfy_matcher_spec.rb`
     Then the output should contain all of these:
-      | 4 examples, 2 failures           |
-      | expected 10 not to satisfy block |
-      | expected 10 to satisfy block     |
+      | 6 examples, 4 failures               |
+      | expected 10 not to satisfy block     |
+      | expected 10 to satisfy block         |
+      | expected 10 not to be greater than 5 |
+      | expected 10 to be greater than 15    |
 
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -767,10 +767,13 @@ module RSpec
     # If you do find yourself in such a situation, you could always write
     # a custom matcher, which would likely make your specs more expressive.
     #
+    # @param description [String] optional description to be used for this matcher.
+    #
     # @example
     #   expect(5).to satisfy { |n| n > 3 }
-    def satisfy(&block)
-      BuiltIn::Satisfy.new(&block)
+    #   expect(5).to satisfy("be greater than 3") { |n| n > 3 }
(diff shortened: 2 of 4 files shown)


JonRowe:
I'm on the fence here, on the one hand having two ways to do the same thing seems redundant, however people evidentially don't know about the existing way to set this up and having a hardcoded description for satisfy does seem silly... WDYT @myronmarston

myronmarston:
I had forgotten this had come up before in #259!

One key difference from then vs now is that matchers are now fully composable...which makes this feature much, much more useful than it would have been at the time.  For example, this allows you to do this:

``` ruby
expect([1, 3]).to include an_object_satisfying("an even object", &:even?)
```

...which will cause "an even object" to be included in the failure message. The suggestion from #259 (of passing a custom failure message) only works forth expectation as a whole and doesn't work for a nested matcher like this.

So yes, I'm in favor of this and would like to see it merged.  I'll go review it in detail now.

features/built_in_matchers/satisfy.feature
```diff
@@ -8,9 +8,14 @@ Feature: `satisfy` matcher
...
     ```
 
-  This flexibility comes at a cost, however: the failure message ("expected [actual] to satisfy
-  block") is not very descriptive or helpful.  You will usually be better served by using one of
-  the other built-in matchers, or writing a custom matcher.
+  The default failure message ("expected [actual] to satisfy block") is not very descriptive or helpful.
+  To add clarification, you can provide your own description as an argument:
+
+    ```ruby
+    expect(11).to satisfy("have a remainder of 0 with modulo 5") do |v|
```
myronmarston: This seems like a verbose way to say "be evenly divisible by 5" or "be a multiple of 5".  Thoughts on using one of those shorter descriptions?

lib/rspec/matchers.rb
```diff
@@ -769,8 +769,9 @@ def respond_to(*names)
     #
     # @example
     #   expect(5).to satisfy { |n| n > 3 }
-    def satisfy(&block)
-      BuiltIn::Satisfy.new(&block)
+    #   expect(5).to satisfy("be greater than 3") { |n| n > 3 }
```
myronmarston: Do you mind adding a `# @param` declaration for the argument as well?  To see how it will render in our docs, use `bundle exec yard server --reload`.

spec/rspec/matchers/built_in/satisfy_spec.rb
```diff
@@ -24,6 +24,30 @@
...
+    end
+
+    it "passes if block returns true" do
+      expect(true).to satisfy("be true") { |val| val }
+      expect(true).to satisfy("be true") do |val|
+        val
+      end
+    end
+
+    it "fails if block returns false" do
```
myronmarston: WDYT of `it "fails with the custom description block returns false"`?  That's an important aspect of the behavior.

spec/rspec/matchers/built_in/satisfy_spec.rb
```diff
@@ -39,4 +63,19 @@
...
+
+  context "when a custom description is provided" do
+    it "passes if block returns false" do
+      expect(false).not_to satisfy("be true") { |val| val }
+      expect(false).not_to satisfy("be true") do |val|
+        val
+      end
+    end
+
+    it "fails if block returns true" do
```
myronmarston: Same here.  I'd suggest: `it "fails with the custom description if block returns true"`.

myronmarston:
Thanks, @ChrisArcand.  Your implementation is very clean and simple, and I appreciate the fact you thought to update the cuke as well :+1:.

ChrisArcand:
Ahh I had forgotten about composable matchers, that makes sense. Sounds good.

features/built_in_matchers/satisfy.feature
ChrisArcand: Definitely, much more sensible. I'll go with 'be a multiple of 5'

ChrisArcand:
Done. Build failure due to the typical 2.0.0 nokogiri woes.

myronmarston:
Thanks @ChrisArcand!

Issue #767, opened 2015-04-11
Get test coverage of all code that's not version-specific up to 100%


myronmarston:
We did this previously in rspec-core in rspec/rspec-core#1905.  All code that is not currently covered should be looked at, and it should be determined which of these categories it's in:
- Dead code that can be deleted.
- Code that's only for older versions of Ruby (or JRuby or whatever). Tag it with `# :nocov:` so it's not included the coverage percent calculation.
- Code that can and should be covered by a new test.

nevinera:
I know this is _ancient_, but it's actually pretty close already - after tagging the version-specific code with `:nocov:`, there are only a total of 18 lines uncovered; I'll tackle the tests for those lines tomorrow. Do we want to _enforce_ 100% coverage in some way after it's achieved, perhaps with a (non-required) github status?

pirj:
> only a total of 18 lines uncovered; I'll tackle the tests for those lines

Nice!

> non-required) github status, or just by cranking the [minimum coverage](https://github.com/rspec/rspec-expectations/blob/main/spec/spec_helper.rb#L5) up to 100%

That would be perfect. If we’d make it a separate GH check, we would have to run it on a specific Ruby version. We can pick the version that gets us closest to 100%

nevinera:
> I'd favour adding a check to our matrix running on the latest Ruby only, please note until I finish the monorepo work we need to apply such changes in `rspec-dev` and it will mean checking some flag to set which repos support 100% coverage enforcement

Meaning 3.3 for now, and not ruby-head?

Ah, if you're working on mono-repo-izing, I'll just do the actual coverage bit in this PR; we can talk enforcement in another ticket. Might be simplest to just defer that part until after you're done. (Guess I should take a look at the coverage of the other gem repos next!)

nevinera:
I think #1458 resolved this finally :-)

PR #784, opened 2015-04-29
Omit all Ruby < 1.9 only code from test coverage

ChrisArcand:
Part of #767, omits code specific to older versions of Ruby from test coverage.


JonRowe:
LGTM, thanks! Merging despite build failure because it's just 1.9.2 barfing on that gem install.

myronmarston:
Any idea how close we are to 100% with these changes?

ChrisArcand:
99.22% IIRC. A few lines scattered across maybe 5-6 files to add tests to.

myronmarston:
Nice!  Almost there then.

PR #789, opened 2015-05-18
Exception failures in `match_when_negated`; Fix `NegativeExpectationHandler::handle_matcher`

### LINKED ISSUE #788: `match_when_negated` does not handle expectation failures like `match` does
The definition of `match` in the matcher DSL rescues `ExpectationNotMetError` and returns `false`, allowing expectations to be used in `match` blocks:

https://github.com/rspec/rspec-expectations/blob/6717b726863f668a9eeebd069706e0187edee99b/lib/rspec/matchers/dsl.rb#L64-L65

However, `match_when_negated` does not do this:

https://github.com/rspec/rspec-expectations/blob/6717b726863f668a9eeebd069706e0187edee99b/lib/rspec/matchers/dsl.rb#L76-L81

This should be fixed, as these two methods should have parity with respect to how expectation failures are treated.

Note that I'm currently working on some edits in this area as part of #776 so whoever works on this may want to hold off until that PR is merged to reduce the likliehood of merge conflicts.

chrisarcand:
Fixes #788 

This fixes two issues; The second was found after testing the first led to it, and is dependent on it.

* Fixes `match_when_negated` to work as `match` does. The definition of #match in the matcher DSL rescues ExpectationNotMetError and returns false, allowing expectations to be used in match blocks. This changes #match_when_negated to act similarly.
* Using a negative matcher _with differing values_ (ie: `expect("foo").not_to eq "bar"`) has returned `false` for a long time, instead of the handler properly raising `ExpectationNotMetError`. This changes the handler to similarly do it as the positive handler does. See commit message in 7880b129963f8196dd288bdf834167f1204fe760 for more details.

### DIFF
--- a/lib/rspec/expectations/handler.rb
+++ b/lib/rspec/expectations/handler.rb
@@ -69,7 +69,7 @@ module RSpec
       def self.handle_matcher(actual, initial_matcher, message=nil, &block)
         ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::NegativeOperatorMatcher.new(actual) unless initial_matcher
-          !(does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated))
+          does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated)
         end
       end
 
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -80,8 +80,14 @@ module RSpec
         # @yield [Object] actual the actual value (i.e. the value wrapped by `expect`)
         def match_when_negated(&match_block)
           define_user_override(:does_not_match?, match_block) do |actual|
-            @actual = actual
-            super(*actual_arg_for(match_block))
+            begin
+              @actual = actual
+              RSpec::Support.with_failure_notifier(RAISE_NOTIFIER) do
+                super(*actual_arg_for(match_block))
+              end
+            rescue RSpec::Expectations::ExpectationNotMetError
+              false
+            end
           end
         end
 
--- a/spec/rspec/matchers/dsl_spec.rb
+++ b/spec/rspec/matchers/dsl_spec.rb
@@ -563,6 +563,68 @@ module RSpec::Matchers::DSL
         expect(self.matcher.matches?('other value')).to be_falsey
       end
 
+      context "with a match_when_negated block" do
+        let(:matcher) do
+          new_matcher(:name, "value") do |expected|
+            match do |actual|
+              expect(actual).to eq expected
+            end
+
+            match_when_negated do |actual|
+              expect(actual).not_to eq expected
+            end
+          end
+        end
+
+        it "returns true if the wrapped expectation passes" do
+          expect(self.matcher.does_not_match?('other value')).to be_truthy
+        end
+
+        it "returns false if the wrapped expectation fails" do
+          expect(self.matcher.does_not_match?('value')).to be_falsey
+        end
+      end
+
+      context "with a negative expectation" do
+        let(:matcher) do
+          new_matcher(:name, "purposely_the_same") do |expected|
+            match do |actual|
+              expect(actual).not_to eq expected
+            end
+          end
+        end
+
+        it "returns true if the wrapped expectation passes" do
+          expect(self.matcher.matches?('purposely_different')).to be_truthy
+        end
+
+        it "returns false if the wrapped expectation fails" do
+          expect(self.matcher.matches?('purposely_the_same')).to be_falsey
+        end
+
+        context "with a match_when_negated block" do
+          let(:matcher) do
+            new_matcher(:name, "purposely_the_same") do |expected|
+              match do |actual|
+                expect(actual).not_to eq expected
+              end
+
+              match_when_negated do |actual|
+                expect(actual).to eq expected
+              end
+            end
+          end
+
+          it "returns true if the wrapped expectation passes" do
+            expect(self.matcher.does_not_match?('purposely_the_same')).to be_truthy
+          end
+
+          it "returns false if the wrapped expectation fails" do
+            expect(self.matcher.does_not_match?('purposely_different')).to be_falsey
+          end
+        end
+      end
+
       it "can use the `include` matcher from a `match` block" do
         RSpec::Matchers.define(:descend_from) do |mod|
           match do |klass|
@@ -618,6 +680,26 @@ module RSpec::Matchers::DSL
             expect(error).to have_attributes(:failures => [
               an_object_having_attributes(:message => "expected `1.even?` to return true, got false"),
               an_object_having_attributes(:message => 'expected "foo" to use an internal expectation')
+            ])
+          end
+        end
+
+        it 'does not aggregate the inner expectation failure (negation)' do
+          use_an_internal_expectation = new_matcher(:use_an_internal_expectation) do
+            match_when_negated do |actual|
+              expect(actual).not_to end_with "o"
+            end
+          end
+
+          expect {
+            aggregate_failures do
+              expect(1).to be_even
+              expect("foo").not_to use_an_internal_expectation
+            end
+          }.to fail do |error|
+            expect(error).to have_attributes(:failures => [
+              an_object_having_attributes(:message => "expected `1.even?` to return true, got false"),
+              an_object_having_attributes(:message => 'expected "foo" not to use an internal expectation')
             ])
           end
         end
--- a/spec/rspec/expectations/handler_spec.rb
+++ b/spec/rspec/expectations/handler_spec.rb
@@ -137,27 +137,52 @@ module RSpec
 
     RSpec.describe NegativeExpectationHandler do
       describe "#handle_matcher" do
-        it "asks the matcher if it doesn't match when the matcher responds to #does_not_match?" do
-          matcher = double("matcher", :does_not_match? => true, :failure_message_when_negated => nil)
-          actual = Object.new
-          expect(matcher).to receive(:does_not_match?).with(actual).and_return(true)
-          RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher)
+        context "with matching values" do
+          context "when the matcher responds to #does_not_match?" do
+            it "returns false" do
+              matcher = double("matcher", :failure_message_when_negated => "Error!")
+              actual = Object.new
+
+              expect(matcher).to receive(:does_not_match?).with(actual).and_return(false)
+              expect {
+                RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher)
+              }.to raise_error(RSpec::Expectations::ExpectationNotMetError, "Error!")
+            end
+          end
+
+          context "when the matcher doesn't respond to #does_not_match?" do
+            it "returns false" do
+              matcher = double("matcher", :failure_message_when_negated => "Error!")
+              actual = Object.new
+
+              expect(matcher).to receive(:matches?).with(actual).and_return(true)
+              expect {
+                RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher)
+              }.to raise_error(RSpec::Expectations::ExpectationNotMetError, "Error!")
+            end
+          end
         end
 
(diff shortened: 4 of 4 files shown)


lib/rspec/expectations/handler.rb
```diff
@@ -69,7 +69,7 @@ class NegativeExpectationHandler
       def self.handle_matcher(actual, initial_matcher, message=nil, &block)
         ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::NegativeOperatorMatcher.new(actual) unless initial_matcher
-          !(does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated))
+          does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated)
```
ChrisArcand: Can someone check my sanity here? If the spec descriptions hold true, what was here doesn't return the matcher value (and it should). If not, then the spec descriptions are wrong and `match`, `matches?`, `does_not_match?`, etc should be returning booleans...I think.

lib/rspec/matchers/dsl.rb
```diff
@@ -80,11 +77,20 @@ def match(&match_block)
         # @yield [Object] actual the actual value (i.e. the value wrapped by `expect`)
         def match_when_negated(&match_block)
           define_user_override(:does_not_match?, match_block) do |actual|
-            @actual = actual
-            super(*actual_arg_for(match_block))
+            begin
+              @actual = actual
+              RSpec::Support.with_failure_notifier(RAISE_METHOD) do
```
ChrisArcand: I'm unsure as to the purpose of this really, but emulated @myronmarston 's recent changes in the `match` method.

lib/rspec/matchers/dsl.rb
```diff
@@ -80,11 +77,20 @@ def match(&match_block)
...
+                super(*actual_arg_for(match_block))
+              end
+            rescue RSpec::Expectations::ExpectationNotMetError
+              false
```
ChrisArcand: This is correct, no? `does_not_match?` should return false if the expectation included in a `match_when_negated` block fails(?)

lib/rspec/expectations/handler.rb
JonRowe: Well they returned booleans before, and thats my understanding of the protocol, this has just inverted the logic which I'd hazard is undesirable...

lib/rspec/matchers/dsl.rb
```diff
@@ -80,11 +77,20 @@ def match(&match_block)
...
+            end
           end
         end
 
+        # @private
+        RAISE_METHOD = method(:raise)
+
```
JonRowe: There was no need to move this

lib/rspec/matchers/dsl.rb
JonRowe: No it should return true as this is the inverse case.

lib/rspec/matchers/dsl.rb
JonRowe: It's to support the recent changes surrounding aggregate failures.

lib/rspec/expectations/handler.rb
myronmarston: > If the spec descriptions hold true

To what spec descriptions are you referring?

Anyhow, this is definitely a breaking change, as removing `!` inverts things from it worked before.  It's not clear to me what you're confused about...

lib/rspec/matchers/dsl.rb
myronmarston: See https://relishapp.com/rspec-staging/rspec-expectations/docs/aggregating-failures for the feature @JonRowe mentioned.  See cc66c173b148ef666271359c8885ac4555f7d473 for the change that went into that commit.  The specs there show a case where this logic is necessary.  It's good you've added this here but we're going to want a spec for it as well.  Can you add one based on https://github.com/rspec/rspec-expectations/blob/cc66c173b148ef666271359c8885ac4555f7d473/spec/rspec/matchers/dsl_spec.rb#L605-L623 ?

lib/rspec/matchers/dsl.rb
myronmarston: > No it should return true as this is the inverse case.

That doesn't sound right to me.  `does_not_match?` is invoked when `expect(x).not_to matcher` is used.  If it returns `true` then that means the expectation will pass.  If you use an expectation within `match_when_negated`, then `does_not_match?` should return `true` if the expectation passes and `false` if the expectation fails -- so within the `rescue` it should return false, I think.

myronmarston:
> I would expect that they return a boolean value of whether or not the matcher matches. According to other spec written for it, it looks like it's intended to return the value of the matcher instead. However, there is a logical ! in the negative version of the method which would force it to a boolean...? More details below.

An expectation expression returns a boolean value.  The return value of `matches?` and the `match` block is treated as a boolean (regardless of if it actually is).

lib/rspec/expectations/handler.rb
ChrisArcand: I'll find a way to make this work without this breaking change, but to the return value confusion I'm talking about.... https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/expectations/handler_spec.rb#L75-L80

lib/rspec/matchers/dsl.rb
ChrisArcand: Sure can.

lib/rspec/matchers/dsl.rb
ChrisArcand: I moved this because it seems more appropriate to place after both `#match` and `#match_when_negated` (it was put between them originally because it was only used in `#match`). Although I think it makes more sense here now, if your intention is just to keep the lines in git history belonging to the original commit where functionality was changed, I understand. I'll move that back.

JonRowe: Well it's just it confuses the diff, its not related to this change so why move it!, it's also now below where it's used which is also confusing

ChrisArcand: Well, it was defined below where it was used before, too - I was just emulating what was already there :) I will certainly move it back to simplify the diff though, sounds good.

lib/rspec/expectations/handler.rb
ChrisArcand: @myronmarston @JonRowe After digging in to this further with a clear head, I see where my confusion lies. `#does_not_match?` here correctly returns `true` (it's true it does not match) and I don't really know why the logic is then being inverted.

Running off of fresh master, this behavior confuses me:

```plaintext
> expect("foo").to eq "foo"
=> true   # Good!

> expect("foo").to eq "bar"
RSpec::Expectations::ExpectationNotMetError:
expected: "bar"
     got: "foo"

(compared using ==)
# Good!

> expect("foo").not_to eq "foo"
RSpec::Expectations::ExpectationNotMetError:
expected: value != "foo"
     got: "foo"

(compared using ==)
# Good!

> expect("foo").not_to eq "bar"
=> false
# Whaaaat? This doesn't seem right.
```

Shouldn't the last one be `true`, because "foo" is not equal to "bar"? :confused: Seems to screw up what we're trying to accomplish in this PR (parity between `match` and `match_when_negated`)

lib/rspec/matchers/dsl.rb
ChrisArcand: Added [here](https://github.com/rspec/rspec-expectations/pull/789/files#diff-6502c624fc79bdd0d26e85e8a1fe8180R647); How's that?

ChrisArcand:
Forced some changes; As you can see by the diff now, I've only added tests and changed `match_if_negated` to act similarly to `match`. The failing test in the build is one added by me that seems to demonstrate my confusion outlined [here](https://github.com/rspec/rspec-expectations/pull/789#discussion_r30668067).

myronmarston:
Hey @ChrisArcand -- sorry for the delay.  I'm going to try to follow up on this in the next couple days so we can get this merged soon.

ChrisArcand:
@myronmarston Absolutely no rush :ok:

mrageh:
@myronmarston, @JonRowe and @ChrisArcand This PR looks like it's nearly complete. What else do we need to do to get this merged in?

ChrisArcand:
Headed out of town for a week, but when I return I'll finish this with any feedback that gets added while I'm gone.

JonRowe:
Probably needs a rebase but also needs a passing build..

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -563,6 +563,28 @@ def foo
...
+
+            match_when_negated do |actual|
+              expect(actual).not_to eq expected
+            end
+          end
+        end
+
+        it "returns true if the wrapped expectation passes" do
+          expect(self.matcher.does_not_match?('other value')).to be_truthy
+        end
```
myronmarston: I think I figured out why this spec is failing -- I think the negation on [this line](https://github.com/rspec/rspec-expectations/blob/1c63b42d1ee1a07c14f9543e8274cc892f829031/lib/rspec/expectations/handler.rb#L72) is wrong.  Note, however, that removing that negation makes another spec fail, but I think that other spec is incorrect.

All that said: removing the negation on that line worries me a little as it's hard to keep track of all the negations going on in my head and I'm not 100% sure that's correct.  What would really help solidify it is to make sure we have specs covering all 8 of these cases:

* A matcher that uses a positive expectation in `match` checked against a matching value
* A matcher that uses a positive expectation in `match` checked against an unmatching value
* A matcher that uses a negative expectation in `match` checked against a matching value
* A matcher that uses a negative expectation in `match` checked against an unmatching value
* A matcher that uses a positive expectation in `match_when_negated` checked against a matching value
* A matcher that uses a positive expectation in `mach_when_negated` checked against an unmatching value
* A matcher that uses a negative expectation in `match_when_negated` checked against a matching value
* A matcher that uses a negative expectation in `match_when_negated` checked against an unmatching value

Can you check to make sure all those cases are covered, and add specs as needed?

chrisarcand: Finally getting around to finishing this PR. Thanks for your patience!

@myronmarston, your speculation here is exactly what I said in [this comment](https://github.com/rspec/rspec-expectations/pull/789#discussion_r30530471). ;) I'm glad you came to the same conclusion I did separately as it's a very head scratching thing. I'll see what I can do about adding specs to make sure we're good but this _does_ seem like if its wrong things should be _quite_ broken in master(?). The behavior I outlined at the end of the comment I mention above is still true in latest master (`expect("foo").not_to eq "bar" => false`)

chrisarcand: Putting this here if only to track my own thoughts to not get lost in this confusing bug :)

chrisarcand:
@myronmarston @JonRowe Sorted this out, added plenty of spec, rebased, and updated the description to this. It's ready for a good glance. I'm kind of blown away about the boolean in `NegativeExpectationHandler::handle_matcher`, so going over my specs in great detail might be wise. I think this is the correct behavior now, though.

chrisarcand:
@myronmarston Very gentle reminder, just in case this got lost in the inbox, that this is still open and ready when you find some time ^_^ Also before I forget about it again, myself!

myronmarston:
Thanks for your gentle reminder, Chris, and your patience.  Looking at it now.

myronmarston:
This is looking great, @chrisarcand.  Thanks for digging into this long-standing issue.  One point of confusion...in the commit message from your 2nd commit, you say:

```
    > expect("foo").not_to eq "bar"
    => false
    WRONG. Should raise ExpectationNotMetError
```

Shouldn't that be this instead?

```
    > expect("foo").not_to eq "bar"
    => false
    WRONG. Should return true.
```

If I'm right, do you mind amending your commit message so that we aren't left with a commit message that's really useful and helpful except for the one part which is confusing? Then I'll happily merge.  Thanks!

chrisarcand:
Hah, right you are. I appreciate the attention to commit messages ;)

myronmarston:
> Hah, right you are. I appreciate the attention to commit messages ;)

You wrote such a great, useful commit message, that I had to read it and check it :).

chrisarcand:
@myronmarston :white_check_mark:

myronmarston:
Thanks @chrisarcand!

:heart: :heart: :heart:

chrisarcand:
:+1: :+1:

PR #793, opened 2015-05-27
Prep for extra options on failure_notifier

JonRowe:
We wrap the `RAISE_METHOD` in a similarly captured proc so that adding options to the failure_notifier won't break it, similarly we prep aggregate_failures for receiving options.


JonRowe:
This isn't needed if we take @myronmarston's approach with rspec/rspec-support#210

PR #796, opened 2015-05-29
Deal with JRuby caller vs raise stacktrace differences.

myronmarston:
This should fix the JRuby failures happening on rspec/rspec-core#1977.


JonRowe:
LGTM

PR #798, opened 2015-06-02
Address 1.8.7 hash ordering

myronmarston:
This addresses the failures from rspec/rspec-core#1971.  Not sure why they started failing and I can't repro locally but this should fix it.


JonRowe:
LGTM

PR #799, opened 2015-06-03
Dedup exceptions

myronmarston:
This is the necessary changes to support rspec/rspec-mocks#956.

### DIFF
--- a/lib/rspec/expectations/failure_aggregator.rb
+++ b/lib/rspec/expectations/failure_aggregator.rb
@@ -41,7 +41,11 @@ module RSpec
 
       # This method is defined to satisfy the callable interface
       # expected by `RSpec::Support.with_failure_notifier`.
-      def call(failure, _options)
+      def call(failure, options)
+        source_id = options[:source_id]
+        return if source_id && @seen_source_ids.key?(source_id)
+
+        @seen_source_ids[source_id] = true
         assign_backtrace(failure) unless failure.backtrace
         failures << failure
       end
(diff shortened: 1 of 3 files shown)


myronmarston:
This is ready for review, @rspec/rspec.

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -41,7 +41,11 @@ def other_errors
 
       # This method is defined to satisfy the callable interface
       # expected by `RSpec::Support.with_failure_notifier`.
-      def call(failure)
+      def call(failure, options)
+        source_id = options[:source_id]
+        return if source_id && @seen_source_ids.key?(source_id)
+
+        @seen_source_ids[source_id] = true
```
JonRowe: Why not write this as a hash with an || assign, and map the failures out at the end, I think that could be neater.

```Ruby
failures[options[:source_id]] ||=
  begin
    assign_backtrace(failure) unless failure.backtrace
    failure
  end
```

JonRowe:
Suggested a refactoring but it's not a merge blocker

lib/rspec/expectations/failure_aggregator.rb
myronmarston: * It's not clear to me what the advantage of that is over what we've got here.  It doesn't seem any neater or better to me.
* It's important that we keep the ordering of failures so that we can list them as they occurred.  While hashes are ordered on the most common rubies (e.g. MRI 1.9 to 2.2), ordering is not part of the core semantic of hashes and is not a behavior that is present on hashes in all rubies (e.g. MRI 1.8.7. Not sure about JRuby/Rbx).

Thus I think it's better to keep the failures array separate from tracking what source ids we've seen.

PR #807, opened 2015-06-22
This adds the `any`-matcher

dg-ratiodata:
You guys already added the `all`-matcher. This is very helpful. Unfortunately there's no `any`-matcher available. I hope I do not oversaw a similar functionality.

To have the `any`-matcher along this would make code at cucumber/aurba easier. This PR adds this matcher. It is based on the `All`-matcher and is a sub-class of this and re-uses most code given there.

BTW:
This PR also changes the behaviour of `all` when an exception - aka `ArgumentError` - is raised. Now the `matches`-call is wrapped with in a `begin; end`-block to make it check all objects and not to stop on the first one where the exception occured.

### DIFF
--- a/features/built_in_matchers/all.feature
+++ b/features/built_in_matchers/all.feature
@@ -30,13 +30,18 @@ Feature: `all` matcher
         it { is_expected.to all( be_a(String) ) }
         it { is_expected.to all( be > 2 ) }
       end
+
+      RSpec.describe [1, 'a', 5] do
+        it { is_expected.to all( be < 10 ) }
+      end
       """
     When I run `rspec array_all_matcher_spec.rb`
     Then the output should contain all of these:
-      | 6 examples, 3 failures                        |
+      | 7 examples, 4 failures                        |
       | expected [1, 3, 5] to all be even             |
       | expected [1, 3, 5] to all be a kind of String |
       | expected [1, 3, 5] to all be > 2              |
+      | expected [1, "a", 5] to all be < 10           |
 
   Scenario: compound matcher usage
     Given a file named "compound_all_matcher_spec.rb" with:
--- a/features/built_in_matchers/any.feature
+++ b/features/built_in_matchers/any.feature
@@ -1,0 +1,61 @@
+Feature: `any` matcher
+
+  Use the `any` matcher to specify that a collection's objects any pass an expected matcher.
+  This works on any enumerable object.
+
+    ```ruby
+    expect([1, 4, 5]).to any( be_odd )
+    expect([1, 3, 'a']).to any( be_an(Integer) )
+    expect([1, 3, 11]).to any( be < 10 )
+    ```
+
+  The matcher also supports compound matchers:
+
+    ```ruby
+    expect([1, 'a', 11]).to any( be_odd.and be < 10 )
+    ```
+
+  Scenario: array usage
+    Given a file named "array_any_matcher_spec.rb" with:
+      """ruby
+      RSpec.describe [1, 4, 'a', 11] do
+        it { is_expected.to any( be_odd ) }
+        it { is_expected.to any( be_an(Integer) ) }
+        it { is_expected.to any( be < 10 ) }
+      end
+
+      RSpec.describe [14, 'a'] do
+        it { is_expected.to any( be_odd ) }
+        it { is_expected.to any( be_an(Symbol) ) }
+        it { is_expected.to any( be < 10 ) }
+      end
+      """
+    When I run `rspec array_any_matcher_spec.rb`
+    Then the output should contain all of these:
+       | 6 examples, 3 failures                        |
+       | expected [14, "a"] to any be odd              |
+       | expected [14, "a"] to any be a kind of Symbol |
+       | expected [14, "a"] to any be < 10             |
+
+  Scenario: compound matcher usage
+    Given a file named "compound_any_matcher_spec.rb" with:
+      """ruby
+      RSpec.describe [1, 'anything', 'something'] do
+        it { is_expected.to any( be_a(String).and include("thing") ) }
+        it { is_expected.to any( be_a(String).and end_with("g") ) }
+        it { is_expected.to any( start_with("s").or include("y") ) }
+      end
+
+      RSpec.describe ['anything', 'something'] do
+        it { is_expected.to any( be_a(Integer).and include("thing") ) }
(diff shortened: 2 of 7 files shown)


dg-ratiodata:
Normally I don't change a project's `Gemfile`, but without that change I was not able to run the test suite.

JonRowe:
> Is there a timeline when you drop support for End of Service-rubies?

Not yet. As a testing framework we should be the last to drop them, but we are vaguely debating dropping 1.8.7 in 4.0

dg-ratiodata:
> Not yet. As a testing framework we should be the last to drop them, but we are vaguely debating dropping 1.8.7 in 4.0

Ok. Good to know. 

What do you think about the PR?

features/built_in_matchers/all.feature
```diff
@@ -30,13 +30,18 @@ Feature: `all` matcher
         it { is_expected.to all( be_a(String) ) }
         it { is_expected.to all( be > 2 ) }
       end
+
+      RSpec.describe [1, 'a', 5] do
+        it { is_expected.to all( be < 10 ) }
+      end
```
JonRowe: This PR is about any but this is for all...

features/built_in_matchers/all.feature
```diff
@@ -30,13 +30,18 @@ Feature: `all` matcher
...
       """
     When I run `rspec array_all_matcher_spec.rb`
     Then the output should contain all of these:
-      | 6 examples, 3 failures                        |
+      | 7 examples, 4 failures                        |
       | expected [1, 3, 5] to all be even             |
       | expected [1, 3, 5] to all be a kind of String |
       | expected [1, 3, 5] to all be > 2              |
+      | expected [1, "a", 5] to all be < 10           |
```
JonRowe: Ditto

features/built_in_matchers/any.feature
```diff
@@ -0,0 +1,61 @@
+Feature: `any` matcher
+
+  Use the `any` matcher to specify that a collection's objects any pass an expected matcher.
```
JonRowe: `a collection's objects any pass an expected` makes no sense, consider rewording to something like `any of a collections object's pass an expected`

features/built_in_matchers/any.feature
```diff
@@ -0,0 +1,61 @@
...
+  This works on any enumerable object.
+
+    ```ruby
+    expect([1, 4, 5]).to any( be_odd )
+    expect([1, 3, 'a']).to any( be_an(Integer) )
+    expect([1, 3, 11]).to any( be < 10 )
```
JonRowe: I think we'd want to add and use aliases here, e.g. `have_a_member` or similar more readable examples

lib/rspec/matchers/built_in/all.rb
```diff
@@ -50,7 +50,13 @@ def match(_expected, _actual)
         def index_failed_objects
           actual.each_with_index do |actual_item, index|
             cloned_matcher = matcher.clone
-            matches = cloned_matcher.matches?(actual_item)
+
+            begin
+              matches = cloned_matcher.matches?(actual_item)
+            rescue StandardError
+              matches = nil
```
JonRowe: It's non obvious why this is here, and generally it's not something we want to do, so reasoning is required before sneaking this under the radar

features/built_in_matchers/all.feature
dg-ratiodata: Sure, this is why is in a separate commit. I found this one, when developing `any`.

lib/rspec/matchers/built_in/all.rb
dg-ratiodata: Ok. Given an exception is raised by using a matcher, say

```ruby
it { expect([1, 'a', 11]).to any(be < 10) }
```

This will stop evaluation of matchers immediately. `11` is not checked anymore.

features/built_in_matchers/any.feature
dg-ratiodata: Ok. Added locally.

myronmarston:
We have intentionally not provided an `any` matcher for two reasons:

* `include` already provides the same capability.  `expect(list).to include(a_value > 10)` passes if `list.any? { |x| x > 10 }`.
* `expect(list).to all be_odd` reads like English; but `expect(list).to any be_odd` is completely nonsensical to a native English speaker.  (I'm actually not sure of the grammatical reasons for why that is.)

As for the `rescue` added to `all`, we take the position that it is the responsibility of each matcher to to return `true`/`false` from `matches?` so for a case like `expect([1, 'a', 11]).to all(be < 10)`, the bug is in hte `be < 10` matcher as it should return `false` rather than raising an error when given `'a'`.  See 7838ca3d4254c168d49ad06783d364ef4b4f0bc5 as an example of how we've done this for other matchers.  Want to work up a separate PR that fixes the `be <` matcher?

Also, if you want to work up a PR to improve our `all` docs to mention that we have `include` instead of `any` we could merge that.

dg-ratiodata:
> include already provides the same capability. expect(list).to include(a_value > 10) passes if list.any? { |x| x > 10 }.

Will this work with `it { expect(all_commands).to include have_output("foo") }`? If yes, does it makes sense to add a new alias `have_at_least_one_member` to make it more readable?

>  expect(list).to all be_odd reads like English; but expect(list).to any be_odd is completely nonsensical to a native English speaker. (I'm actually not sure of the grammatical reasons for why that is.)

Mmh, if this really is the problem, I could add some aliases for this:

```ruby
it { expect([1, 3, 11]).to have_any_member( be < 10 ) }
it { expect([1, 3, 11]).to have_a_member( be < 10 ) }
it { expect([1, 3, 11]).to have_at_least_one_member( be < 10 ) }
```

> Want to work up a separate PR that fixes the be < matcher?
> Also, if you want to work up a PR to improve our all docs to mention that we have include instead of any we could merge that.

Yes, will do that for both tomorrow.

myronmarston:
> Will this work with `it { expect(all_commands).to include have_output("foo") }`?

Yes.

>  If yes, does it makes sense to add a new alias `have_at_least_one_member` to make it more readable?

I don't consider it more readable, but if you want that alias, it's one line of code to add it:

``` ruby
RSpec::Matchers.alias_matcher :have_at_least_one_member, :include
```

maxmeyer:
Ok, perfect. Thanks.

dg-ratiodata:
Ok. I send you 2 PRs.

PR #808, opened 2015-06-23
Add some documentation for custom matchers

dg-ratiodata:
Hi this is a follow up to #807,
this adds some more documentation than requested, but I hope you find that useful. 

I tried to document all the input you - @myronmarston - gave me on different issues here on github. I found that very helpful and thought that might be helpful for others as well.

### DIFF
--- a/features/built_in_matchers/all.feature
+++ b/features/built_in_matchers/all.feature
@@ -1,7 +1,6 @@
 Feature: `all` matcher
 
-  Use the `all` matcher to specify that a collection's objects all pass an expected matcher.
-  This works on any enumerable object.
+  Use the `all` matcher to specify that a collection's objects all pass an expected matcher. This works on any enumerable object.
 
     ```ruby
     expect([1, 3, 5]).to all( be_odd )
@@ -16,6 +15,8 @@ Feature: `all` matcher
     expect([1, 3, 5]).to all( be_odd.and be < 10 )
     expect([1, 4, 21]).to all( be_odd.or be < 10 )
     ```
+
+  If you are looking for "any" member of a collection that passes an expectation, look at the `include`-matcher.
 
   Scenario: array usage
     Given a file named "array_all_matcher_spec.rb" with:
--- a/features/built_in_matchers/include.feature
+++ b/features/built_in_matchers/include.feature
@@ -1,7 +1,6 @@
 Feature: `include` matcher
 
-  Use the `include` matcher to specify that a collection includes one or more expected objects.
-  This works on any object that responds to `#include?` (such as a string or array):
+  Use the `include` matcher to specify that a collection includes one or more expected objects. It succeeds if any object of the given collection passes the specified matcher.. This works on any object that responds to `#include?` (such as a string or array):
 
     ```ruby
     expect("a string").to include("a")
@@ -11,6 +10,9 @@ Feature: `include` matcher
 
     expect([1, 2]).to include(1)
     expect([1, 2]).to include(1, 2)
+    expect([1, 2]).to include(a_kind_of(Integer))
+    expect([1, 2]).to include(be_odd.and be < 10 )
+    expect([1, 2]).to include(be_odd)
     expect([1, 2]).not_to include(17)
     ```
 
@@ -35,11 +37,16 @@ Feature: `include` matcher
         it { is_expected.to include(7) }
         it { is_expected.to include(1, 7) }
         it { is_expected.to include(1, 3, 7) }
+        it { is_expected.to include(a_kind_of(Integer)) }
+        it { is_expected.to include(be_odd.and be < 10) }
+        it { is_expected.to include(be_odd) }
+        it { is_expected.not_to include(be_even) }
         it { is_expected.not_to include(17) }
         it { is_expected.not_to include(43, 100) }
 
         # deliberate failures
         it { is_expected.to include(4) }
+        it { is_expected.to include(be_even) }
         it { is_expected.not_to include(1) }
         it { is_expected.not_to include(3) }
         it { is_expected.not_to include(7) }
@@ -52,7 +59,7 @@ Feature: `include` matcher
       """
     When I run `rspec array_include_matcher_spec.rb`
     Then the output should contain all of these:
-      | 14 examples, 7 failures                       |
+      | 19 examples, 8 failures                       |
       | expected [1, 3, 7] to include 4               |
       | expected [1, 3, 7] not to include 1           |
       | expected [1, 3, 7] not to include 3           |
--- a/features/custom_matchers/define_diffable_matcher.feature
+++ b/features/custom_matchers/define_diffable_matcher.feature
@@ -1,7 +1,6 @@
 Feature: define diffable matcher
 
-  When a matcher is defined as diffable, the output will include a diff of the submitted
-  objects when the objects are more than simple primitives.
+  When a matcher is defined as diffable, the output will include a diff of the submitted objects when the objects are more than simple primitives.
 
   Scenario: define a diffable matcher
     Given a file named "diffable_matcher_spec.rb" with:
@@ -27,3 +26,47 @@ Feature: define diffable matcher
              +two
               lines
       """
+
+  Scenario: Redefine actual
+
+    Sometimes is neccessary to overwrite actual to make diffing work, e.g. if `actual` is a name of a file you want to read from. For this to work you need to overwrite `@actual` in your matcher.
+
+    Given a file named "redefine_actual_matcher_spec.rb" with:
+      """ruby
+      RSpec::Matchers.define :have_content do |expected|
+        match do |actual|
+          @actual = File.read(actual).chomp
+
+          values_match? expected, @actual
+        end
+
+        diffable
+      end
+
+      RSpec.describe 'Compare files' do
+        context 'when content is equal' do
+          it { expect('data.txt').to have_content 'Data' }
+        end
+
+        context 'when files are different' do
+          it { expect('data.txt').to have_content "No\nData\nhere" }
+        end
+      end
+      """
+    And a file named "data.txt" with:
+    """
+    Data
+    """
+    When I run `rspec ./redefine_actual_matcher_spec.rb --format documentation`
+    Then the exit status should not be 0
+    And the output should contain:
+    """
+    2 examples, 1 failure
+    """
+    And the output should contain:
+    """
+           @@ -1,4 +1,2 @@
+           -No
+            Data
+           -here
+    """
--- a/features/custom_matchers/define_matcher.feature
+++ b/features/custom_matchers/define_matcher.feature
@@ -1,9 +1,8 @@
-Feature: define matcher
+Feature: Define a custom matcher
 
-  rspec-expectations provides a DSL for defining custom matchers. These are often useful for
-  expressing expectations in the domain of your application.
+  rspec-expectations provides a DSL for defining custom matchers. These are often useful for expressing expectations in the domain of your application.
 
-  Scenario: define a matcher with default messages
+  Scenario: Define a matcher with default messages
     Given a file named "matcher_with_default_message_spec.rb" with:
       """ruby
       require 'rspec/expectations'
@@ -44,7 +43,7 @@ Feature: define matcher
     And the output should contain "expected 9 to be a multiple of 4"
     And the output should contain "expected 9 not to be a multiple of 3"
 
-  Scenario: overriding the failure_message
+  Scenario: Overriding the failure_message
     Given a file named "matcher_with_failure_message_spec.rb" with:
       """ruby
       require 'rspec/expectations'
@@ -68,7 +67,7 @@ Feature: define matcher
     And the stdout should contain "1 example, 1 failure"
     And the stdout should contain "expected that 9 would be a multiple of 4"
 
-  Scenario: overriding the failure_message_when_negated
+  Scenario: Overriding the failure_message_when_negated
     Given a file named "matcher_with_failure_for_message_spec.rb" with:
       """ruby
       require 'rspec/expectations'
@@ -92,7 +91,7 @@ Feature: define matcher
     And the stdout should contain "1 example, 1 failure"
     And the stdout should contain "expected that 9 would not be a multiple of 3"
 
-  Scenario: overriding the description
+  Scenario: Overriding the description
     Given a file named "matcher_overriding_description_spec.rb" with:
       """ruby
       require 'rspec/expectations'
@@ -120,7 +119,7 @@ Feature: define matcher
     And the stdout should contain "should be multiple of 3"
     And the stdout should contain "should not be multiple of 4"
 
-  Scenario: with no args
+  Scenario: With no args
     Given a file named "matcher_with_no_args_spec.rb" with:
       """ruby
       require 'rspec/expectations'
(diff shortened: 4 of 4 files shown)


dg-ratiodata:
Ping @myronmarston @JonRowe

features/built_in_matchers/include.feature
```diff
@@ -1,7 +1,8 @@
 Feature: `include` matcher
 
-  Use the `include` matcher to specify that a collection includes one or more expected objects.
-  This works on any object that responds to `#include?` (such as a string or array):
+  Use the `include` matcher to specify that a collection includes one or more
+  expected objects. You can also think of if as an `any`-matcher. This works on
```
JonRowe: s/if/it

features/built_in_matchers/include.feature
```diff
@@ -11,6 +12,9 @@ Feature: `include` matcher
 
     expect([1, 2]).to include(1)
     expect([1, 2]).to include(1, 2)
+    expect([1, 2]).to include(be_an(Integer))
```
JonRowe: This reads oddly and given the examples below is a bit superfluous, maybe drop it?

features/built_in_matchers/include.feature
```diff
@@ -35,11 +39,16 @@ Feature: `include` matcher
         it { is_expected.to include(7) }
         it { is_expected.to include(1, 7) }
         it { is_expected.to include(1, 3, 7) }
+        it { is_expected.to include(be_an(Integer)) }
```
JonRowe: Ditto

features/built_in_matchers/all.feature
```diff
@@ -17,6 +17,9 @@ Feature: `all` matcher
     expect([1, 4, 21]).to all( be_odd.or be < 10 )
     ```
 
+  If you think, we might forgot about an `any`-matcher. No, it is just named
+  differently. Please have a look for our `include`-matcher.
```
JonRowe: This reads weirdly to me, perhaps consider rewording, something like: 
> If you are looking for any member of a collection that passes an expectation, look at the `include` matcher.

features/custom_matchers/define_matcher.feature
```diff
@@ -335,3 +335,109 @@ Feature: define matcher
       | 4 examples, 2 failures               |
       | expected 9 to be a multiple of 2     |
       | expected 9 not to be a multiple of 3 |
+
+  Scenario: compare values
+
+    `RSpec` comes with some handy helper methods. One is `values_match?` which
+    is available if you define your matcher using the "matcher"-DSL.
```
JonRowe: We actually start the documentation above, so I feel this should just be named appropriately, like: 

> Scenario: Using `values_match?` to compare values.

features/custom_matchers/define_matcher.feature
```diff
@@ -335,3 +335,109 @@ Feature: define matcher
...
+
+    The order of arguments is important for `values_match?`, e.g. especially if
+    your matcher should handle `Regexp`-objects (`/regex/`): First comes the
+    `expected` value, second the `actual` one.
```
JonRowe: This could then be moved into a code comment for the example.

features/custom_matchers/define_matcher.feature
```diff
@@ -335,3 +335,109 @@ Feature: define matcher
...
+    When I run `rspec ./compare_values_spec.rb --format documentation`
+    Then the exit status should be 0
+
+  Scenario: Error handling
+
+    Make sure your matchers returns either `true` or `false`. Take care to handle
+    exceptions appropriately in your matcher, e.g. most cases you might want
+    your matcher to return `false` if an exception - e.g. ArgumentError -
+    occures, but there might be edge cases where you want to pass the exception
+    to the user.
```
JonRowe: This might be more useful at the preamble.

features/built_in_matchers/include.feature
dg-ratiodata: Mmh... It's different from the other ones.

```ruby
        it { is_expected.to include(be_odd.and be < 10) } # combined matcher
        it { is_expected.to include(be_odd) } # call just odd?
        it { is_expected.to include(be_an(Integer)) } # full matcher checking type
```

What a about an alias for include, like the one you propose on the other PR? `have_a_member`?

features/custom_matchers/define_matcher.feature
dg-ratiodata: It was part of the preamble in one of my first versions. But I thought this might move users in the wrong direction and litters the preamble.

features/custom_matchers/define_matcher.feature
dg-ratiodata: I do not understand.

dg-ratiodata: Ok. I think I got it.

features/custom_matchers/define_matcher.feature
dg-ratiodata: Moved it into the preamble.

lib/rspec/matchers.rb
```diff
@@ -646,6 +646,7 @@ def include(*expected)
     alias_matcher :a_string_including,     :include
     alias_matcher :a_hash_including,       :include
     alias_matcher :including,              :include
+    alias_matcher :have_a_member,          :include
```
dg-ratiodata: I think this one needs to be "highlighted". Is there any other good suggestion to make `it { is_expected.to include(be_an(Integer)) }` sound "good"? Maybe `it { is_expected.to have_a_member(be_an(Integer)) }`?

myronmarston: I don't want to add `have_a_member` to RSpec.  Any user who wants it can add it with a one-liner.

> Is there any other good suggestion to make it { is_expected.to include(be_an(Integer)) }` sound "good"? Maybe `it { is_expected.to have_a_member(be_an(Integer)) }`?

Sure the "noun-phrase" aliases of the matchers are specifically there to read well for situations like this.  You can use:

``` ruby
it { is_expected.to include a_kind_of Integer }
```

features/built_in_matchers/all.feature
```diff
@@ -17,6 +17,9 @@ Feature: `all` matcher
     expect([1, 4, 21]).to all( be_odd.or be < 10 )
     ```
 
+  If you are looking for "any" member of a collection that passes an
+  expectation, look at the `include`-matcher.
```
myronmarston: Relish wraps at the same places we do in the source, unfortunately, so we're moving to making each paragraph one lone unwrapped line in the source, which allows normal HTML flow to wrap it so that it looks good.

features/built_in_matchers/include.feature
```diff
@@ -1,7 +1,8 @@
 Feature: `include` matcher
 
-  Use the `include` matcher to specify that a collection includes one or more expected objects.
-  This works on any object that responds to `#include?` (such as a string or array):
+  Use the `include` matcher to specify that a collection includes one or more
+  expected objects. You can also think of it as an `any`-matcher. This works on
+  any object that responds to `#include?` (such as a string or array):
```
myronmarston: Same here -- please don't wrap this.  Make it one long line so it renders better on relish.

features/custom_matchers/define_diffable_matcher.feature
```diff
@@ -27,3 +27,49 @@ Feature: define diffable matcher
              +two
               lines
       """
+
+  Scenario: Redefine actual
+
+    Sometimes is neccessary to overwrite actual to make diffing work, e.g. if
+    `actual` is a name of a file you want to read from. For this to work you
+    need to overwrite `@actual` in your matcher.
```
myronmarston: Does cucumber support prose between the scenario title and the first step like this?  Even if cucumber supports it, I'm not sure how it'll look on relish...

features/built_in_matchers/include.feature
imtayadeway: just my two penneth here, but i'm not sure how i feel about thinking of `include` as an `any`-matcher, since we can do this with enumerables:

```
expect([]).not_to be_any
```

(though it might not make complete sense). i guess i just find thinking of a matcher in terms of some other matcher a little confusing.

maxmeyer: > how i feel about thinking of include as an any-matcher, 

Mmmh... My thinking was that of a user. I saw that @myronmarston added the `all`-matcher sometime ago, but asked myself where the heck I find the `any`-matcher. I was not aware that the semantics of the `include`-matcher are the same until @myronmarston told me so. This is there reason why I added this stuff here.

maxmeyer: > Same here -- please don't wrap this. Make it one long line so it renders better on relish.

Ok. Will fix this. I was not aware that this is a problem on relish.

features/built_in_matchers/all.feature
maxmeyer: Ok, will fix this.

lib/rspec/matchers.rb
maxmeyer: > I don't want to add have_a_member to RSpec. Any user who wants it can add it with a one-liner.

No, problem at all. :smile: Will suggest it again. Will fix the example by using your suggestion.

features/custom_matchers/define_diffable_matcher.feature
maxmeyer: > Does cucumber support prose between the scenario title and the first step like this? 
This works. All tests pass

> Even if cucumber supports it, I'm not sure how it'll look on relish...
@mattwynne Can you us help here?

features/built_in_matchers/include.feature
dg-ratiodata: @imtayadeway I fixed with wording. Now only `any` as word is mentioned if one tries to search for `any` in the web browser.

features/custom_matchers/define_diffable_matcher.feature
dg-ratiodata: @myronmarston Do you have a "test" publisher available on relish where we can try to release the new version to see if it looks good with documentation after the scenario-title?

dg-ratiodata:
Ok. Fixed the wording at some places and deleted the line wrap.

features/custom_matchers/define_diffable_matcher.feature
JonRowe: I think we just need to merge this and see how it looks after publishing

dg-ratiodata:
@JonRowe Did you push it already on relish? or is there a pull job running anywhere?

dg-ratiodata:
Thanks for merging.

JonRowe:
Nope, @myronmarston needs to do that AFAIK, we do have a staging environment for this to be tried against tho

dg-ratiodata:
> staging environment

ok. is this available to the public?

JonRowe:
Yes, but it also needs manually updating so... https://relishapp.com/rspec-staging/

maxmeyer:
Great. Bookmarked. Ok. I think he will be online later that day (my time)... It's 9 o'clock here.

myronmarston:
I pushed updated docs to the rspec-staging account on relish.

JonRowe:
Well it appears but I worry it gets lost amongst the scenarios...
screen shot 2015-06-24 at 08 21 27

dg-ratiodata:
The comments look not that good...

unbenannt

dg-ratiodata:
> Well it appears but I worry it gets lost amongst the scenarios...

Mmmh... Otherwise the "context" is lost.

dg-ratiodata:
@myronmarston Do you know anyone else than @mattwynne who can help us here. Last time we wrote mails, he was quite busy. 

From my perspective it's more a "design issue". The text is not that separated from the steps than it should be. But moving that information into the preamble makes it harder to understand because the example is further down the page.

unbenannt

dg-ratiodata:
Is there any chance to add some "custom css" to relish documentation? "The problem" is this `<p></p>`-section.

unbenannt

myronmarston:
> @myronmarston Do you know anyone else than @mattwynne who can help us here.

Nope, sorry :(.  That said, when I brought up a rendering issue before he gave me commit access so I could look into fixing it myself.  I didn't have the time or energy to contribute to relish (and still don't) but if you wanted to help out he might be willing to give you commit access.

> Is there any chance to add some "custom css" to relish documentation? "The problem" is this <p></p>-section.

Looks like something that could be fixed for the entire site and not just by us if you get the chance to submit a PR :).

dg-ratiodata:
What do you think about that?

unbenannt

> Looks like something that could be fixed for the entire site and not just by us if you get the chance to submit a PR :).

:smile:

myronmarston:
Works for me.

dg-ratiodata:
Hi @mattwynne are you will to change the rendering of `relish` a bit? Where should I open a PR for that? Is it a private repository?

Given the following cucumber scenario:

```gherkin
Feature: Blub
  Scenario: Error handling

    Make sure your matcher returns either `true` or `false`. Take care to handle exceptions appropriately in your matcher, e.g. most cases you might want your matcher to return `false` if an exception - e.g. ArgumentError - occures, but there might be edge cases where you want to pass the exception to the user.

    You should handle each `StandardError` with care! Do not handle them all in one.

    \`\`\`ruby
    match do |actual|
      begin
        '[...] Some code'
      rescue ArgumentError
        false
      end
    end
    \`\`\`
# [...]
```
The rendering for the scenario preamble is not optimal.

unbenannt

With this kind of `css`

```css
dd.scenario-steps p {
  margin: 10px;
  background-color: #F8F8F8;
  padding: 5px;
  border-radius: 5px;
  border-left: 3px solid #D8D8D8;
}
```
it is easier to read:

unbenannt

Maybe we/you could even wrap each element of the preamble in a div. What do you think?

mattwynne:
Hi all,

I don’t understand the problem here, sorry. If someone can explain clearly what you’d like changed I can adjust the CSS or maybe even the generated HTML for the scenario.

We’re not adding any new features to Relish, all our efforts are concentrated on Cucumber Pro. I hope we’ll have RSpec publishing to Cucumber Pro soon enough.

dg-ratiodata:
The problem is this:

When one wrote a preample for a scenario, the rendered HTML makes it hard to distinguish the preamble from the first step.

```gherkin
Feature: Blub
  Scenario: Scenario #1
   This is paragraph #1

   This is paragraph #2

   Given I have a lot of money
   Then I might be happy
```
This now renders to 

```html
<!-- [...] -->
<dd class="scenario-steps">
<p>
   This is paragraph #1
</p>
<p>
  This is paragraph #2
</p>
<dl class="step-list">
<!-- [...] -->
</dl>
</dd>
<!-- [...] -->
```

Wrapping the rendered markdown in a `<div class="scenario-preamble"></div>`...

```html
<!-- [...] -->
<dd class="scenario-steps">
<div class="scenario-preamble">
<p>
   This is paragraph #1
</p>
<p>
  This is paragraph #2
</p>
</div>
<dl class="step-list">
<!-- [...] -->
</dl>
</dd>
<!-- [...] -->
```
...  plus this css would make it easier to read.

```css
dd.scenario-steps scenario-preamble {
  margin: 10px;
  background-color: #F8F8F8;
  padding: 5px;
  border-radius: 5px;
  border-left: 3px solid #D8D8D8;
}
```

Hope that clarifies the idea?

mattwynne:
@dg-ratiodata your suggested change is now live on Relish. I'm re-generating the docs as we speak, but https://relishapp.com/rspec-staging/rspec-expectations is done.

myronmarston:
Thanks, @mattwynne!

I'm noticing that the nav has a bunch of duplicate entries.  Is that related to the re-generation you just did?

screen shot 2015-08-14 at 2 18 42 pm

mattwynne:
Odd, I don’t see that. What URL are you looking at?

myronmarston:
> Odd, I don’t see that. What URL are you looking at?

https://relishapp.com/rspec-staging/rspec-expectations/docs but it's gone now.  Must have been some caching problem.

maxmeyer:
@mattwynne Great. Thanks.

PR #809, opened 2015-06-23
Fixed bug in be_*-matcher

dg-ratiodata:
Hi this is a also a follow up to #807,

I added a `begin; rescue; end` block to rescue `ArgumentError`. Should I also rescue `NoMethodError`? I thought it would be better not to rescue this one, because it is a severe user error which should not be rescued.

### DIFF
--- a/features/built_in_matchers/comparisons.feature
+++ b/features/built_in_matchers/comparisons.feature
@@ -7,6 +7,7 @@ Feature: Comparison matchers
     expect(9).to be > 6
     expect(3).to be <= 3
     expect(1).to be < 6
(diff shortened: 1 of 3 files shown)


dg-ratiodata:
Ping @myronmarston @JonRowe

JonRowe:
Yeah this is ok, but I wouldn't rescue `NoMethodError`

dg-ratiodata:
Thanks a lot.

myronmarston:
@JonRowe -- can you add a changelog for this?

JonRowe:
I was trying to earlier but flakey train wifi :/

PR #813, opened 2015-07-05
Pin aruba to work around cucumber/aruba#279.

myronmarston:


myronmarston:
The build failures are from the rspec-core part of the build, which'll be fixed by rspec/rspec-core#2022. Merging.

Issue #814, opened 2015-07-06
stack level too deep (SystemStackError) with rspec-expectations


maxmeyer:
Hi @myronmarston,

I'm on my way down the rabbit hole to support 1.9.3. and 1.8.7 for aruba. Now I'm facing this error every time I use a negated alias for a matcher on ruby 1.9.3. I use rspec-expectation 3.3.

```
/home/user/.rvm/gems/ruby-1.9.3-p551/gems/rspec-expectations-3.3.0/lib/rspec/matchers.rb:972: stack level too deep (SystemStackError)
```

and I'm not sure if it's a bug or a mistake by me. This is a minimum failing example which works fine on 1.8.7 and > 2.0, but not on 1.9.3. I think rspec's code takes a "different route" for 1.8.7, so maybe it's the way ruby 1.9.3 behaves in contrast to > 2.0.

```ruby
require 'rspec/expectations'

RSpec::Matchers.define :run_too_long do
  match do |actual|
    true
  end
end

RSpec::Matchers.define_negated_matcher :have_finished_in_time, :run_too_long

include RSpec::Matchers

expect('command').to run_too_long
$stderr.puts 'run too long works'

expect('command').to have_finished_in_time
$stderr.puts 'have finished in time not'
```

These are the installed gems:

```
Gems included by the bundle:
  * archive-tar-minitar (0.5.2)
  * aruba (0.8.0)
  * ast (2.0.0)
  * astrolabe (1.3.0)
  * bcat (0.6.2)
  * binding_of_caller (0.7.2)
  * builder (3.2.2)
  * bundler (1.10.5)
  * childprocess (0.5.6)
  * coderay (1.1.0)
  * columnize (0.9.0)
  * contracts (0.9)
  * cucumber (1.3.20)
  * cucumber-pro (0.0.17)
  * debug_inspector (0.0.2)
  * debugger (1.6.8)
  * debugger-linecache (1.2.0)
  * debugger-ruby_core_source (1.3.8)
  * diff-lcs (1.2.5)
  * docile (1.1.5)
  * domain_name (0.5.24)
  * eventmachine (1.0.7)
  * faye-websocket (0.9.2)
  * ffi (1.9.10)
  * fuubar (2.0.0)
  * gherkin (2.12.2)
  * http-cookie (1.0.2)
  * httparty (0.13.5)
  * json (1.8.3)
  * kramdown (1.7.0)
  * license_finder (2.0.4)
  * method_source (0.8.2)
[...]

PR #815, opened 2015-07-06
WIP: failing spec.

myronmarston:
This is a failing spec for #814.  I'm not sure of a fix yet.


JonRowe:
It is feasible to work round this with a bit of tinkering, but I don't think we should given its a ruby bug (as you point out in #814), what we could do is add a warning on 1.9.3 that explains whats going on?

myronmarston:
@JonRowe -- yep.  See #816.

PR #821, opened 2015-07-15
Fix `respond_to` for targets that redefine `method`

unmanbearpig:
Previously it would crash or not work properly when we were trying to
check arity of any method on target object that had overridden the
`method` method.

### DIFF
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -72,7 +72,7 @@ module RSpec
         def matches_arity?(actual, name)
           return true unless @expected_arity
 
-          signature = Support::MethodSignature.new(actual.method(name))
+          signature = Support::MethodSignature.new(get_method_on_object(actual, name))
           Support::StrictSignatureVerifier.new(signature, Array.new(@expected_arity)).valid?
         end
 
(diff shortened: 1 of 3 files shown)


unmanbearpig:
I'm not sure if I put the test into the right place, since it doesn't matter how many arguments we're expecting, as long as we're checking arity. Please let me know if I should move it somewhere else.

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -84,6 +84,11 @@ def with_arity
         def pp_names
           @names.length == 1 ? "##{@names.first}" : description_of(@names)
         end
+
+        def get_method_on_object(object, method_name)
+          # Can't use object.method since it might be overridden
+          Kernel.instance_method(:method).bind(object).call(method_name)
```
myronmarston: We have a method in rspec-support that handles this plus several other edge cases:

https://github.com/rspec/rspec-support/blob/92e571c80b23d6ba643ec2d8372126e36deddf88/lib/rspec/support.rb#L35-L74

Please use `RSpec::Support.method_handle_for` instead.

myronmarston:
Thanks, @unmanbearpig.  This looks great.  Once you've addressed the comment I left above I'll be happy to merge.

unmanbearpig:
Awesome! Done.

myronmarston:
Thanks!

Issue #759, opened 2015-03-30
`have_attributes` attribute detail is missing when passed to another matcher


myronmarston:
In #668, the `have_attributes` matcher was improved so that the actual attribute values are used in the diff.  This is particularly important when the `inspect` output of the object does not include the attribute values.

However, the issue from #667 remains when you pass `have_attributes` (or an alias) to another matcher. Consider this spec:

``` ruby
RSpec.describe "have_attributes" do
  example "using `have_attributes` directly" do
    expect(1).to have_attributes(:__id__ => 4)
  end

  example "passing `an_object_having_attributes` to another matcher" do
    expect([1, 2]).to match [
      an_object_having_attributes(:__id__ => 4),
      an_object_having_attributes(:__id__ => 5)
    ]
  end
end
```

Running that produces two failures:

```
  1) have_attributes using `have_attributes` directly
     Failure/Error: expect(1).to have_attributes(:__id__ => 4)
       expected 1 to have attributes {:__id__ => 4} but had attributes {:__id__ => 3}
       Diff:
       @@ -1,2 +1,2 @@
       -:__id__ => 4,
       +:__id__ => 3,

  2) have_attributes passing `an_object_having_attributes` to another matcher
     Failure/Error: expect([1, 2]).to match [
       expected [1, 2] to match [(an object having attributes {:__id__ => 4}), (an object having attributes {:__id__ => 5})]
```

The failure from the first spec clearly shows that the value of `__id__` was that did not match.  The failure from the 2nd spec doesn't show that at all, which is annoying because it's hard to tell why the match failed without that information.

mrageh:
@myronmarston I'll start to think off a general solution for `have_attributes` matcher

nellirx:
Any ideas what output we'd like to have when passing to another matcher, i.e. 2nd example?
Would something like this be good (**...but had attributes {:__id__ => 3}**):

```ruby
2) have_attributes passing `an_object_having_attributes` to another matcher
     Failure/Error: expect([1]).to match [
       expected [1] to match [(an object having attributes {:__id__ => 4} but had attributes {:__id__ => 3})]
```

pirj:
@nellirx That would certainly be an improvement. Would you like to hack on it?

nellirx:
> Would you like to hack on it?

yes please, thanks

PR #836, opened 2015-08-01
[Draft] clean up error message with change and has_attributes

AEgan:
I've been working on a way to clean up the error message when a test fails using both the change and has_attributes matchers. Currently, the test

``` ruby
expect { person.age = 10 }.to change{ person }.to have_attributes({:age => 6})
```

will fail with the message `expected result to have changed to have attributes {:age => 6}, but is now #<struct Person name="Correct name", age=10>`

This can become annoying when an object has a lot of attributes because you have to parse through the output to find the attribute you pass into `have_attributes`

This change modifies the error message to say `expected result to have changed to have attributes {:age => 6}, but had attributes {:age => 10}`

While this is the behavior I was looking for, the implementation could be improved. I was wondering if anyone had any ideas on ways to clean this up, and not rely on a special case.

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -176,6 +176,7 @@ module RSpec
           return not_given_a_block_failure unless Proc === @event_proc
           return before_value_failure      unless matches_before?
           return did_not_change_failure    unless @change_details.changed?
+          return after_value_failure_with_attributes if change_with_has_attributes
           after_value_failure
         end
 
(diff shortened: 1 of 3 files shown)


myronmarston:
Thanks for taking a stab at this!  You're absolutely right that the failure message you were getting leaves something to be desired. In fact, this is a general problem with composable matchers that we don't yet have a good solution for: it's easy to use the `description` of the passed matcher in the failure message but failure message of the passed matcher may include important details that would improve the output.  See #759 for another example of this.

Here you've solved the problem specifically for when `have_attributes` is passed to `change` but the problem remains when `have_attributes` is passed to _any other_ matcher and when _any other_ matcher is passed to `change`.  I don't really want to go down the route of special casing matchers like this -- it's going to quickly become a maintenance nightmare.   It's also potentially surprising: why does `change` treat `have_attributes` as special but not other matchers?  That leads to inconsistent behavior.

Do you have any ideas for how to solve this in a more generic way such that we're not special casing matchers?

Regardless, I'm going to close this -- any solution we'd merge for this problem is going to be quite different from this.

Thanks again!

PR #840, opened 2015-08-19
Upgrade aruba

dg-ratiodata:
Let's see what travis tell us. Locally it ran fine.

### DIFF
--- a/features/step_definitions/additional_cli_steps.rb
+++ b/features/step_definitions/additional_cli_steps.rb
@@ -10,7 +10,7 @@ Then /^the output should contain all of these:$/ do |table|
     if RUBY_VERSION == '1.8.7' && string =~ /\{.+=>.+\}/
       warn "Skipping checking #{string} on 1.8.7 because hash ordering is not consistent"
     else
-      assert_partial_output(string, all_output)
+      expect(last_command_stopped).to have_output Regexp.new(Regexp.escape(extract_text(unescape_text(string))))
     end
   end
 end
--- a/features/support/disallow_certain_apis.rb
+++ b/features/support/disallow_certain_apis.rb
@@ -4,7 +4,7 @@
 if defined?(Cucumber)
   require 'shellwords'
   Before('~@allow-disallowed-api') do
-    set_env('SPEC_OPTS', "-r#{Shellwords.escape(__FILE__)}")
+    set_environment_variable('SPEC_OPTS', "-r#{Shellwords.escape(__FILE__)}")
   end
 else
   module DisallowOneLinerShould
--- a/rspec-expectations.gemspec
+++ b/rspec-expectations.gemspec
@@ -39,6 +39,6 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'rake',     '~> 10.0.0'
   s.add_development_dependency 'cucumber', '~> 1.3'
-  s.add_development_dependency "aruba",    "~> 0.6.2" # 0.7 is broken on ruby 1.8.7
+  s.add_development_dependency "aruba",    "~> 0.9.0" # 0.7 is broken on ruby 1.8.7
   s.add_development_dependency 'minitest', '~> 5.2'
 end


features/step_definitions/additional_cli_steps.rb
```diff
@@ -10,7 +10,7 @@
     if RUBY_VERSION == '1.8.7' && string =~ /\{.+=>.+\}/
       warn "Skipping checking #{string} on 1.8.7 because hash ordering is not consistent"
     else
-      assert_partial_output(string, all_output)
+      expect(last_command_stopped).to have_output Regexp.new(Regexp.escape(extract_text(unescape_text(string))))
```
myronmarston: So has `assert_partial_output` been removed?  This new form seems much much messier, given that it requires:

* `unescape_text`
* `extract_text`
* `Regexp.escape`
* `Regexp.new`

...all passed to `have_output`.

I don't think I'd ever be able to come up with that and even now I don't understand the reason for all of it.

dg-ratiodata:
> So has assert_partial_output been removed? 

Not yet, but it will be removed with 1.0.0. Now all `assert_*`-methods are deprecated.

> This new form seems much much messier

I'm every well aware that. Unfortunately today I've no idea how to solve it differently because I would like to have code which is 
* easy to extend, single place for code
* flexible in use, handles partial and full matching
* has only one interface to do similar things, e.g `expect(command).to have_output /partial/`, `expect(command).to have_output 'full'`

> I don't think I'd ever be able to come up with that and even now I don't understand the reason for all of it.


1.  Mix of `assert_*`, `check_*`-methods + even some `expect()`-calls

  If you have a look at the old `aruba`-sources, you will see, that there was a mix of `assert_*`, `check_*`-methods + even some `expect()`-calls from RSpec. We decided to replace all this through `RSpec`-matchers to make the code easier to maintain and to use.

2. Following the Single responsibility principle

  Having a look at `#unescape` you will see that it does two things:

  * unescaping \n \e and the like
  * extracting text from strings containing ANSI escape sequences

  To make those use cases clear, I split up the method. Plus it was originally bound to some instance variable which controls if ansi escape sequences should be removed which is now controlled via a configuration option.

3. Using matchers + `#values_match?`

  There were quite a few methods, which "have a look" at the same thing, but only matching `partial` or `full`.  Using `#values_match?` in our matcher we could get rid of all that methods. Although you may already know: _We get a string from `cucumber`, that's why we need to convert it to a Regex to make partial matching possible. 

  Most of those methods were only helper methods, which were used behind the scenes and littered the `aruba`-API.

> for all of it.

I'm not sure what you mean with "for all of it". Just in the context of the change given above or all changes made recently to `aruba`? If you like and have *time* we can discuss this in IRC or even by voice call? I would like to understand your concerns. If you prefer, I could ask @mattwynne to be around as well.

To sum it up: If you have any suggestion I happy to take that into account.

Cheers,
Dennis

dg-ratiodata:
Maybe I've got an implementation which fits your needs better. Will update this PR soon.

myronmarston:
> If you like and have time we can discuss this in IRC or even by voice call? I would like to understand your concerns. If you prefer, I could ask @mattwynne to be around as well.

I don't have the time, sorry.

Anyhow, my concern is basically that `assert_partial_output` was a good, simple API.  I understand the desire to move in the direction of more composable pieces (e.g. matchers) but could `assert_partial_output` remain and simply delegate to:

``` ruby
expect(...).to have_output Regexp.new(Regexp.escape(extract_text(unescape_text(string))))
```

IMO, that would be better and less disruptive to users.

Alternately, if `have_output` can handle all that complexity internally so that it is:

``` ruby
expect(...).to have_output(string)
```

Or would this work?

``` ruby
expect(...).to have_output(a_string_including(string))
```

?

dg-ratiodata:
Good morning... :-) The idea with `expect(...).to have_output(a_string_including(string))` sounds interesting. I will have a look into that. Thanks a lot.

dg-ratiodata:
In aruba#master are now some new [matchers](https://github.com/cucumber/aruba/tree/master/lib/aruba/matchers/string) hiding those messy steps.

I will re-open this PR when I released 0.10.0 and changed the implementation here.

myronmarston:
> I will re-open this PR when I released 0.10.0 and changed the implementation here.

:+1:

PR #842, opened 2015-08-22
_Technically_, it's 1.9 _plus_

xaviershay:
What's our 1.8 deprecation policy again? Drop it in 4?

@rspec/rspec


JonRowe:
I believe that's what we said...

myronmarston:
Yep.

Issue #425, opened 2014-01-15
Make ExpectationNotMetError subclass Exception


myronmarston:
See rspec/rspec-mocks#203 for background.

Also, if minitest is loaded, it should subclass minitest's assertion class so it is counted as a failure, not an exception.

PR #843, opened 2015-08-24
Reduce scope of rescue in `safe_sort`

danielfone:
Rescue `StandardError` instead of `Exception`. `StandardError` will almost certainly encapsulate every exception we want to rescue (including `NameError`), while `Exception` encapsulates numerous exceptions we probably don't want to rescue (including `Interrupt`).  

I stumbled across this and haven't checked the rest of the codebase for similar `rescue`s. If this is helpful, I can check for other instances of this.
 
Further discussion:
http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby
https://robots.thoughtbot.com/rescue-standarderror-not-exception
http://daniel.fone.net.nz/blog/2013/05/28/why-you-should-never-rescue-exception-in-ruby/ (My own thoughts)

### DIFF
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -61,7 +61,7 @@ module RSpec
 
         def safe_sort(array)
           array.sort
-        rescue Exception
+        rescue
           array
         end
 


danielfone:
Hmm, this passed locally. Will investigate further.

danielfone:
On closer inspection, in 1.9+ the test in question appears to raise `#<ArgumentError: comparison of RSpec::Mocks::Double with RSpec::Mocks::Double failed>` while 1.8 raises [`MockExpectationError`](https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/error_generator.rb#L6) which is a subclass of `Exception`. I couldn't trace the origin of this code back far enough, but it's been that way for [a long time](https://github.com/rspec/rspec-mocks/blame/3138bbab1686e3e56b42dea0b754273ac017c63a/lib/rspec/mocks/errors.rb#L4). For me this raises the question of whether `MockExpectationError` should be subclassing `Exception` directly, or whether it should inherit from `StandardError`.

myronmarston:
> For me this raises the question of whether MockExpectationError should be subclassing Exception directly, or whether it should inherit from StandardError.

`MockExpectationError` subclasses `Exception` because when a mock expectation fails, the user expects it to cause the example to fail, and if any user code has a `rescue => error` in there, it should not be able to interfere with that process.  We similarly changed rspec-expectations expectation failures to subclass `Exception` in RSpec 3.  See #426, #425 and rspec/rspec-mocks#203.

Anyhow, we are well aware of the "don't rescue Exception" advice commonly given, and we agree with it in general...but frameworks that invoke large quantities of user code like Rails, Sinatra and RSpec often need to not follow that advice.  For example: if Rails didn't rescue `Exception`, then an error in a controller or model could make the server crash rather than it rendering a `500` error.  Similarly, in RSpec, if some user code raises an error subclassed from `Exception`, and we didn't rescue `Exception`, it could cause RSpec to crash.

In this specific case, I think `rescue Exception` is what we want.  The sorting is merely an optimization (although, it's one that makes a significant difference in some cases -- the algorithm that gets used if sorting doesn't work can get quite slow on large inputs).  When a user writes `expect(collection).to contain_exactly(a, b)` the user isn't trying to test whether `a` and `b` can be sorted.  The fact that we try to sort is merely an implementation detail.  The user may have valid reasons for allowing `a <=> b` to raise an error subclassed from `Exception` and we shouldn't allow such a thing to interfere with answering the question, "does `collection` contain `a` and `b`"?  From a user perspective, it would look like a bug for `expect(collection).to contain_exactly(a, b)` to raise an error simply because `a` and `b` can't be sorted

Is the `rescue Exception` causing you a concrete problem, or are you trying to change this simply because it's generally bad to `rescue Exception`?

danielfone:
@myronmarston thanks for the detailed feedback!

> Is the rescue Exception causing you a concrete problem, or are you trying to change this simply because it's generally bad to rescue Exception?

I was working with an array of objects that had a very slow sort, and pressing ^C didn't interrupt the sorting the first time. When I investigated I found this line, hence the PR. Reproducing this now, I suspect this was a red herring, since this is consistent with RSpec's usual interrupt handling behaviour.

Thanks again for your time and sorry for the bother.

Issue #831, opened 2015-07-29
Add capturing support to the `match` built-in matcher


jasonkarns:
Some regexes are complicated enough to warrant their own specs, which naturally leads to specifying not only that the regex matches a given string, but that it captures the correct substrings.

It would be great if there were a method chain on `match()` that accepted an array or hash to match against the captures from the regex. (Array if using indexed capture groups, Hash if using named capture groups)

Example:

``` ruby
module Indexed
  COMPLEX_REGEX = /^.*(\d+).*$/
end
module Named
  COMPLEX_REGEX = /^.*(?<num>\d+).*$/
end

describe "complicated regex" do
  it "should capture the digits" do
    expect(Indexed::COMPLEX_REGEX).to match("foo123bar").capturing(["123"])
    expect(Named::COMPLEX_REGEX).to match("foo123bar").capturing(num: "123")
  end
end
```

The example above doesn't demonstrate it, but I would expect the matching of named captures (specified using a hash) to do normal hash matching, thus allowing nested matchers in the hash itself.

I have already created a custom matcher that inherits from `RSpec::Matchers::BuiltIn::Match` that supports chaining the `capturing` method to specify expected capture groups. https://gist.github.com/jasonkarns/0b4c9483d11be82bae56.  Description of how the matcher works is commented: https://gist.github.com/jasonkarns/0b4c9483d11be82bae56#gistcomment-1515425

(To be clear, the custom matcher linked above is a quick-fix attempt that covers my scenario. It does *not* handle indexed captures, only named captures matched using a hash. And it does some brittle ivar swapping to provide better error message which would not be suitable for production as-is.)

myronmarston:
I think this is a great idea, although I wonder if the chained method should be `with_captures` instead of `capturing` -- which would align better with `raise_error(...).with_message(...)`.  I don't feel strongly about the name, though.

Any desire to work on a PR for this?  My time these days to work on new features is unfortunately short :(.

jasonkarns:
I like `with_captures`.

I *might* be able to work on a PR. One area of concern is how this will interact with the other uses of `match` (when `actual` is the string instead of the regex, or when using `match` to compare deeply nested objects)

Hopefully I'll get enough ground work on the PR to at least expose the issues and generate some more discussion.

myronmarston:
> when `actual` is the string instead of the regex,

A `MatchData` object is returned in either case so it shouldn't be a problem:

```
[...]

PR #848, opened 2015-09-14
Add a match(...).with_captures(...) matcher.

samphippen:
See #831 for details/discussion.

### DIFF
--- a/spec/rspec/matchers/built_in/captures_spec.rb
+++ b/spec/rspec/matchers/built_in/captures_spec.rb
@@ -1,0 +1,87 @@
+RSpec.describe "expect(regex).to match(string).with_captures" do
+  context "with a string target" do
+    it "does match a regex with a missing capture" do
+      expect("a123a").to match(/(a)(b)?/).with_captures("a", nil)
+    end
+
+    it "does not match a regex with an incorrect match" do
+      expect("a123a").not_to match(/(a)/).with_captures("b")
+    end
+
+    it "matches a regex without named captures" do
+      expect("a123a").to match(/(a)/).with_captures("a")
+    end
+
+    it "uses the match description if the regex doesn't match" do
+      expect {
+        expect(/(a)/).to match("123").with_captures
+      }.to fail_with(/expected \/\(a\)\/ to match "123"/)
+    end
+
+    if RUBY_VERSION != "1.8.7"
+      it "matches a regex with named captures" do
+        expect("a123a").to match(Regexp.new("(?<num>123)")).with_captures(:num => "123")
+      end
+      it "matches a regex with a nested matcher" do
+        expect("a123a").to match(Regexp.new("(?<num>123)(asdf)?")).with_captures(hash_including(:num => "123"))
+      end
+
+      it "does not match a regex with an incorrect named group match" do
+        expect("a123a").not_to match(Regexp.new("(?<name>a)")).with_captures(:name => "b")
+      end
+
+      it "has a sensible failure description with a hash including matcher" do
+        expect {
+          expect("a123a").not_to match(Regexp.new("(?<num>123)(asdf)?")).with_captures(a_hash_including(:num => "123"))
+        }.to fail_with(/num => "123"/)
+      end
+
+      it "matches named captures when not passing a hash" do
+        expect("a123a").to match(Regexp.new("(?<num>123)")).with_captures("123")
+      end
+    end
+  end
+
+  context "with a regex target" do
+    it "does match a regex with a missing capture" do
+      expect(/(a)(b)?/).to match("a123a").with_captures("a", nil)
+    end
+
+    it "does not match a regex with an incorrect match" do
+      expect(/(a)/).not_to match("a123a").with_captures("b")
+    end
+
+    it "matches a regex without named captures" do
+      expect(/(a)/).to match("a123a").with_captures("a")
+    end
+
+    it "uses the match description if the regex doesn't match" do
+      expect {
+        expect(/(a)/).to match("123").with_captures
+      }.to fail_with(/expected \/\(a\)\/ to match "123"/)
+    end
+
+    if RUBY_VERSION != "1.8.7"
+      it "matches a regex with named captures" do
+        expect(Regexp.new("(?<num>123)")).to match("a123a").with_captures(:num => "123")
+      end
+      it "matches a regex with a nested matcher" do
+        expect(Regexp.new("(?<num>123)(asdf)?")).to match("a123a").with_captures(hash_including(:num => "123"))
+      end
+
+      it "does not match a regex with an incorrect named group match" do
+        expect(Regexp.new("(?<name>a)")).not_to match("a123a").with_captures(:name => "b")
+      end
+
+      it "has a sensible failure description with a hash including matcher" do
+        expect {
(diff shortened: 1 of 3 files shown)


samphippen:
So ruby 1.8.7 doesn't support named captures, does anyone have a thought on how we can avoid the syntax error here @rspec/rspec

samphippen:
@rspec/rspec got it working on 1.8.7, can I get a review?

lib/rspec/matchers/built_in/captures.rb
```diff
@@ -0,0 +1,77 @@
...
+        def description
+          return BuiltIn::Match.new(expected).description unless @expected_captures
+          "match with string #{@expected} have captures #{surface_descriptions_in(@expected_captures).inspect}"
+        end
+
+      private
+
+        def match(expected, actual)
+          return BuiltIn::Match.new(expected).matches?(actual) unless Regexp === actual
+          return BuiltIn::Match.new(expected).matches?(actual) unless @expected_captures
```
JonRowe: Although minor perf wise, this is a bunch of extra instantiations of the same thing (3 across the class)

lib/rspec/matchers/built_in/captures.rb
```diff
@@ -0,0 +1,77 @@
...
+        def initialize(match_data)
+          @match_data = match_data
+        end
+
+        # @api private
+        # Returns match data names for named captures
+        # @return Array
+        def names
+          match_data.names
+        rescue NameError
```
JonRowe: So is this the part that was incompatible with 1.8.7?

samphippen: This, and regular expressions cannot have named captures (so I have to use Regexp.new) instead of literals.

samphippen:
@JonRowe is this good to merge?

lib/rspec/matchers/built_in/captures.rb
myronmarston: I'd prefer to see the 1.8.7-specific logic extracted into a conditionally defined version of the method. That makes it obvious what the logic is for and that it can be removed when we drop support for 1.8.7. As this currently exists that is non-obvious.

lib/rspec/matchers.rb
```diff
@@ -698,7 +698,11 @@ def all(expected)
     #       matchers (due to how the custom matcher DSL was evaluated in 2.x,
     #       `match` could not be used there), but is no longer needed in 3.x.
     def match(expected)
-      BuiltIn::Match.new(expected)
+      if ::String === expected
+        BuiltIn::Captures.new(expected)
+      else
+        BuiltIn::Match.new(expected)
+      end
```
myronmarston: What about the inverse match? E.g.

expect("2015-01-01").to match(/(\d{4})-/).with_captures(...)

lib/rspec/matchers/built_in/captures.rb
```diff
@@ -0,0 +1,78 @@
+module RSpec
+  module Matchers
+    module BuiltIn
+      # @api private
+      # Provides the regex with captures matcher
+      class Captures < BaseMatcher
```
jasonkarns: I'm not usually one for inheritance, but is this a legit case for Captures to inherit from Match?

lib/rspec/matchers/built_in/captures.rb
```diff
@@ -0,0 +1,78 @@
...
+            end
+          else
+            false
+          end
+        end
+      end
+
+      # @api private
+      # Used to wrap match data and make it reliable for 1.8.7
+      class ReliableMatchData
```
jasonkarns: Suggestion: this looks like a prime use case for DelegateClass. `initialize` doesn't need defined, nor `captures` ?

JonRowe: We don't use the standard library if we can avoid it as it causes false positives for other developers.

lib/rspec/matchers.rb
JonRowe: Do you mean `to_not`?

lib/rspec/matchers/built_in/captures.rb
JonRowe: Actually I think that's a good improvement, it would eliminate the need to instantiate it at all, you could just call `super`

lib/rspec/matchers/built_in/captures.rb
JonRowe: :+1:

spec/rspec/matchers/built_in/captures_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it "matches a regex without named captures" do
+    expect(/(a)/).to match("a123a").with_captures("a")
+  end
+
+  if RUBY_VERSION != "1.8.7"
+    it "matches a regex with named captures" do
+      expect(Regexp.new("(?<num>123)")).to match("a123a").with_captures(:num => "123")
+    end
+    it "matches a regex with a nested matcher" do
```
JonRowe: Needs an extra line break

lib/rspec/matchers.rb
samphippen: no, this refers to the opposite case to the one now, where the regex is the target and the string is the value passed into the matcher.

myronmarston: Regardless, the point I was trying to make is that we should support with_captures for both regex matching against a strong and a string matching against a regex. All the specs I saw in this PR were only one way (the opposite of what I wrote here, I thought) and this logic looks like it only supports it one way.

Maybe it would be better to add a with_captures method to Match and have this always return a match instance? Besides, it seems odd to return the Captures instance if the user has not called with_captures and may not ever call that.

JonRowe: The other PR helped me understand that, should have said, sorry!

jasonkarns: @myronmarston adding with_captures to Match is the approach I took in #849. Two subclasses of Match are created (MatchWithCaptures, MatchWithNamedCaptures) and the appropriate one is instantiated/returned when with_captures is invoked. If it's never invoked, the matcher remains an instance of Match. The choice between MatchWithCaptures/MatchWithNamedCaptures is made based on whether the expected_captures are a Hash or hash-matcher.

samphippen:
@JonRowe @myronmarston I've merged the code into the match matcher and I'm pretty pleased with how this is all looking, what do you think? Shall I squash, changelog and merge on green?

jasonkarns:
I have a couple edge cases that I don't believe would be covered by this. Should I submit them as failing tests in a PR to this PR?

samphippen:
@jasonkarns can you just provide a gist with failing examples?

jasonkarns:
Sure, I'll get them up later today

jasonkarns:
gisted: https://gist.github.com/jasonkarns/f009fa582315fae87cfb

samphippen:
@myronmarston/@jonr do you have thoughts on this example:

```
it "matches indexed captures against named captures" do
  expect(Regexp.new("(?<num>123)")).to match("a123a").with_captures("123")
end
```

I could go either way with this passing.

myronmarston:
I'd expect it to pass.

JonRowe:
This LGTM, I'd merge once squashed.

PR #849, opened 2015-09-14
MatchWithCaptures

### LINKED ISSUE #831: Add capturing support to the `match` built-in matcher

jasonkarns:
My attempt at MatchWithCaptures to close #831

originally based on #848 with a couple added tests

The key differences from #848 are:

The logic for `MatchWithCaptures` is a refinement of behavior that *only takes effect when `with_captures` is invoked*. Otherwise, the default `Match` behavior is used. So rather than duplicating the behavior of `Match`, this PR inherits `Match` to `MatchWithCaptures` and `MatchWithNamedCaptures`. These subclasses are only instantiated when `with_captures` is invoked. (Thus, if `with_captures` is not invoked, the normal `Match` behavior applies without needing to be duplicated or manually invoked.)

- logic for creating a hash of named captures is pushed into the `ReliableMatchData` (it's only a concern of the matchdata object, after all. the alternative feels a bit to ask-don't-tell since we're mucking with the properties of matchdata rather than relying on it to give us a proper hash)

### DIFF
--- a/lib/rspec/matchers/built_in/match.rb
+++ b/lib/rspec/matchers/built_in/match.rb
@@ -5,19 +5,10 @@ module RSpec
       # Provides the implementation for `match`.
       # Not intended to be instantiated directly.
       class Match < BaseMatcher
-        def initialize(expected)
-          super(expected)
-
-          @expected_captures = nil
-        end
         # @api private
         # @return [String]
         def description
-          if @expected_captures && @expected.match(actual)
-            "match with string #{@expected} have captures #{surface_descriptions_in(@expected_captures).inspect}"
-          else
-            "match #{surface_descriptions_in(expected).inspect}"
-          end
+          "match #{surface_descriptions_in(expected).inspect}"
         end
 
         # @api private
@@ -29,81 +20,59 @@ module RSpec
         # Used to specify the captures we match against
         # @return [self]
         def with_captures(*captures)
-          @expected_captures = captures
-          self
+          expected_captures = ExpectedCaptures.new(captures.last)
+
+          if expected_captures.named?
+            MatchWithNamedCaptures.new @expected, expected_captures.stringify_keys
+          else
+            MatchWithCaptures.new @expected, captures
+          end
         end
 
       private
 
-        def captures_error
-          ArgumentError.new("`with_captures` must be called with a regular expression, multiple strings, or a hash of named captures")
-        end
-
         def match(expected, actual)
-          return match_captures(expected, actual) if @expected_captures
-          return true if values_match?(expected, actual)
-          return false unless can_safely_call_match?(expected, actual)
-          actual.match(expected)
+          if can_safely_call_match?(expected, actual)
+            actual.match(expected)
+          else
+            values_match?(expected, actual)
+          end
         end
 
         def can_safely_call_match?(expected, actual)
           return false unless actual.respond_to?(:match)
+          return false if Regexp === expected && Regexp === actual
 
           !(RSpec::Matchers.is_a_matcher?(expected) &&
             (String === actual || Regexp === actual))
         end
 
-        def match_captures(expected, actual)
-          match = actual.match(expected)
-          if match
-            match = ReliableMatchData.new(match)
-            if match.names.empty?
-              values_match?(@expected_captures, match.captures)
-            else
-              expected_matcher = @expected_captures.last
-              values_match?(expected_matcher, Hash[match.names.zip(match.captures)]) ||
-                values_match?(expected_matcher, Hash[match.names.map(&:to_sym).zip(match.captures)]) ||
-                values_match?(@expected_captures, match.captures)
-            end
-          else
-            false
+        class ExpectedCaptures
+          def initialize(captures)
+            @captures = captures
+          end
+
+          def named?
+            Hash === @captures || is_a_hash_matcher?
+          end
+
+          def stringify_keys
+            captures_hash.replace captures_hash.reduce({}) {|h, (k, v)| h[k.to_s] = v; h }
+            @captures
+          end
+
+          private
+
+          def is_a_hash_matcher?
+            RSpec::Matchers.is_a_matcher?(@captures) &&
+              @captures.expected.respond_to?(:first) &&
+              Hash === @captures.expected.first
+          end
+
+          def captures_hash
+            is_a_hash_matcher? ? @captures.expected.first : @captures
           end
         end
-      end
-
-      # @api private
-      # Used to wrap match data and make it reliable for 1.8.7
-      class ReliableMatchData
-        def initialize(match_data)
-          @match_data = match_data
-        end
-
-        if RUBY_VERSION == "1.8.7"
-          # @api private
-          # Returns match data names for named captures
-          # @return Array
-          def names
-            []
-          end
-        else
-          # @api private
-          # Returns match data names for named captures
-          # @return Array
-          def names
-            match_data.names
-          end
-        end
-
-        # @api private
-        # returns an array of captures from the match data
-        # @return Array
-        def captures
-          match_data.captures
-        end
-
-      protected
-
-        attr_reader :match_data
       end
     end
   end
--- a/lib/rspec/matchers/built_in/match_with_captures.rb
+++ b/lib/rspec/matchers/built_in/match_with_captures.rb
@@ -1,0 +1,78 @@
+module RSpec
+  module Matchers
+    module BuiltIn
+      # @api private
+      # Extends Match behavior to also match against expected captures
+      class MatchWithCaptures < Match
+        def initialize(expected, captures)
+          super(expected)
+          @expected_captures = captures
+        end
+
+        # @api private
+        # @return [String]
+        def description
+          "#{super} with captures #{surface_descriptions_in(@expected_captures).inspect}"
+        end
+
+      private
+
+        def match(expected, actual)
+          (match = super) && values_match?(@expected_captures, actual_captures(match))
+        end
+
+        def actual_captures(match_data)
+          ReliableMatchData.new(match_data).captures
+        end
+      end
+
+      # @api private
+      # Extends MatchWithCaptures behavior to match against *named* captures
+      class MatchWithNamedCaptures < MatchWithCaptures
+      private
+
+        def actual_captures(match_data)
+          ReliableMatchData.new(match_data).named_captures
+        end
+      end
+
+      # @api private
+      # Used to wrap match data and make it reliable for 1.8.7
+      class ReliableMatchData
+        def initialize(match_data)
+          @match_data = match_data
+        end
+
+        # @api private
+        # returns an array of captures from the match data
+        # @return Array
+        def captures
+          @match_data.captures
+        end
+
+        if RUBY_VERSION == "1.8.7"
+          # @api private
+          # Returns match data names for named captures
+          # @return Array
+          def names
+            []
+          end
+        else
+          # @api private
+          # Returns match data names for named captures
+          # @return Array
+          def names
+            @match_data.names
+          end
+        end
+
+        # @api private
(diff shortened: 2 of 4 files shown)


jasonkarns:
The key differences from #848 are:

- `MatchWithCaptures` inherits from `Match` as the with-captures behavior is a refinement of the `Match` matcher
- logic for creating a hash of named captures is pushed into the `ReliableMatchData` (it's only a concern of the matchdata object, after all. the alternative feels a bit to ask-don't-tell since we're mucking with the properties of matchdata rather than relying on it to give us a proper hash)
- logic for getting hash (named captures) vs array (simple indexed captures) is pushed into `ReliableMatchData`. (same rationale as above)
- check `responds_to? :name` instead of dealing with `NameError` for rubies that don't support named captures

Again: biggest blocker on this being complete is a clean way for `with_captures` to accept hash (named captures) and also be variadic (simple indexed captures) while still supporting nested matchers. https://github.com/rspec/rspec-expectations/pull/849/files#diff-3eee7938c499889ac0c5c1d6e629c689R19

jasonkarns:
annoyed with rubocop's blind hate of `and`. This is a legit use case: https://github.com/rspec/rspec-expectations/pull/849/files#diff-3eee7938c499889ac0c5c1d6e629c689R16

lib/rspec/matchers/built_in/match.rb
```diff
@@ -5,6 +5,22 @@ module BuiltIn
       # Provides the implementation for `match`.
       # Not intended to be instantiated directly.
       class Match < BaseMatcher
+        # Used to specify the captures we match against
+        # Accepts variadic non-named captures
+        # @return [self]
+        def with_captures(*captures)
+          @expected_captures = captures
+          extend WithCaptures
+        end
```
JonRowe: I have two problems with this, firstly dynamic `extend`? Yuk. Both from a stylistic pov and a performance pov. Second, this allows `with_captures` on strings, which is non ideal. I prefer @samphippen's dual class approach for this.

lib/rspec/matchers/built_in/match_with_captures.rb
```diff
@@ -0,0 +1,62 @@
+module RSpec
+  module Matchers
+    module BuiltIn
+      # @api private
+      # Extends Match behavior to also match against expected captures
+      module WithCaptures
+        # @api private
+        # @return [String]
+        def description
+          "#{super} with captures #{surface_descriptions_in(@expected_captures).inspect}"
```
JonRowe: This however is better for tackling the 'using bits of Match'

lib/rspec/matchers/built_in/match_with_captures.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        def actual_captures(match)
+          ReliableMatchData.new(match).captures
+        end
+
+        # @api private
+        # Used to wrap match data and make it reliable for 1.8.7
+        class ReliableMatchData
+          def initialize(match_data)
+            @match_data = match_data || NullMatchData.new
```
JonRowe: If we go down this route then `NullMatchData` should be a singleton.

JonRowe: Actually I prefer @samphippen's implementation here mainly, but like @myronmarston i'd rather conditionally define `names` on 1.8.7 to be `[]`

spec/rspec/matchers/built_in/match_with_captures_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  it "does not match a regex that doesn't match without captures" do
+    expect(/(a)/).not_to match("123").with_captures()
+  end
+
+  if RUBY_VERSION != "1.8.7"
+    it "matches a regex with named captures" do
+      expect(Regexp.new("(?<num>123)")).to match("a123a").with_named_captures(:num => "123")
+    end
+    it "matches a regex with a nested matcher" do
```
JonRowe: I know this is copy pasta from #848 but there should be an additional line break here.

lib/rspec/matchers/built_in/match.rb
jasonkarns: Don't we *want* it to apply to Strings as well? (I still need to add tests for that usage)
https://github.com/rspec/rspec-expectations/pull/848#discussion_r39408152

JonRowe: No, the expected should be regexp, strings don't have captures.

lib/rspec/matchers/built_in/match_with_captures.rb
jasonkarns: is this `if RUBY_VERSION.to_f == 1.8` the preferred method for 1.8 conditional method def? or < 1.9?

JonRowe: `if RUBY_VERSION == '1.8.7'` is sufficient, you can't install rspec on anything lower than 1.8.7 anyway

lib/rspec/matchers/built_in/match.rb
jasonkarns: The original use case (for which I opened #831) was precisely for testing a regexp. So the expected *is* a string. The actual is the regexp.

JonRowe: Sorry misspeak, I meant actual. (I often muddle them up because `expect(actual)`)

jasonkarns: Cool. Though this fits with my use case (testing a regexp), I'm still wondering if it makes sense to have `with_captures` available on a general `Match`.  The actual could be a string that is expected to match a regex.

Per @myronmarston's example:

```
expect("2015-01-01").to match(/(\d{4})-/).with_captures("2015")
```

JonRowe: Ahh, I see, also explains @myronmarston's comment, well you still need to prevent:

`expect("2015-01-01").to match("2015-01-01").with_captures`

jasonkarns: Suggested error to raise for that condition? Any prior art in rspec for mis-use of a matcher?

Since we can accept a string either as expected or actual, we won't know at matcher-instantiation-time if it's a legal method call. So the earliest moment that we will know it's invalid is invocation of `.match(expected, actual)`, at which point we can raise an exception if both expected and actual are strings.

JonRowe: I'd return false, as one of the values should have been a regexp when using `with_captures` so they don't "match", and in the failure message say that expected with supposed to be a regular expression when actual is a string.

jasonkarns: Hmm, not sure I agree. Returning false is akin to saying "the predicate tested is valid but does not match". Whereas doing with_captures on a string/string assertion isn't even a valid predicate.

Put another way, `match` returning false with a message is generally intended to be fixed by changing the code under test. However, in this case, the test itself is flawed so I would expect an exception to be raised to indicate the test itself is invalid.

JonRowe: > Whereas doing with_captures on a string/string assertion isn't even a valid predicate.
Yeah but at 'match' time we only know that expected should have been a regular expression, consider the case where a string is accidentally returned instead of a regular expression, it would make sense to use `with_captures` but have it fail because it doesn't match....

jasonkarns: That example *might* sway me. However, it would require the code to be horribly mangled; the description would need to know not just that the match failed, but that it failed for this weird exceptional case. Which would require adding a bunch of additional state to the Match class in order to carry this knowledge from `match` to `description`. no?

jasonkarns:
I'm pretty happy the the structure of the PR. However, the last bit that I'm struggling to overcome is knowing when to check the named captures vs indexed captures. #848 always checks named captures if `.names` exists and isn't empty on the match. However, that approach would fail this test:

``` ruby
expect(/(?<num>123)/).to match("a123a").with_captures("123")
```

IMO, the *expected* captures should be the determining factor in comparing named vs indexed captures. And this information is available as soon as the expectation is created. This way, the appropriate matcher can be instantiated, since there's no need to wait for the *actual* value before deciding.

Thus, the argument(s) to `with_captures` are all that should be necessary for this determination. If `with_captures` is given a hash, it's clearly a named-captures comparison. However, if given another matcher, it's ambiguous.

``` ruby
# matcher but intending to compare named captures
expect(/(?<num>123)(asdf)?/).to match("a123a").with_captures(hash_including(:num => "123"))

# matcher but intending to compare unnamed/indexed captures
expect(/(123)(asdf)?/).to match("a123a").with_captures(a_string_including("123"))
```

Is there a way to distinguish the type of a matcher? Specifically, we only want to compare named captures if the argument to `with_captures` is a literal `Hash` or a Hash matcher (`Include` matcher I believe).

myronmarston:
One option is to err on the side of permissiveness and try matching both ways (named and index), and, if it matches either way, consider it a match.  Unless there's some fool-proof way to know for sure what the user intends I don't know if there's any other option.

jasonkarns:
We can already tell if the expected captures is a Hash (since the `value`
of a capture can't be a hash). The only one left is if we can distinguish a
Hash matcher from other types of matchers. If that's possible, then we
would know for sure they are checking named captures.

jasonkarns:
I'm stuck in a situation where rubocop has two incompatible rules enabled:

- Extra empty line detected at body beginning.
- Keep a blank line before and after private.

In a class with only private methods, these two rules are directly conflicting and fixing one violates the other. What do?

jasonkarns:
This PR should be good for review. The current failing job is unrelated to the PR.

samphippen:
@jasonkarns are you happy that my merge fixes everything you wanted to see here? If so, I'll close this, otherwise can you rebase and add any further changes so we can discuss them?

jasonkarns:
@samphippen thanks. i submitted two separate PRs, each minor. and I'm going to rebase this as a separate pr. One question I have is how we should handle string/string match.

@JonRowe [mentioned](https://github.com/rspec/rspec-expectations/pull/849#discussion_r39466439) that we ought to prevent them. With #848 merged, a string/string match with captures will work by converting the `expected` string into a regex.

```
it "does not match strings with other strings with captures" do
  expect("a123a").not_to match("(a)").with_captures("a")
end
```

I don't really have a strong opinion either way, but it seemed that @JonRowe wanted to prevent this. Personally, I think matching (with captures) two strings ought to work. Since `match

PR #851, opened 2015-09-18
MatchWithCaptures cleanup

jasonkarns:
needed newlines
removed unused method
consistent use of a_hash_including matcher


samphippen:
:+1:

PR #852, opened 2015-09-18
reword MatchWithCaptures description

jasonkarns:
The description for failed captures should read just like the normal 'match' description, but with the addition of the captures clause.

We shouldn't specify that either expected or actual is a string (since either could be a string).
We shouldn't use rely on to_s but rather inspec, that way the expected is printed in a consistent fashion for both the captures and non-captures cases.


samphippen:
:+1:

samphippen:
@JonRowe merge?

JonRowe:
:+1:

PR #857, opened 2015-10-07
Fix SystemStackError raised when diffing a recursive Enumerable object

yujinakayama:
This patch fixes SystemStackError raised when diffing an Enumerable object whose `#each` includes the object itself:

```ruby
array = ['something']
array << array
expect(array).to eq([])
```

```
     Failure/Error:
       expect(array).to eq([])
     SystemStackError:
       stack level too deep
```


myronmarston:
LGTM

yujinakayama:
#856 needs to be merged to fix the build.

PR #858, opened 2015-10-07
Fix failing test due to extra spacing on lines 4 and 5

zsyed91:


myronmarston:
Thanks for taking a crack at this but it already got fixed in #855 yesterday.

PR #859, opened 2015-10-09
Always print compound failure messages multi-line.

### LINKED ISSUE #747: Consider using multi-line form for all compound matcher failure messages
Given this expectation:

``` ruby
      expect(last_cmd_stdout).to include(
        "2 examples, 0 failures",
        "is selected by line", "is selected by tag"
      ).and exclude("not selected")
```

I'm getting a failure message like so:

```
       expected "Run options: include {:tag => true, :locations => {\"./tmp/aruba/spec/file_1_spec.rb\" => [2]}}\n\nFile 1\n  is selected by line\n\nFile 2\n  is not selected\n  is selected by tag\n\nFinished in 0.00059 seconds (files took 0.12448 seconds to load)\n3 examples, 0 failures\n" to include "2 examples, 0 failures", "is selected by line", and "is selected by tag" and expected "Run options: include {:tag => true, :locations => {\"./tmp/aruba/spec/file_1_spec.rb\" => [2]}}\n\nFile 1\n  is selected by line\n\nFile 2\n  is not selected\n  is selected by tag\n\nFinished in 0.00059 seconds (files took 0.12448 seconds to load)\n3 examples, 0 failures\n" not to include "not selected"
```

There's also a diff which helps (although that has problems of its own -- see #746) but the message itself is pretty hard to read.  I think that we should consider making compound (`and`/`or`) always use the multi-line form (where they use `...and: ` and indentation) because that will be more readable.  Or perhaps we should still use the single-line form if the message is below some length threshold?  For short messages it still reads OK.

Thoughts?

myronmarston:
They are harder to read when put on a single line.

Closes #747.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,7 +12,8 @@ Enhancements:
   [match matcher](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/match-matcher)
   which allows a user to specify expected captures when matching a regex
   against a string. (Sam Phippen, #848)
-
+* Always print compound failure messages in the multi-line form. Trying
+  to print it all on a single line didn't read very well. (Myron Marston, #859)
 
 Bug Fixes:
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,6 @@ install:
   - gem install bundler
   - bundler --version
   - bundle install
-  - cinst ansicon
 
 test_script:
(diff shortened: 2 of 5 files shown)


myronmarston:
Anyone from @rspec/rspec want to review this?

soulcutter:
Definitely an improvement

PR #860, opened 2015-10-10
Remove parallel assignment and break up string that was too long. 

zsyed91:
Made some minor refactors as stated below. 

* Removed parallel assignment in initialize. IMO makes it easier to read as well
* Break up string from one to two lines as length was very long
* Only 2 rubocop issues remain `matches?` and `warn_for_false_positives`


JonRowe:
Thanks!

Issue #822, opened 2015-07-18
raise_error warning should include current error message


xaviershay:
In upgrade some applications to latest RSpec, I've noticed a common pattern:

* Get a warning about using `raise_error` without an argument.
* Comment out `raise_error` expectation and re-run test to see what an appropriate argument would be from the failure.
* Add argument, re-run test to see it green.

Looking at `BuiltIn::RaiseError#matches?` seems like it could be pretty straight-forward to include the exception message (if raised) in the warning.

xaviershay:
I'll probably tackle this when I get a moment, just leaving a reminder here since I'm not going to get to it tonight.

myronmarston:
I think this is a good idea, but there are a couple potential issues to bear in mind:

* The error currently being raised may not be the one the user actually cares about.  For example, if the false positive issue the warning warns about has occurred (e.g. the method being tested was renamed without the call site in the `raise_error` block being updated) than the error isn't the one that user cares about.
* Writing a non-brittle test requires judging which parts of the exception identify the exception while being unlikely to change.  Simply putting the entire current exception message in the expectation is a bad idea as the spec will fail at a simple punctuation, capitolization, whitespace or spelling change.

So I think it's great for the warning to give the user more detail about the exception currently being raised as long as it doesn't direct them to encode the full details of the current exception into the expectation.

xaviershay:
> So I think it's great for the warning to give the user more detail about the exception currently being raised as long as it doesn't direct them to encode the full details of the current exception into the expectation.

Totally agree.

I actually discovered a `NoMethodError` this way, it just took me more messing around than seemed necessary.

mrageh:
@xaviershay I was thinking of attempting to fix this issue, but so far I don't fully understand what the issue is. Can you please provide a test that describes the problem in more detail?

xaviershay:
``` ruby
describe do
  it 'works' do
    expect {
      raise "it failed"
    }.to raise_error
  end
end
```

Gives this output
```
> rspec test_spec.rb
[...]

PR #862, opened 2015-10-14
Include current error message in raise_error warning

mrageh:
Fix for issue #822

https://github.com/rspec/rspec-expectations/issues/822

When the raise_error() matcher triggers a warning it does not include
the cause of the actual error. This commit changes the warning message
to include the actual error that was raised.

I've moved setting the warning message `RaiseError#warn_about_bare_error`
further down in `RaiseError##matches` because the warning message needs
an instance variable to be set in order to display the current error.

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -46,7 +46,6 @@ module RSpec
           @eval_block = false
           @eval_block_passed = false
 
-          warn_about_bare_error if warning_about_bare_error && !negative_expectation
           return false unless Proc === given_proc
 
           begin
@@ -59,6 +58,7 @@ module RSpec
             end
           end
 
+          warn_about_bare_error if warning_about_bare_error && !negative_expectation
           eval_block if !negative_expectation && ready_to_eval_block?
 
           expectation_matched?
@@ -160,6 +160,7 @@ module RSpec
                         "will match when Ruby raises a `NoMethodError`, `NameError` or " \
                         "`ArgumentError`, potentially allowing the expectation to pass " \
                         "without even executing the method you are intending to call. " \
+                        "Actual error raised was #{description_of(@actual_error)}. "\
                         "Instead consider providing a specific error class or message. " \
                         "This message can be supressed by setting: " \
                         "`RSpec::Expectations.configuration.warn_about_potential_false_positives = false`")
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -160,7 +160,7 @@ module RSpec
                         "will match when Ruby raises a `NoMethodError`, `NameError` or " \
                         "`ArgumentError`, potentially allowing the expectation to pass " \
                         "without even executing the method you are intending to call. " \
-                        "Actual error raised was #{description_of(@actual_error)}. "\
+                        "#{warning}"\
                         "Instead consider providing a specific error class or message. " \
                         "This message can be supressed by setting: " \
                         "`RSpec::Expectations.configuration.warn_about_potential_false_positives = false`")
@@ -214,6 +214,11 @@ module RSpec
         def raise_message_already_set
           raise "`expect { }.to raise_error(message).with_message(message)` is not valid. " \
                 'The matcher only allows the expected message to be specified once'
+        end
+
+        def warning
+          warning = "Actual error raised was #{description_of(@actual_error)}. "
+          warning if @actual_error
         end
       end
       # rubocop:enable RescueException
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -18,6 +18,17 @@ RSpec.describe "expect { ... }.to raise_error" do
     expect { raise StandardError.new, 'boom'}.to raise_error
   end
 
+  it 'issues a warning that does not include current error when its not present' do
+    expect(::Kernel).to receive(:warn) do |message|
+      ex = /Actual error raised was/
+      expect(message).not_to match ex
+    end
+
+    expect {
+      expect{ '' }.to(raise_error)
+    }.to fail_with("expected Exception but nothing was raised")
+  end
+
   it "can supresses the warning when configured to do so", :warn_about_potential_false_positives do
     RSpec::Expectations.configuration.warn_about_potential_false_positives = false
     expect_no_warnings
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -13,6 +13,11 @@ RSpec.describe "expect { ... }.to raise_error" do
     expect { raise }.to raise_error
   end
 
+  it 'issues a warning that includes the current error when used without an error class or message' do
+    expect_warning_with_call_site __FILE__, __LINE__+1, /Actual error raised was #<StandardError: boom>/
+    expect { raise StandardError.new, 'boom'}.to raise_error
+  end
+
   it "can supresses the warning when configured to do so", :warn_about_potential_false_positives do
     RSpec::Expectations.configuration.warn_about_potential_false_positives = false
     expect_no_warnings


mrageh:
When I was working on the `RaiseError` matcher I noticed something strange that may need refactoring. We set an instance variable [here] (https://github.com/mrageh/rspec-expectations/blob/include-error-message-in-raise-error-warning/lib/rspec/matchers/built_in/raise_error.rb#L15) and yet we use a method with the same name as the instance variable [here] (https://github.com/mrageh/rspec-expectations/blob/include-error-message-in-raise-error-warning/lib/rspec/matchers/built_in/raise_error.rb#L157). However that method is completely unrelated to the instance variable, Does this seem confusing to anyone else? Does it need to be refactored and should I do that in this PR?

mrageh:
cc @xaviershay

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -46,7 +46,6 @@ def matches?(given_proc, negative_expectation=false, &block)
           @eval_block = false
           @eval_block_passed = false
 
-          warn_about_bare_error if warning_about_bare_error && !negative_expectation
           return false unless Proc === given_proc
```
xaviershay: I'm surprised that moving this after the `return false` didn't cause anything to fail.

That would be this case I think:

```
> rspec test_spec.rb
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /private/tmp/test_spec.rb:3:in `block (2 levels) in <top (required)>'.
F

Failures:

  1) blah should raise Exception
     Failure/Error: expect(nil).to raise_error
       expected Exception but was not given a block
     # ./test_spec.rb:3:in `block (2 levels) in <top (required)>'

Finished in 0.0212 seconds (files took 0.21044 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./test_spec.rb:2 # blah should raise Exception
```

I'm ok with not issuing a warning here, and don't consider warnings part of backwards compatibility guarantee.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -160,6 +160,7 @@ def warn_about_bare_error
                         "will match when Ruby raises a `NoMethodError`, `NameError` or " \
                         "`ArgumentError`, potentially allowing the expectation to pass " \
                         "without even executing the method you are intending to call. " \
+                        "Actual error raised was #{description_of(@actual_error)}. "\
```
xaviershay: is there a test already covering when `@actual_error` is `nil`? (If the spec passes?) Intuitively seems like the output would be wrong with current code.

xaviershay:
You're right instance var/method thing is pretty weird. Naming is hard :( I don't feel too strongly that we need to change it, but if you have better names feel free to add it to this PR.

xaviershay:
Thank you for having a go at this! The only comment we need to address before merge is the one about `nil` `@actual_error`.

lib/rspec/matchers/built_in/raise_error.rb
mrageh: Yeah I was surprised that the build did'nt fail after moving that line of code, but I thought that's the cleanest/simplest way of capturing what the actual_error was.

Maybe I could move that line back and then update the warning message when `@actual_error` is not `nil`?

lib/rspec/matchers/built_in/raise_error.rb
mrageh: Yeah you're right there is no test to cover this case, I'll add one now. 

I think I can improve my work here by setting the warning at the beginning and then modifying the warning message when `@actual_error` is present.

mrageh:
Thank you for reviewing my work @xaviershay

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: I'm surprised we return `false` here actually, we should be raising an `ArgumentError`

myronmarston: I disagree -- with composable matchers in RSpec 3 a matcher needs to be able to be asked if it matches against _any_ type of object, and it should signal it does or does not by returning true/false from `matches?`.  We should only raise errors from `matches?` when the matcher itself is invalid and cannot be matched against any object (such as when the user has done `be_within(delta)` without chaining `of(value)`).

JonRowe: We raise an error when someone attempts to use a block matcher without a block, I think the inverse should hold true too...

myronmarston: The user is properly notified of their mistake.  See 79582c2c.  Changing to raise an ArgumentError would be a breaking change.

lib/rspec/matchers/built_in/raise_error.rb
mrageh: @xaviershay and @myronmarston I tried to write a test for when `@actual_error` is `nil` and I may be doing something wrong, can I please get some pointers on how to rewrite this test?

```
it 'issues a warning that does not include current error when its not present' do
  expect(::Kernel).to receive(:warn) do |message|
    ex = /Actual error raised was/
    expect(message).not_to match ex
  end
  expect{ '' }.to(raise_error())
 end
```

xaviershay: that works for me....


```
  1) expect { ... }.to raise_error issues a warning that does not include current error when its not present
     Failure/Error: RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
       expected "WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was nil. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/xavier/Code/ex/rspec-expectations-isolate/spec/rspec/matchers/built_in/raise_error_spec.rb:26:in `block (2 levels) in <top (required)>'." not to match /Actual error raised was/
       Diff:
       @@ -1,2 +1,2 @@
       -/Actual error raised was/
       +"WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was nil. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/xavier/Code/ex/rspec-expectations-isolate/spec/rspec/matchers/built_in/raise_error_spec.rb:26:in `block (2 levels) in <top (required)>'."

     # /Users/xavier/.gem/ruby/2.2.2/bundler/gems/rspec-support-81e33d4ed3d9/lib/rspec/support.rb:87:in `block in <module:Support>'
```

mrageh: @xaviershay initially it does work but when you implement your solution and make `expect(message).not_to match ex` pass, the second test fails. The reason why it fails is because it expects an error to be raised;
```
1) expect { ... }.to raise_error issues a warning that does not include current error when its not present
     Failure/Error: RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
       expected Exception but nothing was raised
   ```

Comment (xaviershay) [inline lib/rspec/matchers/built_in/raise_error.rb]: oh right. You can do something like this:

```
  it "fails if nothing is raised" do
    expect {
      expect { }.to raise_error Exception
    }.to fail_with("expected Exception but nothing was raised")
  end
```

mrageh: Thank you!!!!

mrageh:
@xaviershay this PR may be ready to merge, can you take one last look at it

xaviershay:
Thank you!

PR #864, opened 2015-10-16
Minor refactors to line lengths

zsyed91:
* Shortened failure_message method by moving the failure message string out
* Moved the longer string lines into separate named methods
* Removed some duplication in the string interpolation into a separate method
* Fixed minor rubocop issue with spacing in assignment

With the changes, these lines are now <= 80 characters in length. And IMO slightly easier to read+maintain since we don't have to space pad the strings.


myronmarston:
Thanks!

PR #865, opened 2015-10-16
Minor refactors to line lengths and rubocop issues

zsyed91:
* Shortened strings spanning more than 80 characters
* Replaced double quotes to single quotes inside string interpolation per rubocop
* Replaced `gsub` with `tr` in string interpolation for the performance gain of 1-1 replacements

### DIFF
--- a/benchmarks/gsub_vs_tr_single_character.rb
+++ b/benchmarks/gsub_vs_tr_single_character.rb
@@ -1,0 +1,28 @@
+require 'benchmark/ips'
+
+Benchmark.ips do |x|
+  y = '1_2_3_4_5_6_7_8_9_10'
+
+  x.report('gsub') do |_times|
+    y.gsub('_', ' ')
+  end
+
+  x.report('tr') do |_times|
+    y.tr('_', ' ')
+  end
+
+  x.compare!
+end
+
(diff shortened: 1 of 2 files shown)


myronmarston:
Thanks!  The changes look good.  The one thing I'd ask is for a benchmark showing that `tr` is faster than `gsub` for these.  We like to commit benchmarks to the repo with any perf-related changes.  Do you mind adding a benchmark to the `benchmarks` directory comparing `tr` and `gsub`?  Our preferred benchmarking tool these days is benchmark-ips.  Here's a semi-recent example from rspec-core showing an example of usage:

https://github.com/rspec/rspec-core/blob/e37b7b88fb72a2f2f0a0eabf07ffee761370491d/benchmarks/shuffle_vs_sort_by_for_random_ordering.rb

zsyed91:
Hey thanks for reviewing at this @myronmarston. I will go ahead and submit a benchmark as well.

zsyed91:
I went ahead and added the benchmark comparing `gsub` to `tr`. I hope the code looks right, this was my first crack at using the benchmark-ips library.

myronmarston:
Thanks!

PR #866, opened 2015-10-21
Fix various rubocop issues and simplify yield

zsyed91:
* Convert non-interpolated strings to single quotes
* Simplified description method
* Swapped out + with \ for multiline strings
* Shortened some line lengths
* Remove parallel assignment


JonRowe:
Thanks

PR #867, opened 2015-10-24
Add relish documentation for block_arg.

xaviershay:
Previously there was no way to discover this.

@rspec/rspec please merge if green and you have no feedback!


xaviershay:
note: copied example from spec.

PR #868, opened 2015-10-24
Fix rubocop issues in base_matcher

zsyed91:
Fixed some rubocop issues in this base_matcher.
* Prefer single quotes over double when there are no interpolations
* Use shorthand `&:method` instead of `|var| var.method` notation
* rubocop spacing for multi-line `if` clause


JonRowe:
Thanks

PR #870, opened 2015-11-12
Use a relish URL that is not version-specific.

myronmarston:


JonRowe:
Is the build broken in general?

myronmarston:
> Is the build broken in general?

Yes.  I just pushed a fix.

PR #872, opened 2015-11-17
Fix rubocop issues in built_in matcher be.rb

zsyed91:
Minor refactoring to fix rubocop issues:
- Removed parallel assignment
- Changes double quotes to single quotes where interpolation is not needed
- Shortened some longer lines

### DIFF
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -137,7 +137,8 @@ module RSpec
         include BeHelpers
 
         def initialize(operand, operator)
-          @expected, @operator = operand, operator
+          @expected = operand
+          @operator = operator
           @args = []
         end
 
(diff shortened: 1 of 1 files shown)


lib/rspec/matchers/built_in/be.rb
```diff
@@ -75,7 +75,7 @@ module BeHelpers
       private
 
         def args_to_s
-          @args.empty? ? "" : parenthesize(inspected_args.join(', '))
+          @args.empty? ? '' : parenthesize(inspected_args.join(', '))
```
JonRowe: Our own rubocop rules don't enforce this, so in future it'd be good to avoid the git churn.

zsyed91: Noted, won't commit these again!

myronmarston: If you run `bundle exec rubocop lib` you can see what our rubocop rules enforce.

lib/rspec/matchers/built_in/be.rb
```diff
@@ -151,13 +152,15 @@ def matches?(actual)
         # @api private
         # @return [String]
         def failure_message
-          "expected: #{@operator} #{expected_formatted}\n     got: #{@operator.to_s.gsub(/./, ' ')} #{actual_formatted}"
+          "expected: #{@operator} #{expected_formatted}\n     " \
+          "got: #{@operator.to_s.gsub(/./, ' ')} #{actual_formatted}"
```
myronmarston: This string is designed so that the expected and actual ("got") value align, so it'd be nice to format the source string that way, too:

``` ruby
"expected: #{@operator} #{expected_formatted}\n" \
"     got: #{@operator.to_s.gsub(/./, ' ')} #{actual_formatted}"
```

JonRowe: :+1:

zsyed91: Ah good point, I will make this change.

lib/rspec/matchers/built_in/be.rb
zsyed91: Ah that make sense, I was simply running `rubocop lib`. Thank you for pointing that out.

zsyed91:
Made the changes you guys suggested.

myronmarston:
Thanks!  Do you mind squashing into one commit before we merge?

zsyed91:
I knew I forgot to do something :)

JonRowe:
Thanks :)

PR #873, opened 2015-11-17
[wip] ensure prerequisites for nokogiri are installed

JonRowe:


myronmarston:
Why do we need nokogiri for rspec-expectations? I thought we only needed it for rspec-rails.

JonRowe:
@myronmarston I think it's coming from the `rspec-core` build in this instance (which will use it for validating the html formatter) but it's failing the build here, my goal was if this worked would be just to add the prerequirests (`libxml2`) to all the travis builds...

myronmarston:
Thanks for explaining.  I hadn't seen the build failures yet.

It looks like Nokogiri 1.6.6.3 was released today; given this is a new problem, my guess is that it's only a problem on 1.6.6.3.  A simple solution might be to lock nokogiri in rspec-core's gemfile to `< 1.6.6.3`.

JonRowe:
Fixed by pinning nokogiri instead, FWIW it's apparently an issue with docker (which travis uses under the hood IIRC) and nokogiri, hopefully they'll fix it :/

PR #882, opened 2016-01-01
Updates from rspec-dev (2016-01-01)

yujinakayama:
These are some updates, generated from rspec-dev's rake tasks.

### DIFF
--- a/.rubocop_rspec_base.yml
+++ b/.rubocop_rspec_base.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2015-10-13T18:52:08-07:00 from the rspec-dev repo.
+# This file was generated on 2016-01-01T10:42:21+09:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # This file contains defaults for RSpec projects. Individual projects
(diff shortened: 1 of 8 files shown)


yujinakayama:
No idea why the installation of nokogiri 1.6.6.2 is failing, even though it's passed in rspec-core's build:

* https://travis-ci.org/rspec/rspec-expectations/builds/99672410
* https://travis-ci.org/rspec/rspec-core/builds/99692949

myronmarston:
> No idea why the installation of nokogiri 1.6.6.2 is failing, even though it's passed in rspec-core's build:

It's probably some inconsistency on the travis VMs.  It's happening for the rspec-mocks build, too:

https://travis-ci.org/rspec/rspec-mocks/jobs/99672433

Given all the recent problems we've had with nokogiri, I think it's worth considering alternate strategies.  A couple ideas:

* Instead of specifying versions of nokogiri not to be installed, we could specify an exact known good, known installable version of nokogiri.  We probably wouldn't have to update it very often.
* It's worth considering if we really need nokogiri.  I forget what we are using it for, honestly.  It's a dev dependency, though, so we should be able to swap it out with another XML parser.

Thoughts?

yujinakayama:
> It's worth considering if we really need nokogiri. I forget what we are using it for, honestly.

In repos except rspec-rails, only rspec-core is using nokogiri to filter backtrace in HtmlFormatter spec. I just opened https://github.com/rspec/rspec-core/pull/2149 for removing nokogiri dep in rspec-core.

For rspec-rails, nokogiri is a sub-dep of direct dependencies (like capybara) so it cannot be removed. Maybe we should take the first strategy in rspec-rails:

> Instead of specifying versions of nokogiri not to be installed, we could specify an exact known good, known installable version of nokogiri. We probably wouldn't have to update it very often.

yujinakayama:
Closing in favor of #884.

PR #883, opened 2016-01-04
Remove "literally" from `lib/rspec/matchers/built_in/raise_error.rb`

adelevie:


myronmarston:
The use of "literally" in this warning was intentional. Why do you find this to be an improvement?

adelevie:
It reads much better (to me, at least) without "literally". No worries if you disagree.

JonRowe:
I'm with @myronmarston on this, but thanks for the suggestion and your time :)

PR #891, opened 2016-01-15
Demonstrate a bug with the output matcher.

samphippen:
An expectation of the form

```ruby
expect { ... }.to output.to_stderr
```

will work if anything is written to `$stderr` during the execution of the
block, but not if anything is written to `STDERR`. I encountered this
during working on a project and found the behaviour to be confusing.

Ideally both would work.

### DIFF
--- a/spec/rspec/matchers/built_in/output_spec.rb
+++ b/spec/rspec/matchers/built_in/output_spec.rb
@@ -144,6 +144,13 @@ module RSpec
           $stderr.print(msg)
         end
       }
+
+      include_examples "output_to_stream", :stderr, :to_stderr, Module.new {
+        def print_to_stream(msg)
+          STDERR.print(msg)
+        end
+      }
+
     end
 
     RSpec.describe "output.to_stdout matcher" do


samphippen:
@rspec/rspec is this actually a bug?

myronmarston:
It's surprising behavior, but it's explicitly called out in our docs

http://rspec.info/documentation/3.4/rspec-expectations/RSpec/Matchers.html#output-instance_method
https://relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers/output-matcher

soulcutter:
One reason I always use `$stderr` and its ilk is because they can be substituted easily.

myronmarston:
> One reason I always use $stderr and its ilk is because they can be substituted easily.

Yep.  We could potentially fix `STDOUT`/`STDERR` by using constant stubbing on them...but I'm against that for a few reasons:

* That's in rspec-mocks and doesn't really belong here.
* Even if we did use it, constant stubbing is scoped to an example not to one expectation.
* While we could fix `STDOUT/STDERR` there are stil similar bugs that remain.  For example, if someone does `MY_OUTPUT_STREAM = $stdout` or `MY_OUTPUT_STREAM = STDOUT` at the top of their class, and then used that constant to print in their logic...the matcher is not going to work, and I know of no way to make the matcher work for this case.  The docs also mention this case.

Overall, I think the best we can do is document the gotchas, which we've done.

Issue #890, opened 2016-01-14
Block matchers swallow nested expectations' failures


severinraez:
Hi

As @trotzig and me found out in https://github.com/brigade/db-query-matchers/issues/11

    RSpec::Matchers.define :foo do
      match do |actual|
        actual.call
        true
      end
    
      def supports_block_expectations?
        true # or some logic
      end
    
      def failure_message
        'I failed'
      end
    end
    
    describe '#reraise' do
      context 'when some other expectation in the block fails' do
        subject {
          raise RSpec::Expectations::ExpectationNotMetError.new('other')
        }
    
        it 'reraises the error' do
          expect do
            expect { subject }.to foo
          end.to raise_error(RSpec::Expectations::ExpectationNotMetError, /other/)
        end
      end
    end

Fails with

    F
    
    Failures:
    
      1) #reraise when some other expectation in the block fails reraises the error
         Failure/Error:
           expect do
             expect { subject }.to foo
           end.to raise_error(RSpec::Expectations::ExpectationNotMetError, /other/)
    
           expected RSpec::Expectations::ExpectationNotMetError with message matching /other/, got #<RSpec::Expectations::Expecta
    tionNotMetError: I failed> with backtrace:
             # ./spec/db_query_matchers/make_database_queries_spec.rb:26:in `block (4 levels) in <top (required)>'
             # ./spec/db_query_matchers/make_database_queries_spec.rb:25:in `block (3 levels) in <top (required)>'
         # ./spec/db_query_matchers/make_database_queries_spec.rb:25:in `block (3 levels) in <top (required)>'
    
    Finished in 0.01832 seconds (files took 1.15 seconds to load)
    1 example, 1 failure

This increases the difficulty of finding the cause of a failing spec that uses nested expectations as the spec will provide an unusable hint in the failure message.

Is this by design or would you consider it a bug? Any suggestions on how one could contribute a fix?

Cheers,
Severin

myronmarston:
This is (kind of) working as designed.  When you use the matcher DSL, we support you defining your match logic by delegating to another expectation:

``` ruby
RSpec::Matchers.define :be_positive do
  match do |value|
    expect(value).to be > 0
  end
end
```

(In this case, the `match` logic could be simplified to `value > 0` but you can imagine cases where it's easier to express the logic using an expectation than a boolean expression).

[...]

PR #892, opened 2016-01-18
Stop block matchers from swallowing nested expectations

JonRowe:
This is an attempt at a solution for #890 , by wrapping actual "value" blocks in another closure, we should be able to captures any expectation errors internally before re-raising them, which in turn should allow us to maintain existing block behaviour but also raise nested expectations correctly from the matcher.

Thoughts? @myronmarston

### DIFF
--- a/features/custom_matchers/define_block_matcher.feature
+++ b/features/custom_matchers/define_block_matcher.feature
@@ -42,3 +42,37 @@ Feature: define a matcher supporting block expectations
       """
     When I run `rspec ./block_matcher_spec.rb`
     Then the example should pass
+
+  Scenario: define a block matcher using shortcut
+    Given a file named "block_matcher_spec.rb" with:
+      """ruby
+      RSpec::Matchers.define :support_blocks_with_errors do
+        match(:notify_expectation_failures => true) do |actual|
+          actual.call
+          true
+        end
+
+        supports_block_expectations
+      end
+
+      RSpec.describe "a custom block matcher" do
+        specify do
+          expect {
+            expect(true).to eq false
+          }.to support_blocks_with_errors
+        end
+      end
+      """
+    When I run `rspec ./block_matcher_spec.rb`
+    Then it should fail with:
+      """
+      Failures:
+
+        1) a custom block matcher should support blocks with errors
+           Failure/Error: expect(true).to eq false
+
+             expected: false
+                  got: true
+
+             (compared using ==)
+      """
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -55,16 +55,23 @@ module RSpec
         #     expect(3).to be_even     # fails
         #     expect(4).not_to be_even # fails
         #
+        # By default the match block will swallow expectation errors (e.g.
+        # caused by using an expectation such as `expect(1).to eq 2`), if you
+        # with to allow these to bubble up, pass in the option
+        # `:notify_expectation_failures => true`.
+        #
+        # @param [Hash] options for defining the behavior of the match block.
         # @yield [Object] actual the actual value (i.e. the value wrapped by `expect`)
-        def match(&match_block)
+        def match(options={}, &match_block)
           define_user_override(:matches?, match_block) do |actual|
-            begin
-              @actual = actual
-              RSpec::Support.with_failure_notifier(RAISE_NOTIFIER) do
+            @actual = actual
+            RSpec::Support.with_failure_notifier(RAISE_NOTIFIER) do
+              begin
                 super(*actual_arg_for(match_block))
+              rescue RSpec::Expectations::ExpectationNotMetError
+                raise if options[:notify_expectation_failures]
+                false
               end
-            rescue RSpec::Expectations::ExpectationNotMetError
-              false
             end
           end
         end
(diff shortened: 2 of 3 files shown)


lib/rspec/matchers/dsl.rb
```diff
@@ -358,6 +373,9 @@ class Matcher
         # object wrapped by `expect`.
         attr_reader :actual
 
+        # @private
+        attr_accessor :actual_error
```
myronmarston: Is it necessary to expose `actual_error` as an instance variable and attribute method?  Can't it just be a local?

myronmarston:
It'd be nice to not have this complexity, but the idea is sound.  That said, it's a breaking change for a case like:

``` ruby
RSpec::Matchers.define :custom_be do |expected|
  match { |actual| expected.equal?(actual) }
end

it 'does not break block identity checks' do
  block = lambda { }
  expect(block).to custom_be(block)
end
```

This passes in `master` and fails on this branch:

```
  1) RSpec::Matchers::DSL::Matcher does not break block identity checks
     Failure/Error: expect(block).to custom_be(block)
       expected #<Proc:0x007fc051b953e8@/Users/myron/code/rspec-dev/repos/rspec-expectations/lib/rspec/matchers/dsl.rb:64> to custom be #<Proc:0x007fc051b95708@/Users/myron/code/rspec-dev/repos/rspec-expectations/spec/rspec/matchers/dsl_spec.rb:748 (lambda)>
     # ./spec/rspec/matchers/dsl_spec.rb:749:in `block (2 levels) in <module:DSL>'
```

myronmarston:
As it so happens, this isn't just a theoretical problem: the build failure is happening precisely due to this situation arising in the rspec-support spec suite:

https://github.com/rspec/rspec-support/blob/4b938ec512d2fbee945913afbb957b510244ecb3/spec/rspec/support/fuzzy_matcher_spec.rb#L49-L52

severinraez:
Wild thought about an alternative approach: Could ExpectationNotMetError carry information about the matcher that caused it and that information then in turn be used to delegate to the proper `#failure_message` ?

JonRowe:
@severinraez thats not going to help, we deliberately swallow the exception for custom defined matchers which is the issue, your suggestion would help for multiple nested exceptions but thats not what's going on here.

@myronmarston I can't think of an elegant way to solve the block equality issue atm, wrapping both seems kinda hacky and I can't find a way to check an exceptions source location atm

myronmarston:
> @myronmarston I can't think of an elegant way to solve the block equality issue atm, wrapping both seems kinda hacky and I can't find a way to check an exceptions source location atm

I think we're going to be better off if we make this configurable on a per matcher basis.  Something like:

``` ruby
RSpec::Matchers.define :my_matcher do
  match(notify_expectation_failures: true) do |actual|
    # ...
  end
end
```

The idea is that `notify_expectation_failures: true` makes the matcher simply allow expectation failures to bubble up normally instead of rescuing them.  No trickery or magic.

Unfortunately, I'd rather it be the other way; having it notify by default is safer since people who would benefit from the option aren't likely to realize it exists or they need it.  It'd be better to do something like `match(handle_expectation_failures: true)` to opt-in to the existing behavior.  But changing to that would be a breaking change.

Given that we've only ever gotten one complaint about this (#890), maybe we should leave this as-is in RSpec 3 and then change to make it opt-in in RSpec 4 via an API like what I've suggested above?

JonRowe:
@myronmarston I fixed this up to make it optional

JonRowe:
Anyone want to re-review @rspec/rspec ?

features/custom_matchers/define_block_matcher.feature
```diff
@@ -42,3 +42,28 @@ Feature: define a matcher supporting block expectations
...
+          expect {
+            expect {
+              raise RSpec::Expectations::ExpectationNotMetError.new('original')
+            }.to support_blocks_with_errors
+          }.to raise_error(RSpec::Expectations::ExpectationNotMetError, /original/)
+        end
+      end
+      """
+    When I run `rspec ./block_matcher_spec.rb`
+    Then the example should pass
```
myronmarston: I don't think this scenario works very well as documentation -- it's written very similarly to our specs but doesn't approximate real usage.  A few things I'd recommend:

* Don't use `expect { }.to raise_error(RSpec::Expectations::ExpectationNotMetError)` -- instead, change the cucumber step from `Then the example should pass` to `Then it should fail with....` (or whatever the wording is for that).  After all, when users are using this feature they're not going to be wrapping things with `expect { }.to raise_error(RSpec::Expectations::ExpectationNotMetError)`.
* Don't use `raise RSpec::Expectations::ExpectationNotMetError.new('original')` -- instead just use a failing expectation in there.  Again, users don't write `raise RSpec::Expectations::ExpectationNotMetError.new('original')` and using it here obscures things, I think.

lib/rspec/matchers/dsl.rb
```diff
@@ -56,15 +56,16 @@ module Macros
         #     expect(4).not_to be_even # fails
         #
         # @yield [Object] actual the actual value (i.e. the value wrapped by `expect`)
-        def match(&match_block)
+        def match(opts={}, &match_block)
```
myronmarston: Can you update the API docs for this method to mention the option?

features/custom_matchers/define_block_matcher.feature
JonRowe: Done.

lib/rspec/matchers/dsl.rb
JonRowe: And done.

JonRowe:
Care to re-review?

myronmarston:
LGTM

PR #895, opened 2016-01-26
Raise error deprecation

williamcodes:
Modify deprecation warning error as per discussion in issue #286

PR #896, opened 2016-01-26
Modify deprecation warning for raise_error matcher

williamcodes:


JonRowe:
See conversation in #895 :)

PR #897, opened 2016-01-27
Expand warning message and reduce max line count for file to 100

williamcodes:
I'm honestly not sure how necessary this is anymore because the error message is significantly clearer in 3.x than it was in 2.x, however I reduced the max line count for the file, which it seems like the maintainers are trying to work toward based on the comments in the rubocop yaml file, so hopefully that adds some value.


JonRowe:
This actually increases the max line count ;) possibly decreases overall line length though :P, merging despite build failure because it's from rspec-rails with 1.8.7

PR #898, opened 2016-02-10
Fix deprecation check for `should` with BasicObject.

jcoleman:
The deprecation check uses `__method__` which is a Kernel method
that is not available on BasicObject. We can call it directly on
Kernel instead so that we avoid any `method_missing` chains that
may attempt to dispatch it to BasicObject.

### DIFF
--- a/lib/rspec/expectations/syntax.rb
+++ b/lib/rspec/expectations/syntax.rb
@@ -41,12 +41,12 @@ module RSpec
 
         syntax_host.module_exec do
           def should(matcher=nil, message=nil, &block)
-            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(__method__)
+            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(::Kernel.__method__)
             ::RSpec::Expectations::PositiveExpectationHandler.handle_matcher(self, matcher, message, &block)
           end
 
           def should_not(matcher=nil, message=nil, &block)
-            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(__method__)
+            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(::Kernel.__method__)
             ::RSpec::Expectations::NegativeExpectationHandler.handle_matcher(self, matcher, message, &block)
           end
         end
--- a/spec/rspec/expectations/extensions/kernel_spec.rb
+++ b/spec/rspec/expectations/extensions/kernel_spec.rb
@@ -43,6 +43,16 @@ RSpec.describe Object, "#should" do
     it 'works properly on BasicObject-subclassed proxy objects' do
       expect(proxy_class.new(Object.new)).to be_proxied
     end
+
+    it 'does not break the deprecation check on BasicObject-subclassed proxy objects' do
+      begin
+        should_enabled = RSpec::Expectations::Syntax.should_enabled?
+        RSpec::Expectations::Syntax.enable_should unless should_enabled
+        proxy_class.new(BasicObject.new).should be_proxied
+      ensure
+        RSpec::Expectations::Syntax.disable_should if should_enabled
+      end
+    end
   end
 end
 


lib/rspec/expectations/syntax.rb
```diff
@@ -41,12 +41,12 @@ def enable_should(syntax_host=default_should_host)
 
         syntax_host.module_exec do
           def should(matcher=nil, message=nil, &block)
-            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(__method__)
+            ::RSpec::Expectations::Syntax.warn_about_should_unless_configured(Kernel.__method__)
```
myronmarston: This should probably use `::Kernel` instead of `Kernel`.  Inside a basic object, you need to fully qualify constants:

```
irb(main):003:0> class Foo < BasicObject
irb(main):004:1> Kernel
irb(main):005:1> end
NameError: uninitialized constant Foo::Kernel
        from (irb):4:in `<class:Foo>'
        from (irb):3
        from /Users/myron/.rubies/ruby-2.1.5/bin/irb:11:in `<main>'
```

```
irb(main):012:0> class Foo < BasicObject
irb(main):013:1> ::Kernel
irb(main):014:1> end
=> Kernel
```

jcoleman: Changed.

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -43,6 +43,16 @@ def method_missing(name, *args)
...
       expect(proxy_class.new(Object.new)).to be_proxied
     end
+
+    it 'does not break the deprecation check on BasicObject-subclassed proxy objects' do
+      begin
+        should_enabled = RSpec::Expectations::Syntax.should_enabled?
+        RSpec::Expectations::Syntax.enable_should unless should_enabled
+        proxy_class.new(BasicObject.new).should be_proxied
+      ensure
+        RSpec::Expectations::Syntax.disable_should if should_enabled
```
JonRowe: This call is idempotent and you will always enable should above so this if check can be dropped.

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -43,6 +43,16 @@ def method_missing(name, *args)
     it 'works properly on BasicObject-subclassed proxy objects' do
       expect(proxy_class.new(Object.new)).to be_proxied
     end
+
+    it 'does not break the deprecation check on BasicObject-subclassed proxy objects' do
+      begin
+        should_enabled = RSpec::Expectations::Syntax.should_enabled?
+        RSpec::Expectations::Syntax.enable_should unless should_enabled
```
JonRowe: This call is idempotent so you don't have to check that it's not enabled before calling it

JonRowe:
Nice find! Just a bit of clean up and I'll get this merged

myronmarston:
BTW, what concrete problem are you seeing?  Is there an error that you get without this fix?  I'm trying to understand how this worked OK before.

JonRowe:
@myronmarston `__method__` doesn't exist on `BasicObject`, so checking `should` exists in that way would break on one.

myronmarston:
> @myronmarston __method__ doesn't exist on BasicObject, so checking should exists in that way would break on one.

But we run our specs (which include some specs against the `should` syntax) on Rubies that have `BasicObject` all the time, and we don't get any failures.  Why?  Does it fail in the wild if you enable the `should` syntax on a ruby interpretter that has `BasicObject` but not fail in our suite somehow?

spec/rspec/expectations/extensions/kernel_spec.rb
jcoleman: @JonRowe Shouldn't we only disable though if it was previously disabled before? Without this check if it's enabled before this test starts it will be disabled after, which seems wrong.

jcoleman:
@myronmarston We discovered it in the wild, specifically using Octopus to shard in ActiveRecord. Specifically, the issue happened with a test against an object that implements `method_missing` (well, a chain of `method_missing`s actually) but the object (or the final object; can't remember now) is an instance of a subclass of `BasicObject`, and so it tries to call `__method__` on that object which doesn't implement it.

See the bug report on Octopus: https://github.com/thiagopradi/octopus/issues/309

jcoleman:
@myronmarston To clarify, it doesn't break just because `BasicObject` exists, it only breaks if the `should` deprecation check ends up executing in a context that is `BasicObject` instead of `Object`.

PR #900, opened 2016-02-23
Allow raising on false positives

### LINKED ISSUE #888: Suggestion: raise_about_potential_false_positives option
In the `expect { }.to raise_error` (without explicit exception) War™ (see https://github.com/rspec/rspec-expectations/issues/655), I'm of the opinion that raise_error without an argument is dangerous enough to wish to _prevent_ its authorship in projects - not just a warning in the output that can be ignored. 

Perhaps a `raise_about_potential_false_positives` option should be added? I'm assuming `on_potential_false_positives = :warn | :raise` is too much of an API change.

Considering that talk in the aforementioned issue is that this behavior might not be allowed at all in RSpec 4, a stricter prevention mechanism would only encourage a smoother transition in an upgrade.

JonRowe:
Change our `warn_about_false_positives` behaviour to also allow raising on false positives, for those that like an even stricter behaviour. (we allow this with deprecations so why not!)

Closes #888

### DIFF
--- a/lib/rspec/expectations/configuration.rb
+++ b/lib/rspec/expectations/configuration.rb
@@ -22,6 +22,7 @@ module RSpec
       FALSE_POSITIVE_BEHAVIOURS =
         {
           :warn    => lambda { |message| RSpec.warning message },
+          :raise   => lambda { |message| raise ArgumentError, message },
           :nothing => lambda { |_| true },
         }
 
@@ -162,7 +163,7 @@ module RSpec
       # Configures what RSpec will do about matcher use which will
       # potentially cause false positives in tests.
       #
-      # @param [Symbol] behavior can be set to :warn or :nothing
+      # @param [Symbol] behavior can be set to :warn, :raise or :nothing
       def on_potential_false_positives=(behavior)
         unless FALSE_POSITIVE_BEHAVIOURS.key?(behavior)
           raise ArgumentError, "Supported values are: #{FALSE_POSITIVE_BEHAVIOURS.keys}"
--- a/lib/rspec/expectations/configuration.rb
+++ b/lib/rspec/expectations/configuration.rb
@@ -151,19 +151,35 @@ module RSpec
       # @param [Boolean] boolean
       def warn_about_potential_false_positives=(boolean)
         if boolean
-          @on_potential_false_positives = :warn
+          self.on_potential_false_positives = :warn
         elsif warn_about_potential_false_positives?
-          @on_potential_false_positives = :nothing
+          self.on_potential_false_positives = :nothing
         else
           # no-op, handler is something else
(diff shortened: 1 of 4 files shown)


chrisarcand:
Thanks for getting this rolling :heart: I sadly just haven't had time to follow up that issue.

JonRowe:
Thoughts @rspec/rspec?

lib/rspec/expectations/configuration.rb
```diff
@@ -141,14 +149,43 @@ def self.format_backtrace(backtrace)
       # Configures whether RSpec will warn about matcher use which will
       # potentially cause false positives in tests.
       #
-      # @param value [Boolean]
-      attr_writer :warn_about_potential_false_positives
+      # @param [Boolean] boolean
+      def warn_about_potential_false_positives=(boolean)
+        if boolean
+          self.on_potential_false_positives = :warn
+        elsif warn_about_potential_false_positives?
```
mrageh: Why do we set `on_potential_false_positives` to `:nothing` when it's `:warn`?

JonRowe: As the value could be `:raise` and `warn_about_potential_false_positives = false` shouldn't change that, so we only change it to `:nothing` when it's `:warn` already.

mrageh: Ok that kind of makes sense, but I am missing some context. Why should'nt we change the value of `on_potential_false_positives` when it's  set to `:raise` and `warn_about_potential_false_positives=false`?

JonRowe: `warn_about_potential_false_positives = false` controls wether we issue a warning or not on encountering a potential false positive. If we are raising an exception when encountering a potential false positive we are not issuing a warning, so we don't need to change anything.

mrageh: :+1: thanks then this LGTM

chrisarcand:
Code looks great to me, @JonRowe! My only question: You seemed to deliberately leave the older `warn_about...` config option in the warning message; I see slight value in leaving it, but maybe we should recommend the newer option?

Would introduce this flexibility to users more quickly and have more people using the option sooner to make upgrading later (I think the old option should be removed at some point...more confusing to keep both) easier. No harm to those that already have the `warn_about...` option set anyway, as they won't see the message to begin with. Thoughts?

samphippen:
LGTM

samphippen:
oh, needs a changelog entry.

JonRowe:
Sorry @samphippen, tweaked the message if you wouldn't mind casting an eye over that, changelog will follow post merge (that's just how I do them :))

samphippen:
:+1:

PR #904, opened 2016-03-17
Fix minitest integration to work on Minitest::Spec 5.6+.

### LINKED ISSUE #902: Minitest breaks rspec-expectations minitest_integration
Hi there,

first, thank you for this great gem!

While upgrading our application from Rails 3 to Rails 4 we run into the same problem as described [here](https://github.com/codeodor/minitest-rspec_mocks/issues/4) with rspec-expectations.

I'm not sure if the workaround I've found [here](https://github.com/codeodor/minitest-rspec_mocks/commit/e67ac54402843878671e11ea407df86730ea8c0c) is the way to go and if similar code should be added to `lib/rspec/expectations/minitest_integration.rb`.

Please let me know what you think about.

myronmarston:
Minitest::Spec 5.6+ introduced an `expect` method which
was conflicting which ours. To make our integration work
we have to define `to`/`not_to`/`to_not` on `Minitest::Expectation`.

Fixes #902.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,8 @@ Enhancements:
   `notify_expectation_failures: true` option for the `match` method to
   allow expectation failures to be raised as normal instead of being
   converted into a `false` return value for `matches?`. (Jon Rowe, #892)
+* Make `rspec/expectations/minitest_integration` work on Minitest::Spec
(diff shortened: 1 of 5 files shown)


JonRowe:
LGTM!

PR #905, opened 2016-03-18
Add missing alias having_attributes for have_attributes matcher

yujinakayama:


myronmarston:
LGTM.  This definitely brings it into parity with the other matchers.

Issue #907, opened 2016-03-30
The `raise_block_syntax_error` message too confident.


alexaltair:
I'd like to suggest changing the message for this error (currently [here](https://github.com/rspec/rspec-expectations/blob/52bfebf53e8025371d2869aac2b9f272e5b60f3/lib/rspec/matchers/built_in/change.rb#L89)) which says "The block passed to the \`change\` matcher must use \`{ ... }\` instead of do/end".

My first thought was, "How does it know??" because I didn't expect the ruby parser to pass that information on. As it turns out, it doesn't know for sure, it just guesses based on which method received the block argument (discussion [here](https://github.com/rspec/rspec-expectations/issues/137#issuecomment-46356503)). This means that false positives and negatives are possible.

This passes;
```ruby
expect{ subject }.to((change do some_query.count end).by(1))
```

and this fails with the `raise_block_syntax_error`;
```ruby
expect{ subject }.to(change) { some_query.count }.by(1)
```

I'd like the message to say something like, "The \`to\` method does not take a block. Perhaps you used do/end instead of \`{ ... }\`?" (Feel free to close if this issue has been discussed and decided before.)

myronmarston:
If the error message is confusing, I definitely want to see it change.  Want to work up a PR?

JonRowe:
Closing in favour of #908

PR #908, opened 2016-03-30
Clarify what happened in error message.

alexaltair:
C.f. issue https://github.com/rspec/rspec-expectations/issues/907.

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -87,8 +87,8 @@ module RSpec
         end
 
         def raise_block_syntax_error
-          raise SyntaxError, "The block passed to the `change` matcher must " \
-            "use `{ ... }` instead of do/end"
+          raise SyntaxError, "Block not received by the `change` matcher. " \
+          "Perhaps you want to use `{ ... }` instead of do/end?"
         end
 
         def positive_failure_reason
--- a/spec/rspec/matchers/built_in/change_spec.rb
+++ b/spec/rspec/matchers/built_in/change_spec.rb
@@ -223,7 +223,7 @@ RSpec.describe "expect { ... }.to change { block }" do
   it "warns if passed a block using do/end instead of {}" do
     expect do
       expect {}.to change do; end
-    end.to raise_error(SyntaxError, /block passed to the `change` matcher/)
+    end.to raise_error(SyntaxError, /Block not received by the `change` matcher/)
   end
 
   it "provides a #description" do
@@ -257,7 +257,7 @@ RSpec.describe "expect { ... }.not_to change { block }" do
   it "warns if passed a block using do/end instead of {}" do
     expect do
       expect {}.not_to change do; end
-    end.to raise_error(SyntaxError, /block passed to the `change` matcher/)
+    end.to raise_error(SyntaxError, /Block not received by the `change` matcher/)
   end
 end
 


lib/rspec/matchers/built_in/change.rb
```diff
@@ -87,8 +87,8 @@ def initialize(receiver=nil, message=nil, &block)
         end
 
         def raise_block_syntax_error
-          raise SyntaxError, "The block passed to the `change` matcher must " \
-            "use `{ ... }` instead of do/end"
+          raise SyntaxError, "Block passed to the wrong method. Perhaps you " \
```
JonRowe: How about `Block not received by the `change` matcher.` Rather than to `the wrong method`, the later is ambigious as to where the message is coming from.

alexaltair: If I run
```ruby
expect{ subject }.to(change{}){ some_query.count }.by(1)
```
it throws the error. In that case there *is* a block given to the `change` matcher. The check that's done in the code is a positive `block_given?` check through the `to` method (and I think some others?). I'd like to write `Block passed to #{__method__} method` except that I haven't figured out how to do that yet.

JonRowe: If someone does this: `expect{ subject }.to(change{}){ some_query.count }.by(1)` they can expect a weird message because they're doing something completely unsupported, the goal is to provide a clearer message to the majority of use cases, which assumes one block attempting to be passed to the `change` matcher.

alexaltair: Good point.

spec/rspec/matchers/built_in/change_spec.rb
```diff
@@ -223,7 +223,7 @@ def ==(other)
   it "warns if passed a block using do/end instead of {}" do
     expect do
       expect {}.to change do; end
-    end.to raise_error(SyntaxError, /block passed to the `change` matcher/)
+    end.to raise_error(ArgumentError, /Block not received by the `change` matcher/)
```
JonRowe: I think this should still be a `SyntaxError` it's not a Ruby syntax error but it is an RSpec one.

yujinakayama: IMO it should be `ArgumentError` since `SyntaxError` means unparsable Ruby code. Also, `SytaxError` is _not_ a `StandardError`:

```bash
$ ruby -e 'p SyntaxError.ancestors'
[SyntaxError, ScriptError, Exception, Object, Kernel, BasicObject]
```

* http://ruby-doc.org/core-2.3.0/ScriptError.html
* http://ruby-doc.org/core-2.3.0/SyntaxError.html

JonRowe: Well from our perspective it is unparsable! Also not being a `StandardError` doesn't bother us, we inherit from `Exception` for our own errors on purpose.

alexaltair: To me a `SyntaxError` is really profound; it means we haven't even been able to convert a text file to a ruby program yet. Certainly there is a concept of proper Rspec syntax, but `SyntaxError` is Ruby's thing. I'd be totally comfortable with something like `RspecSyntaxError`.

JonRowe: I'm not happy changing this from it's existing usage, please change it back.

JonRowe:
Thanks, can you squash all of these into one commit please and then it LGTM

JonRowe:
Thanks

PR #915, opened 2016-05-25
Implement Range, keywords support for RespondTo matcher.

sleepingkingstudios:


features/built_in_matchers/respond_to.feature
```diff
@@ -19,6 +19,17 @@ Feature: `respond_to` matcher
...
+    expect(obj).to respond_to(:xyz).with_unlimited_arguments
+    ```
+
+  If your Ruby version supports keyword arguments, you can specify a list of keywords accepted
+  by the method.
+
+    ```ruby
+    expect(obj).to respond_to(:foo).with_keywords(:ichi, :ni)
+    expect(obj).to respond_to(:bar).with(2).arguments.and_keywords(:san, :yon)
+    expect(obj).to respond_to(:baz).with_arbitrary_keywords
```
JonRowe: Can you add some scenarios showing usage as portrayed above? They'd need to be tagged so they only run on the relevant rubies however.

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -67,18 +107,54 @@ def find_failing_method_names(actual, filter_method)
           @failing_method_names = @names.__send__(filter_method) do |name|
             @actual.respond_to?(name) && matches_arity?(actual, name)
           end
+          @failing_method_names
```
JonRowe: Is this necessary, the above should return that anyway right?

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -23,6 +26,43 @@ def with(n)
...
+        #
+        # @example
+        #   expect(obj).to respond_to(:message).with_keywords(:color, :shape)
+        # @example with an expected number of arguments
+        #   expect(obj).to respond_to(:message).with(3).arguments.and_keywords(:color, :shape)
+        def with_keywords(*keywords)
+          @expected_keywords = keywords
+          self
+        end
+        alias :and_keywords :with_keywords
```
JonRowe: I guess it's ok that technically you could to `respond_to(:message).and_keywords(:color, :shape)`

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -67,18 +107,54 @@ def find_failing_method_names(actual, filter_method)
...
+          expectation.keywords = @expected_keywords
+          expectation.expect_unlimited_arguments = @unlimited_arguments
+          expectation.expect_arbitrary_keywords  = @arbitrary_keywords
+
+          return true if expectation.empty?
 
           signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
-          Support::StrictSignatureVerifier.new(signature, Array.new(@expected_arity)).valid?
+
+          Support::StrictSignatureVerifier.new(signature, []).with_expectation(expectation).valid?
```
JonRowe: Does this ever not get an empty array passed in now?

JonRowe:
Some minor feedback, nothing major, hopefully the windows build issues will have been fixed if you rebase too

features/built_in_matchers/respond_to.feature
sleepingkingstudios: @JonRowe I could actually use some help with that, actually - I toyed around with it at the time but couldn't figure out how to tag scenarios to run or not run on specific rubies. Are there other examples I could use as a baseline?

lib/rspec/matchers/built_in/respond_to.rb
sleepingkingstudios: Doesn't hurt anything, I think, and makes the interface more fluent. Just to play devil's advocate, there are a couple of alternatives I can think of:

- Have the 'and_' methods raise an error if there is not already an expectation. Relatively simple to implement and test, but adds a bit of complexity and possibly frustration (e.g. if a user modifies an existing test from `respond_to().with().and_...` to `respond_to().and_...).

- Have the 'with` methods return an object that wraps the matcher and adds the `and_` methods. Much more complicated to implement and test.

Ultimately at least in my opinion the simplicity and being consistent with the object API outweighs the "I can do something kinda funny sounding but it works". As a comparison, RSpec doesn't require composable aliases - you can say `expect().to include(be_a String)`, even if the message would make more sense if you used the `an_instance_of` alias.

lib/rspec/matchers/built_in/respond_to.rb
sleepingkingstudios: At least when I wrote this initially, `Support::StrictSignatureVerifier` was being used in `RSpec::Mocks` as well, for the use case of "does this method match these specific arguments". Before this PR, we were basically faking that by creating an array of empty items to generate our expectation. The `with_expectation(expectation)` call creates an alternate method of defining the behavior the method is supposed to match, but unless that has changed `RSpec::Mocks` still uses the other approach (which makes sense, seeing as it has a list of actual arguments it wants to check).

We probably could make the second argument optional, although that might require another PR to `RSpec::Support`.

JonRowe: It should probably be optional if it's not used at times, another PR is fine.

sleepingkingstudios: https://github.com/rspec/rspec-support/pull/282

features/built_in_matchers/respond_to.feature
JonRowe: Tag them with something like `@skip-when-keyword-args-unsupported` and then add a hook like this into a file in `features/support`.
```Ruby
Around "@skip-when-keyword-args-unsupported" do |_, block|
  require 'rspec/support/ruby_features'
  block.call unless ::RSpec::Support::RubyFeatures.kw_args_supported?
end
```

JonRowe:
I've merged this as is, we're looking at an RSpec 3.5 release this week and I'd like to get this out there with it, but it'd be good to get some scenarios as a followup. If you don't have time this week ping me and I'll take a look at them.

sleepingkingstudios:
@JonRowe https://github.com/rspec/rspec-expectations/pull/921

I may have overdone the features a bit, actually, but that should be reasonably comprehensive. Also fixes a bug in generating failure messages for 3 or more expected keywords.

PR #916, opened 2016-05-28
Fix include matcher so that it provides a valid diff for hashes

### LINKED ISSUE #887: include matcher shows misleading diff for hashes
With the following example:

```ruby
RSpec.describe "hash matching" do
  it "shows a diff when include matcher fails" do
    actual = {foo: "bar", baz: "qux"}
    expected = {foo: "bar", baz: "quux"}
    expect(actual).to include(expected)
  end
end
```

I get this when it fails:

```
Failures:

  1) hash matching shows a diff when include matcher fails
     Failure/Error: expect(actual).to include(expected)
     
       expected {:foo => "bar", :baz => "qux"} to include {:baz => "quux"}
       Diff:
       @@ -1,2 +1,3 @@
       -[{:foo=>"bar", :baz=>"quux"}]
       +:baz => "qux",
       +:foo => "bar",
       
     # ./hash_include_spec.rb:5:in `block (2 levels) in <top (required)>'
```

There are a couple of things that strike me as wrong here:
1. It seems to indicate that the expected is an array, when it is a hash
2. It didn't show that the `:foo` value was correct

Could this be an issue? Or is this just showing the unavoidable complexity of checking for inclusion here?

yujinakayama:
This fixes #887.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -5,8 +5,11 @@ module RSpec
       # Provides the implementation for `include`.
       # Not intended to be instantiated directly.
       class Include < BaseMatcher
-        def initialize(*expected)
-          @expected = expected
+        # @private
+        attr_reader :expecteds
+
+        def initialize(*expecteds)
+          @expecteds = expecteds
         end
 
         # @api private
@@ -24,7 +27,7 @@ module RSpec
         # @api private
         # @return [String]
         def description
-          improve_hash_formatting("include#{readable_list_of(expected)}")
+          improve_hash_formatting("include#{readable_list_of(expecteds)}")
         end
 
         # @api private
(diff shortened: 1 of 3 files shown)


lib/rspec/matchers/built_in/include.rb
```diff
@@ -5,8 +5,11 @@ module BuiltIn
       # Provides the implementation for `include`.
       # Not intended to be instantiated directly.
       class Include < BaseMatcher
-        def initialize(*expected)
-          @expected = expected
+        # @private
+        attr_reader :expecteds
```
mrageh: Would the var name `expectations` make more sense?

JonRowe: It's only needed internally but we'd normally call this `expected_as_array`

JonRowe: Actually I think this is ok, it's not expectations because it's expected values but it's ambiguous as to what these are

PR #918, opened 2016-05-31
Fix diffing regression introduced by #916.

myronmarston:
We only want to return a single hash from `expected`
if only one hash was provided.  If multiple hashes
were provided we do not want to merge them for the
purposes of diffing.

Also, we tend to favor `klass === object` checks
over `object.is_a?(klass)` since `object` could
be any user object that defines `is_a?` however
(or does not define it at all), whereas monkey
patching `===` on a core type like `Hash` would
be far less common and we simply don't support that.

@JonRowe / @yujinakayama -- can you review this since you were involved in #916?

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -51,8 +51,8 @@ module RSpec
         # @api private
         # @return [Array, Hash]
         def expected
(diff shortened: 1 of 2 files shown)


JonRowe:
LGTM

PR #929, opened 2016-07-07
Updates from changes in rspec-support

ChaeOkay:
See [rspec-support PR #290](https://github.com/rspec/rspec-support/pull/290)

Fixes [rspec-support #70](https://github.com/rspec/rspec-support/issues/70)

### DIFF
--- a/features/diffing.feature
+++ b/features/diffing.feature
@@ -25,7 +25,7 @@ Feature: diffing
     Then the output should contain:
       """
              Diff:
-             @@ -1,4 +1,4 @@
+             @@ -1,5 +1,5 @@
               this is the
              -  expected
              +  actual
@@ -72,7 +72,33 @@ Feature: diffing
     Then the output should contain:
       """
              Diff:
-             @@ -1,2 +1,4 @@
+             @@ -1,2 +1,5 @@
+             -/expected/m
+             +this is the
+             +  actual
+             +    string
+      """
+
+  Scenario: diff for a multiline string and a regexp
+    Given a file named "example_spec.rb" with:
+      """ruby
+      RSpec.describe "a multiline string" do
+        it "is like another string" do
+          expected = /expected/m
+          actual = <<-ACTUAL
+      this is the
+        actual
+          string
+      ACTUAL
+          expect(actual).to match expected
+        end
+      end
+      """
+    When I run `rspec example_spec.rb`
+    Then the output should contain:
+      """
+             Diff:
+             @@ -1,2 +1,5 @@
              -/expected/m
              +this is the
              +  actual
--- a/features/diffing.feature
+++ b/features/diffing.feature
@@ -38,7 +38,7 @@ Feature: diffing
       RSpec.describe "a multiline string" do
         it "is like another string" do
           expected = "Chewy\nbagel."
-          actual = "Chewy\nbagel.\n"
+          actual = "Chewy\n\nbagel.\n"
           expect(actual).to eq(expected)
         end
       end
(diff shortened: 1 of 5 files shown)


features/diffing.feature
```diff
@@ -25,13 +25,60 @@ Feature: diffing
...
+      end
+      """
+    When I run `rspec example_spec.rb`
+    Then the output should contain:
+      """
+             Diff:
+             @@ -1,3 +1,4 @@
+              Chewy
+              bagel.
+             +
```
JonRowe: This output is slightly wrong, it's actually `-bagel.` `+bagel.` but with a correct line ending, I'm not sure how to reflect this yet but this isn't the way. (Posix specification specifies all lines end with a \n, git would normally warn about the line ending at the end)

ChaeOkay: Ah right.

JonRowe: Git prints `\ No newline at end of file` perhaps we should be looking to print `\ No newline at end of input`

ChaeOkay: Actually, this output is what I'm suggesting as the fix :)

From what I understood, the goal was to represent a trailing newline in the diff output instead of nothing at all. This option represents trailing newline diffs the same ways that other newlines diffs in non-trailing situations do. 

I explored adding a string representation where trailing newlines occurred as git does, but it became pretty involved since we add and remove trailing newlines as the output message is crafted in several areas of the code. 

Do you have any ideas of what you'd like to see instead?

ChaeOkay: Just for the records, [google-diff-match-patch](https://code.google.com/p/google-diff-match-patch/) represents trailing newlines no different than other newlines. 

Are you sure you want all of the code involved with putting in "\ No newline at end of input" in?  I'm totally for it, just checking one more time :)

JonRowe: This output doesn't represent a trailing newline tho, it represents `previous\n\n`, which would currently be diffed like this correctly. Github's web interface would show the little red symbol at the end of the line.

ChaeOkay: Seems reasonable, will continue at it!

ChaeOkay: Hm, do we want a `+` or `-` in front of `\ No newline at end of input` depending on how it relates to the expected?

Also why I thought maybe just a `+` or `-` might be acceptable:
screen shot 2016-07-06 at 9 16 30 pm
screen shot 2016-07-06 at 9 15 53 pm

JonRowe: If you looked at the characters of that file you would find `Chewy\nbagel.\n\n` `Chewy\nbagel.\n`

JonRowe: Having `+\ No newline at end of input` `-\ No newline at end of input` works for me.

ChaeOkay: Updated specs, still working on escaping, not sure what's going on there.

So, I was using VIM to create t1.txt! I ended up re-learning `$ echo -n` to see the ouput message we are talking about! Thanks for being patient with me.

features/diffing.feature
```diff
@@ -51,13 +73,59 @@ Feature: diffing
...
              Diff:
-             @@ -1,2 +1,4 @@
+             @@ -1,2 +1,5 @@
              -/expected/m
              +this is the
              +  actual
              +    string
       """
 
+  Scenario: diff for a multiline string and a regexp
```
JonRowe: This seems to be a duplicate of another scenario, was this committed accidentally?

ChaeOkay: Yes.

PR #936, opened 2016-08-11
Reduce allocations

nathell:
I and @k3rni have profiled rspec-expectations' test suite for memory allocations using `allocation_stats`. The results are at http://danieljanus.pl/rspec-allocations.txt. Most are non-trivial to change, but we did find some (not much) low-hanging fruit. This is a PR for #766.

Specific changes are documented in individual commit messages.

### DIFF
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -4,7 +4,7 @@ module RSpec
       # @private
       def differ
         RSpec::Support::Differ.new(
-          :object_preparer => self,
+          :object_preparer => DifferObjectPreparer,
           :color => RSpec::Matchers.configuration.color?
         )
       end
@@ -27,7 +27,10 @@ module RSpec
         RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
       end
 
-      def call(object)
+    end
+
+    module DifferObjectPreparer
+      def self.call(object)
         RSpec::Matchers::Composable.surface_descriptions_in(object)
       end
     end
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -4,7 +4,7 @@ module RSpec
       # @private
       def differ
         RSpec::Support::Differ.new(
-          :object_preparer => lambda { |object| RSpec::Matchers::Composable.surface_descriptions_in(object) },
+          :object_preparer => self,
           :color => RSpec::Matchers.configuration.color?
         )
       end
@@ -25,6 +25,10 @@ module RSpec
         message = ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(expected).message_with_diff(message, differ, actual)
 
         RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
+      end
+
+      def call(object)
+        RSpec::Matchers::Composable.surface_descriptions_in(object)
       end
     end
   end
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -64,8 +64,8 @@ module RSpec
       def diffs(differ, actual)
         @expected_list.inject("") do |output, (expected, diff_label)|
           diff = differ.diff(actual, expected)
-          output << diff_label << diff << "\n".freeze unless /\A[[:space:]]*\z/ === diff
-          output
+          next output if /\A[[:space:]]*\z/ === diff
+          output << diff_label << diff << "\n".freeze
         end.tap(&:rstrip!)
       end
     end
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -62,11 +62,11 @@ module RSpec
       end
 
       def diffs(differ, actual)
-        @expected_list.map do |(expected, diff_label)|
+        @expected_list.inject("") do |output, (expected, diff_label)|
           diff = differ.diff(actual, expected)
-          next if diff.strip.empty?
-          "#{diff_label}#{diff}"
-        end.compact.join("\n")
+          output << diff_label << diff << "\n".freeze unless /\A[[:space:]]*\z/ === diff
+          output
+        end.tap(&:rstrip!)
       end
     end
   end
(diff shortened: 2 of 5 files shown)


spec/rspec/expectations/fail_with_spec.rb
```diff
@@ -47,7 +47,6 @@
       |@@ -1,2 +1,2 @@
       |-["poo", "car"]
       |+[(a string matching /foo/), (a string matching /bar/)]
-      |
```
JonRowe: We're not (generally) interested in reducing allocations in our own test suite, and I'm pretty certain this is a change in behaviour, can you restore it?

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -439,7 +439,7 @@ def foo
         diff = e.message.sub(/\A.*Diff:/m, "Diff:").gsub(/^\s*/,'')
       end
 
-      expect(diff).to eq "Diff:\n@@ -1,3 +1,3 @@\n-line1\n+LINE1\nline2\n"
+      expect(diff).to eq "Diff:\n@@ -1,3 +1,3 @@\n-line1\n+LINE1\nline2"
```
JonRowe: As above

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -62,11 +62,11 @@ def self.truncated(description)
...
       def diffs(differ, actual)
-        @expected_list.map do |(expected, diff_label)|
+        @expected_list.inject("") do |output, (expected, diff_label)|
           diff = differ.diff(actual, expected)
-          next if diff.strip.empty?
-          "#{diff_label}#{diff}"
-        end.compact.join("\n")
+          next output if /\A[[:space:]]*\z/ === diff
+          output << diff_label << diff << "\n".freeze
+        end.tap(&:rstrip!)
```
JonRowe: Given the changes to the specs below, are you sure this is identical behaviour?

nathell: See the note in the commit message.

> Refactor ExpectedsForMultipleDiffs#diffs to allocate less.
> This also subtly changes the behaviour of the `diffs` method.  In the
> old version, the diff would sometimes end with a newline, and sometimes
> not. After the refactor, it never does. I had to make some changes
> to tests to make the suite still pass; it is more consistent now.

> Check for blank was shamelessly stolen from ActiveSupport's
> `String#blank?`.

nathell: Of course, we can omit this particular change if it's deemed to affect stability, but I maintain the point that it makes things more consistent.

nathell: The other consistent behaviour would be to allow newlines at the end of a diff (obtainable by removing the final `tap(rstrip!)`). Then these tests would pass, but the following would fail:

```
rspec ./spec/rspec/expectations/fail_with_spec.rb:11 # RSpec::Expectations#fail_with includes a diff if expected and actual are diffable
rspec ./spec/rspec/expectations/fail_with_spec.rb:64 # RSpec::Expectations#fail_with with --color tells the differ to use color
rspec ./spec/rspec/expectations/fail_with_spec.rb:41 # RSpec::Expectations#fail_with with matchers uses matcher descriptions in place of matchers in diffs
rspec ./spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb:101 # RSpec::Matchers::ExpectedsForMultipleDiffs#message_with_diff returns message with diff and truncated matcher description if it is too long
rspec ./spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb:81 # RSpec::Matchers::ExpectedsForMultipleDiffs#message_with_diff returns regular message with diff when single expected
rspec ./spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb:90 # RSpec::Matchers::ExpectedsForMultipleDiffs#message_with_diff returns message with diff and matcher description when single expected with matcher
rspec ./spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb:50 # RSpec::Matchers::ExpectedsForMultipleDiffs.for_many_matchers has a diff for all matchers with their description
```

This shows that rspec-expectations' own test suite currently expects inconsistent behaviour (likely to match an implementation artifact) wrt newlines at the end of diffs.

myronmarston:
Thanks for taking a stab at this!  Could you add your allocation benchmarks to a directory under `benchmarks`?  See, for example, in rspec-core:

https://github.com/rspec/rspec-core/tree/master/benchmarks/allocations

Having files like these would answer all sorts of questions I have, and answer them for posterity by having the answers under source control:

* What code did you run the allocation tracking over?
* What was the number of allocations before, and what was it after? (IOW, what are the numbers that specific changes reduced?)

You may want to use a helper like https://github.com/rspec/rspec-core/blob/master/benchmarks/allocations/helper.rb; feel free to copy it over here.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -62,11 +62,11 @@ def self.truncated(description)
       end
 
       def diffs(differ, actual)
-        @expected_list.map do |(expected, diff_label)|
+        @expected_list.inject("") do |output, (expected, diff_label)|
           diff = differ.diff(actual, expected)
-          next if diff.strip.empty?
-          "#{diff_label}#{diff}"
-        end.compact.join("\n")
+          next output if /\A[[:space:]]*\z/ === diff
```
myronmarston: I don't think `[[:space]]` is supported in regexes on 1.8.7, so this is going to have to change to support 1.8.7, or you can define an alternate implementation for 1.8.7.  We often do that like so:

https://github.com/rspec/rspec-expectations/blob/a271631e9b8e1749be876dbcc65dbd7a6a40d2db/lib/rspec/matchers/built_in/match.rb#L77-L91

lib/rspec/expectations/fail_with.rb
```diff
@@ -26,6 +26,10 @@ def fail_with(message, expected=nil, actual=nil)
 
         RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
       end
+
+      def call(object)
+        RSpec::Matchers::Composable.surface_descriptions_in(object)
+      end
```
myronmarston: This exposes `RSpec::Expectations.call` as a method, which is undesirable.  It's not clear what "calling `RSpec::Expectations`" should mean.  Instead, can you define a module like this and pass that instead of `self` above?

``` ruby
module RSpec
  module Expectations
    module DifferObjectPreparer
      def self.call(object)
        RSpec::Matchers::Composable.surface_descriptions_in(object)
      end
    end
  end
end
```

JonRowe: Or prehaps assign the lamba to a constant and use that? Then it's only allocated once right?

JonRowe:
Thanks for kicking this off, I've just merged #941 which takes two of your ideas here, if you feel strongly that you can improve the diff allocations, please feel free to take another attempt at it, but I'd really like it to not change the current specs for the diffs.

PR #939, opened 2016-09-05
composable, dont enumarate sets

gajewsky:


myronmarston:
Thanks for the PR!  I took your change and revised it a little (there is a lot of nuance in how the composability works) in #940.  I plan to merge that once we get a code review.

PR #940, opened 2016-09-05
939 fixups

### LINKED ISSUE #878: `change.from().to()` does not work properly with sets
As reported by @listrophy on twitter:

https://twitter.com/listrophy/status/671832869946441728

The problem is happening in `Composable#values_match?`:

https://github.com/rspec/rspec-expectations/blob/acc409a0c4055ac0b1b098bd6df29d86807f233b/lib/rspec/matchers/composable.rb#L67

`with_matchers_cloned` treats any enumerable as an ordered list and converts it to an array as part of its logic.  This is bad as it breaks set equality.  We should fix `with_matches_cloned` to not do that.

More generally, I'm thinking that we should stop treating enumerables generically, and only handle arrays in this fashion.  Converting arbitrary enumerables to an array before checking equality is a bad idea.  Or maybe we can find a different way to handle the case `with_matchers_cloned` is there for.

myronmarston:
Revise the fix from #939 to work a bit better.

- We do not want to load stdlib `set` unnecessarily.
- I realized that liberally enumerating anything enumerable
  when cloning matchers is problematic and caused the bug in
  #878. Instead, we should only do it for matchers in arrays
  and hashes, as those are the data structures we support
  nested matchers in.
- This has the side effect of making ranges work with
  `contain_exactly` and preserves the fix for `change`
  to work with sets implemented in #939.

Fixes #878.

### DIFF
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -1,6 +1,7 @@
 module RSpec
   module Matchers
     module BuiltIn
+      # rubocop:disable ClassLength
       # @api private
       # Provides the implementation for `contain_exactly` and `match_array`.
       # Not intended to be instantiated directly.
@@ -85,7 +86,7 @@ module RSpec
         def convert_actual_to_an_array
           if actual.respond_to?(:to_ary)
             @actual = actual.to_ary
-          elsif should_enumerate?(actual) && actual.respond_to?(:to_a)
+          elsif actual.respond_to?(:to_a) && !to_a_disallowed?(actual)
             @actual = actual.to_a
           else
             return false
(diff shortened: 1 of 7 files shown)


lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -98,6 +98,22 @@ def safe_sort(array)
...
+          def to_a_allowed?(object)
+            case object
+              when NilClass, String then false
+              else Kernel != RSpec::Support.method_handle_for(object, :to_a).owner
+            end
+          end
+        else
+          def to_a_allowed?(object)
+            case object
+              when NilClass then false
```
JonRowe: Rubocop doesn't like this indenting :/

JonRowe:
This LGTM when rubocop is satisfied

PR #941, opened 2016-09-06
Alternate reduce allocations

JonRowe:
An alternative to #936, this takes the low hanging fruit but leaves the more controversial diff change in behaviour behind.

### DIFF
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -2,9 +2,17 @@ module RSpec
   module Expectations
     class << self
       # @private
+      class Differ
+        # @private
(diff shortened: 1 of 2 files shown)


mrageh:
How does this reduce the number of objects allocated?
The reasoning in #936 was hard to follow...

JonRowe:
> How does this reduce the number of objects allocated?
The reasoning in #936 was hard to follow...

The `lamba` keyword creates a new instance each time, where as assigning it to a constant does not.

myronmarston:
LGTM

mrageh:
Thanks @JonRowe, :ship: it

PR #951, opened 2016-11-03
Add configuration for max object formatter length

fables-tales:


samphippen:
Fixes: https://github.com/rspec/rspec-support/issues/252#issuecomment-257839067

r? @rspec/rspec

xaviershay:
kinda makes me sad that we need to extend configuration API further but LGTM pending green build

liam-m (changes requested):

spec/rspec/expectations/configuration_spec.rb
```diff
@@ -87,6 +87,20 @@ def delegated?; true; end
         end
       end
 
+      describe "#max_formatted_output_length=" do
+        let(:object_with_large_inspect_string) { Struct.new(:value).new("a"*300) }
+
+        it "sets the maximum object formatter length" do
+          config.max_formatted_output_length = 10
```
JonRowe: This needs to be restored after it's changed, or otherwise isolated, as it's currently leaking and breaking the build :(

JonRowe (changes requested):

benoittgt:
I'm interested to work on this. So we need to make this setting only available for a `context`, `describe`, `it` block? 

Does it means that every time a test run we should check after that `max_formatted_output_length` is not different from a default value. If it's different we change it to default?

Looks like a hack for me. 😟

Maybe I could work instead on displaying differences like mentioned here https://github.com/rspec/rspec-core/issues/2535#issue-312110983 even if this look quick hard to implement at first sight.

JonRowe:
@benoittgt no the setting is fine, its the test, the test needs to ensure that the config is restored to its original state afterwards

benoittgt:
Oh ok thanks @JonRowe. Will make a PR for that if it's ok.

JonRowe:
👍

benoittgt:
See https://github.com/rspec/rspec-expectations/pull/1056 for potential test fix.

PR #960, opened 2016-12-27
Raise an error when a block is passed to the BeWithin matcher

twalpole:
The BeWithin matcher is aliased as `within`, which collides with Capybaras `within` scoping method when used together.  This wouldn't really be an issue if errors occurred, however since Capybaras `within` scopes a passed block and the BeWithin matcher just ignores the passed block it can lead to tests silently passing when they haven't actually done anything.  This PR makes the BeWithin matcher raise an ArgumentError when passed a block so the collision becomes known to users.

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -410,6 +410,7 @@ module RSpec
     #   expect(result).to     be_within(0.5).of(3.0)
     #   expect(result).not_to be_within(0.5).of(3.0)
     def be_within(delta)
+      raise ArgumentError, "The RSpec BeWithin matcher does not take a block" if block_given?
       BuiltIn::BeWithin.new(delta)
     end
     alias_matcher :a_value_within, :be_within
--- a/spec/rspec/matchers/built_in/be_within_spec.rb
+++ b/spec/rspec/matchers/built_in/be_within_spec.rb
@@ -66,6 +66,12 @@ module RSpec
           expect(nil).to be_within(0.1).of(0)
         }.to fail_with("expected nil to be within 0.1 of 0, but it could not be treated as a numeric value")
       end
+
+      it "raises an error if a block is passed" do
+        expect {
+          within("test") { }
+        }.to raise_error(ArgumentError, /does not take a block/)
+      end
     end
 
     RSpec.describe "expect(actual).to be_within(delta).percent_of(expected)" do


JonRowe:
Generally, Capybara's DSL takes precedence, in what scenarios have you encountered this the other way around?

twalpole:
@JonRowe  It depends on require ordering and/or rspec include ordering.  I've come across a number of instances of it while trying to solve peoples issues, the latest was http://stackoverflow.com/questions/41137611/rspec-capybara-expectpage-and-page-within-cannot-see-loaded-page

The fact that it just silently skips the whole block while producing any type of error makes its difficult to diagnose, especially for beginners.

JonRowe:
Would be better to solve the problem by not overriding the original definition

twalpole:
@JonRowe Or remove the `within` alias of be_within from RSpec -- doesn't really fit with natural language anyway -  expect(a).to within(3)... ???

JonRowe:
Can't do that for back compatibility reasons

twalpole:
@JonRowe Ok then, if this isn't acceptable I'll close it and we can just go back to silent errors and RSpec's matchers can keep trampling Capybara's methods and vice versa

JonRowe:
I'd actually suggest that a PR to capybara to prevent it using methods defined in the host would be better (can't trample over things you own, capybara overrides us not the other way around) but I'm willing to work on a solution here, I would suggest adding a conditional so we don't override methods when aliasing matchers if they're in use already, and reporting an error to the user.

twalpole:
Can't do that due to backwards compatibility since `within` is used by pretty much ever Capybara user -- and to claim RSpec owns the `within` method name is ludicrous.  How about the `all` method added to RSpec a few versions back that broke a ton of peoples Capybara tests when it trampled over Capybaras method of the same name that had been in use for years?  If a user of Capybara chooses to also use RSpec then 99% of the time they want Capybara's `within` not RSpec's.  I've suggested a quick easy change that would at least provide some warning for users.  You don't like that change, and that's fine it's your project, but I've put in the amount of time I'm currently willing to spend on it.

JonRowe:
Ok thanks for the suggestion, the context of the execution is an RSpec class; we define minimal methods as best we can, and in general our methods are loaded first so they can be overridden by other libraries, but the problem here lies upstream, the vast majority of developers don't suffer the cause of this problem (capybara some how being loaded before rspec's own internals) and this is not the solution to this problem. I'm afraid I also don't have the time to invest to fix this correctly.

PR #963, opened 2017-01-06
Remove rubocop on 1.9.2

JonRowe:
2.2.1 is broken on 1.9.2, its a dependency of Rubocop.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,8 @@ end
 
 if RUBY_VERSION >= '1.9' && RUBY_VERSION <= '2.1'
   gem 'rubocop', "~> 0.23.0"
+  # The latest version (2.2.1) breaks on ruby 1.9.2
+  gem 'rainbow', "~> 2.1.0" if RUBY_VERSION == "1.9.2"
 end
 
 eval File.read('Gemfile-custom') if File.exist?('Gemfile-custom')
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -36,6 +36,8 @@ end
 
 if RUBY_VERSION >= '1.9' && RUBY_VERSION <= '2.1'
   gem 'rubocop', "~> 0.23.0"
+  # The latest version (2.2.1) breaks on ruby 1.9.2
+  gem 'rainbow', "~> 2.1.0" if RUBY_VERSION == "1.9.2"
```
myronmarston: Instead, why don't we limit rubocop to running only on 2.0?  I don't think we even need to run rubocop on multiple ruby versions.

JonRowe: No objections here.

JonRowe:
Ready for re-review @myronmarston

myronmarston:
The change looks fine but 1.9.2 is still failing -- is that expected?

JonRowe:
Yes, until rspec/rspec-rails#1769 is merged (its turtles all the way down).

JonRowe:
I should probably seperate my attempts to fix 1.8.7 from 1.9.2 on rspec-rails to finally get a green build.

PR #965, opened 2017-01-16
yield_with_args: match immediately

alyssais:
This change makes `yield_with_args(matcher)` evaluate the matcher as soon as the arguments are yielded, rather than after the block has run.

I'll demonstrate the problem this solves with an example using Rails's `ActiveRecord::Relation.create`. This method takes a block. It constructs a record object, yields it, saves the record in the database, then returns it.

I'd like to be able to write a test like this:

```ruby
it "yields the record before saving" do
  expect { |b| Dog.create(&b) }.to yield_with_args(be_new_record)
end

it "returns the saved record" do
  expect(Dog.create).not_to be_new_record
end
```

Currently, this doesn't work. The `be_new_record` matcher is only evaluated _after_ the block is finished executing, which means that the record has already been saved and so no longer matches the matcher.

This is a very quick patch to demonstrate the problem and show an example of how it might be fixed. I'm happy to clean it up, write some tests, and adapt it based on any feedback you may have for me, but I wanted to get an indication of whether the patch was likely to be accepted before I worked on it any further.

Thanks!

### DIFF
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -369,17 +369,17 @@ module RSpec
 
         # @private
         def matches?(block)
-          @actual = []
-          @args_matched_when_yielded = true
-          @total_calls = 0
+          args_matched_when_yielded = true
+          yield_count = 0
 
-          @probe = YieldProbe.probe(block) do |arg|
-            @actual << arg
-            check_latest_arg_matches
+          @probe = YieldProbe.probe(block) do |args|
+            args_matched_when_yielded &&= values_match?(@expected[yield_count], args)
+            yield_count += 1
           end
 
+          @actual = @probe.successive_yield_args
           return false unless @probe.has_block?
-          called_expected_number_of_times? && args_matched_when_yielded?
+          args_matched_when_yielded && yield_count == @expected.length
         end
 
         def does_not_match?(block)
@@ -414,20 +414,6 @@ module RSpec
           value = @expected[@total_calls]
           @total_calls += 1
           value
-        end
-
-        def called_expected_number_of_times?
-          @total_calls == @expected.size
-        end
-
-        def args_matched_when_yielded?
-          @args_matched_when_yielded
-        end
-
-        def check_latest_arg_matches
-          expected = next_expected
-          actual = @actual.last
-          @args_matched_when_yielded &&= values_match?(expected, actual)
         end
 
         def expected_arg_description
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -276,14 +276,15 @@ module RSpec
 
         # @private
         def matches?(block)
-          args_matched_when_yielded = true
+          @args_matched_when_yielded = true
           @probe = YieldProbe.new(block) do
             @actual = @probe.single_yield_args
-            args_matched_when_yielded &&= args_currently_match?
+            @actual_formatted = actual_formatted
+            @args_matched_when_yielded &&= args_currently_match?
           end
           return false unless @probe.has_block?
           @probe.probe
-          @probe.yielded_once?(:yield_with_args) && args_matched_when_yielded
+          @probe.yielded_once?(:yield_with_args) && @args_matched_when_yielded
         end
 
         # @private
@@ -328,10 +329,10 @@ module RSpec
         def negative_failure_reason
           if !@probe.has_block?
             'was not a block'
-          elsif all_args_match?
+          elsif @args_matched_when_yielded && !@expected.empty?
             'yielded with expected arguments' \
               "\nexpected not: #{surface_descriptions_in(@expected).inspect}" \
-              "\n         got: #{actual_formatted}"
+              "\n         got: #{@actual_formatted}"
           else
             'did'
           end
@@ -346,7 +347,7 @@ module RSpec
           unless (match = all_args_match?)
             @positive_args_failure = 'yielded with unexpected arguments' \
               "\nexpected: #{surface_descriptions_in(@expected).inspect}" \
-              "\n     got: #{actual_formatted}"
+              "\n     got: #{@actual_formatted}"
           end
 
           match
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -11,10 +11,7 @@ module RSpec
         def self.probe(block, &callback)
           probe = new(block, &callback)
           return probe unless probe.has_block?
-          probe.assert_valid_expect_block!
-          block.call(probe)
-          probe.assert_used!
-          probe
+          probe.probe
         end
 
         attr_accessor :num_yields, :yielded_args
@@ -29,6 +26,13 @@ module RSpec
 
         def has_block?
           Proc === @block
+        end
+
+        def probe
+          assert_valid_expect_block!
+          @block.call(self)
+          assert_used!
+          self
         end
 
         def to_proc
@@ -272,12 +276,14 @@ module RSpec
 
         # @private
         def matches?(block)
-          @probe = YieldProbe.probe(block) do |*args|
-            @actual = args
-            check_args_match
+          args_matched_when_yielded = true
+          @probe = YieldProbe.new(block) do
+            @actual = @probe.single_yield_args
+            args_matched_when_yielded &&= args_currently_match?
           end
           return false unless @probe.has_block?
-          @probe.yielded_once?(:yield_with_args) && args_matched_when_yielded?
+          @probe.probe
+          @probe.yielded_once?(:yield_with_args) && args_matched_when_yielded
         end
 
         # @private
@@ -329,14 +335,6 @@ module RSpec
           else
             'did'
           end
-        end
-
-        def args_matched_when_yielded?
-          @args_matched_when_yielded
-        end
-
-        def check_args_match
-          @args_matched_when_yielded = args_currently_match?
         end
 
         def args_currently_match?
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -417,7 +417,7 @@ module RSpec
         end
 
         def called_expected_number_of_times?
-          @total_calls == @expected.count
+          @total_calls == @expected.size
         end
 
         def args_matched_when_yielded?
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -21,7 +21,7 @@ module RSpec
 
         def initialize(block, &callback)
           @block = block
-          @callback = callback || proc {}
+          @callback = callback || Proc.new {}
           @used = false
           self.num_yields = 0
           self.yielded_args = []
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -21,7 +21,7 @@ module RSpec
 
         def initialize(block, &callback)
           @block = block
-          @callback = callback
+          @callback = callback || proc {}
           @used = false
           self.num_yields = 0
           self.yielded_args = []
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -8,8 +8,8 @@ module RSpec
       # yield matchers is used. Provides information about
       # the yield behavior of the object-under-test.
       class YieldProbe
-        def self.probe(block)
-          probe = new(block)
+        def self.probe(block, &callback)
+          probe = new(block, &callback)
           return probe unless probe.has_block?
           probe.assert_valid_expect_block!
           block.call(probe)
@@ -19,8 +19,9 @@ module RSpec
 
         attr_accessor :num_yields, :yielded_args
 
-        def initialize(block)
+        def initialize(block, &callback)
           @block = block
+          @callback = callback
           @used = false
           self.num_yields = 0
           self.yielded_args = []
@@ -37,6 +38,7 @@ module RSpec
           Proc.new do |*args|
             probe.num_yields += 1
             probe.yielded_args << args
+            @callback.call(*args)
             nil # to indicate the block does not return a meaningful value
           end
         end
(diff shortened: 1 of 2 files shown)


myronmarston:
Good catch!  Matching yielded args when they are yielded vs at the end is definitely a subtlety we didn't consider when we created the `yield` matchers.  I'd consider it a bug that we don't match when the args are yielded, because that's what `yield_with_args` describes.  So yes, we'd like to merge a PR that fixes this.

This is going to need some specs (as you pointed out) and I also think we need to apply this to the `yield_successive_args` matcher because it has the same issue.

One other thing: what you have in your PR so far adds a `args_matched?` method in addition to the `args_match?` method, which is bound to cause confusion.  WDYT about renaming them to `args_currently_matched?` and `args_matched_when_yielded?`?

alyssais:
> This is going to need some specs (as you pointed out)

On it

> I also think we need to apply this to the `yield_successive_args` matcher because it has the same issue.

👍

WDYT about renaming them to `args_currently_matched?` and `args_matched_when_yielded?`?

👍 (I mixed them up myself immediately after writing them, so that's not a good sign…)

myronmarston:
Actually, `args_currently_match?` (no -ed) is probably a better name given it's meant to be a present-tense thing :).

alyssais:
Oh, yeah

alyssais:
I wasn't quite sure exactly where the specs should go (which examples should be used).

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -617,6 +674,16 @@ def each_arg(*args, &block)
       }.not_to yield_successive_args(a_string_matching(/foo/), a_string_matching(/bar/))
     end
 
+    it 'passes when the successively yielded args match the matchers at yield time only' do
```
alyssais: I chose these names to match the ones above, but they don't actually look right to me. Shouldn't they be the other way round — e.g.

```ruby
it 'passes when the successively yielded args don't match the matchers'
```

alyssais:
Oh, wow, didn't realise 1.8.7 was still supported.

alyssais:
New CI failures look like a Gemfile needs updated or something…

spec/rspec/matchers/built_in/yield_spec.rb
myronmarston: Yes, you are right!  Please update them.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -628,6 +695,15 @@ def each_arg(*args, &block)
         |         got: ["food", "barn"]
       EOS
     end
+
+    it 'fails when the successively yielded args match the matchers at return time only' do
+      expect { |b|
+        %w[ barn food ].each do |val|
+          _yield_with_args(val, &b)
+          val.delete!(val)
```
myronmarston: I've not seen `String#delete!` before, and this looks pretty odd.  It takes me a minute to realize what it's doing.  WDYT about using `val.sub!(/.+/, '')` (e.g. replace the contents of the entire string with a blank string) instead?

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -617,6 +674,16 @@ def each_arg(*args, &block)
...
 
+    it 'passes when the successively yielded args match the matchers at yield time only' do
+      expect { |b|
+        %w[ barn food ].each do |eventual|
+          initial = String.new
+          _yield_with_args(initial, &b)
+          initial << eventual
+        end
+      }.not_to yield_successive_args(a_string_matching(/foo/), a_string_matching(/bar/))
+    end
```
myronmarston: This test isn't quite right.  You can tell by running `git checkout master -- lib` to undo your changes to `yield.rb`, and when you run this spec file, this spec passes.  Which means it passes both w/ your changes and w/o your changes, so it's not actually testing your changes.

I think the problem is that you have `%w[ barn food ]`, but then the `/foo/`, `/bar/` in the expectation -- since the order is different it already does not match, so yield time vs return time does not come into play.  To make yield time vs return time come into play, you need to switch the order of either `%w[ barn food ]` or the matchers passed to `yield_successive_args`.

If you make that change, you should be able to see this spec fail w/o your change to `yield.rb`, and pass with it, which means it is in fact testing it properly.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -628,6 +695,15 @@ def each_arg(*args, &block)
...
     end
+
+    it 'fails when the successively yielded args match the matchers at return time only' do
+      expect { |b|
+        %w[ barn food ].each do |val|
+          _yield_with_args(val, &b)
+          val.delete!(val)
+        end
+      }.not_to yield_successive_args(a_string_matching(/foo/), a_string_matching(/bar/))
+    end
```
myronmarston: This test also passes both on master and this branch, so it's not actually testing it properly.  Also, as your doc string says, you're expecting failure in this case, but you haven't wrapped it in `expect { ... }.to fail_with(...)` to express that it should fail with an expectation.

As above, this also needs to swap the order of either expected or actual to force return vs yield time to come into play.  Here's what I came up with:

``` ruby
it 'fails when the successively yielded args match the matchers at return time only' do
  expect {
    expect { |b|
      %w[ food barn ].each do |val|
        _yield_with_args(val, &b)
        val.delete!(val)
      end
    }.not_to yield_successive_args(a_string_matching(/foo/), a_string_matching(/bar/))
  }.to fail_with(dedent <<-EOS)
    |expected given block not to yield successively with arguments, but yielded with expected arguments
    |expected not: [(a string matching /foo/), (a string matching /bar/)]
    |         got: ["", ""]
  EOS
end
```

I *think* that's right, and it fails on master but passes on your branch.

(And yes, this is quite confusing!).

alyssais:
Thanks for such a detailed review! It might take me a while to get through it all, but I'll get started. :)

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -326,7 +331,15 @@ def negative_failure_reason
...
         end
 
-        def args_match?
+        def args_matched_when_yielded?
+          @args_matched_when_yielded
+        end
+
+        def check_args_match
+          @args_matched_when_yielded = args_currently_match?
+        end
```
alyssais: I just did it to be consistent with the other matcher. Inlining seems fine too.

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -390,8 +410,24 @@ def supports_block_expectations?
...
-        def args_match?
-          values_match?(@expected, @actual)
+        def next_expected
+          value = @expected[@total_calls]
+          @total_calls += 1
+          value
+        end
+
+        def called_expected_number_of_times?
+          @total_calls == @expected.count
```
alyssais: I had no idea. Thanks!

spec/rspec/matchers/built_in/yield_spec.rb
alyssais: Probably more obvious. 👍

My guess is it'd be a bit less efficient, but that doesn't matter. I had originally used `val.clear!` here, but that method isn't defined in 1.8 and I found `val.delete!(val)` as a 1.8-compatible alternative.

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -269,10 +272,12 @@ def initialize(*args)
 
         # @private
         def matches?(block)
-          @probe = YieldProbe.probe(block)
+          @probe = YieldProbe.probe(block) do |*args|
+            @actual = args
```
alyssais: Do you mean `@actual = @probe.successive_yield_args`?

alyssais:
> According to the message, it got what it expected, and yet it failed! This is a consequence of the fact that the failure message is generated at the end, after the yielded values have been mutated, so the mutated values are listed in the message even though the matching happened against the original values.

> I'm not sure of the best way to solve that yet. Thoughts?

This works correctly with `yield_with_args`, so I guess we should try to change `yield_successive_args` to generate failure messages at yield time like it does:


```ruby
unless values_match?(@expected[yield_count], args)
  args_matched_when_yielded = false
  # generate and store failure messages here
end
```

Right?

spec/rspec/matchers/built_in/yield_spec.rb
alyssais: > I _think_ that's right

I independently tried getting this test failing on master and passing on this branch and came up with the same thing, so that's a good sign!

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -390,8 +408,10 @@ def supports_block_expectations?
 
       private
 
-        def args_match?
-          values_match?(@expected, @actual)
+        def next_expected
+          value = @expected[@total_calls]
+          @total_calls += 1
+          value
         end
```
myronmarston: It looks like this method isn't used anywhere.  Can you remove it?

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -326,7 +337,7 @@ def negative_failure_reason
           end
```
myronmarston: Here in `negative_failure_reason` I am realizing that `actual_formatted` is evaluated, which means that it will include the yielded args as they exist at the end.  For parity with `positive_failure_reason` it would be good for it to show the args as they were when yielded.

I think the simplest solution might be to evaluate `actual_formatted` at yield time and store in a variable, which can be used for both the positive and negative failure messages.

Also: I believe that there is not test coverage for this, to enforce that the failure messages for `yield_with_args` include the formatted versions of what was yielded at the time they were yielded.  Can you add test coverage?

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -321,6 +329,16 @@ def each_arg(*args, &block)
...
     end
 
+    it 'fails if the matchers match at return time only' do
+      expect {
+        expect { |b|
+          val = [1]
+          _yield_with_args(val, &b)
+          val.clear
+        }.to yield_with_args(be_empty)
+      }.to fail_with(/but yielded with unexpected arguments/)
```
myronmarston: As I mentioned above, I think we need spec coverage of the fact that the failure messages shows the args as they were at yield time.  I think your `fail_with` expectation here could be expanded to cover that for the positive expectation case.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -335,6 +353,16 @@ def each_arg(*args, &block)
...
     end
 
+    it 'fails if the matchers match at yield time only' do
+      expect {
+        expect { |b|
+          val = []
+          _yield_with_args(val, &b)
+          val << 1
+        }.not_to yield_with_args(be_empty)
+      }.to fail_with(/expected given block not to yield with arguments, but did/)
```
myronmarston: ...and this seems like the spot where you could cover the negative failure message.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -535,6 +580,18 @@ def each_arg(*args, &block)
         expect { |b| [ [:a, 1], [:b, 3] ].each(&b) }.to yield_successive_args([:a, 1], [:b, 2])
       }.to fail_with(/but yielded with unexpected arguments/)
     end
+
+    it 'fails if the matched at return time only' do
```
myronmarston: I think you can remove "the" from this.

lib/rspec/matchers/built_in/yield.rb
alyssais: Am I right in thinking that it shouldn't be calling `all_args_match?` either, and should instead be checking whether the args matched when yielded?

alyssais: Ah, and because of that it's now failing since I've (locally) added test coverage for the messages as you requested.

alyssais:
Think that's all your `yield_with_args` feedback addressed. I still need to do the `yield_successive_args` fixes.

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -317,16 +329,16 @@ def expected_arg_description
         def negative_failure_reason
           if !@probe.has_block?
             'was not a block'
-          elsif all_args_match?
+          elsif @args_matched_when_yielded && !@expected.empty?
```
alyssais: I think this logic is the same at what was there previously (except this uses the yield-time values), but I'd appreciate a second opinion.

myronmarston: Seems fine.

myronmarston:
He @alyssais!  Thanks for taking this as far as you've taken it!  It looks like there are still some final fixes to `yield_successive_args` needed.  Do you want to take care of those or should one of us take this over and get it across the finish line?  We definitely want to see this included in 3.6!

alyssais:
@myronmarston hey! I've been a little inundated recently, but am expecting to have a bit more time from now so I'll try to get it finished. :)

alyssais:
I'm not sure how to address the shared example that's failing after the most recent change. It requires that failure messages are generated using a single call to `RSpec::Support::ObjectFormatter.format`, but I don't think it'll be possible to do it that way since we have to build up the failure message incrementally, calling `#inspect` on each object as it's generated.

myronmarston:
> I'm not sure how to address the shared example that's failing after the most recent change. It requires that failure messages are generated using a single call to `RSpec::Support::ObjectFormatter.format`, but I don't think it'll be possible to do it that way since we have to build up the failure message incrementally, calling `#inspect` on each object as it's generated.

Great question!  It's actually not requiring a specific number of calls to that method, but it is required that it is called, and that the output of that is used in the failure message somehow.  (Basically, it is requiring each matcher to use that for object formatting, instead of doing `object.inspect`, since that has some formatting niceties that we want applied universally to all matchers).

The problem is that the `actual` in that shared example for this matcher yields _nothing_, and previously it passed because it called `RSpec::Support::ObjectFormatter.format([])`.  Now that you call it for each element, it is not called at all.

So we just need to change the `:invalid_value` for this example.  This patch should do it:

``` diff
diff --git a/spec/rspec/matchers/built_in/yield_spec.rb b/spec/rspec/matchers/built_in/yield_spec.rb
index ccd0b49..7cec208 100644
--- a/spec/rspec/matchers/built_in/yield_spec.rb
+++ b/spec/rspec/matchers/built_in/yield_spec.rb
@@ -548,7 +548,7 @@ RSpec.describe "yield_successive_args matcher" do

   it_behaves_like "an RSpec matcher",
       :valid_value => lambda { |b| [1, 2].each(&b) },
-      :invalid_value => lambda { |b| _dont_yield(&b) } do
+      :invalid_value => lambda { |b| [3, 4].each(&b) } do
     let(:matcher) { yield_successive_args(1, 2) }
   end
```

Instead of not yielding at all being the failing example, we can have it yield different values than we expected.

alyssais:
That makes sense. Thanks!

I _think_ that's all of your feedback addressed. I might have missed something though because there was a lot to keep track of! :D

myronmarston:
Thanks @alyssais! This definitely improves the `yield` matchers significantly!

PR #968, opened 2017-01-22
Pin diff-lcs to under 1.3 if we're running 1.8.7

JonRowe:
`diff-lcs` 1.3 drops support for 1.8.7 so once again the builds have broken. I'm not sure if its worthwhile just pinning to `< 1.3` for all so I've attempted to only do it on `1.8.7`. 

We could also consider dropping them a note pointing out that they've just broken semver (we were already checking less than <2)

Counterpart to rspec/rspec-mocks#1140


xaviershay:
urgh, lgtm

JonRowe:
From discussions over on halostatue/diff-lcs#40 I don't think this is the fix, still investigating.

JonRowe:
From discussions over on halostatue/diff-lcs#40 I don't think this is the fix, still investigating.

PR #970, opened 2017-02-02
Correct grammar errors

ajdlp:
This corrects both "a" vs "an" and "its" vs "it's" grammar errors.


JonRowe:
Thanks, in future can you add `[skip ci]` to commits that only correct documentation or spec docs.

PR #972, opened 2017-02-16
Treat no method errors as not comparable

### LINKED ISSUE #971: Comparison matchers fail poorly when compared against non-comparable object
Example:

``` ruby
expect(foo: nil).to include(foo: a_value > 10)
```

Results in:

```
     Failure/Error: expect(foo: nil).to include(foo: a_value > 10)

     NoMethodError:
       undefined method `>' for nil:NilClass
```

Instead of trying to check `nil > 10` and getting an error, it should only run the comparison if `actual.respond_to?(:>)` and return `false` if it is non-comparable.

JonRowe:
Fixes #971

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+### Development
+[Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.6.0.beta2...master)
+
+Enhancements:
+
+* Treat NoMethodError as a failure for comparison matchers. (Jon Rowe, #972)
+
 ### 3.6.0.beta2 / 2016-12-12
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.6.0.beta1...v3.6.0.beta2)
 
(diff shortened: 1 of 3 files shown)


spec/rspec/matchers/built_in/contain_exactly_spec.rb
```diff
@@ -105,6 +105,10 @@ def array.send; :sent; end
     }.to fail_with(/expected collection contained/)
   end
 
+  it 'works with hashes' do
+    expect([{ :a => 1 }]).to_not contain_exactly({ :a => 1 }, { :a => 2 }) 
+  end
```
myronmarston: What does this spec have to do with the rest of this PR?  Seems completely unrelated...

JonRowe: Nothing, evidently I did a `git ci -a` oops :)

JonRowe:
Merging as only change was removing unrelated spec.

Issue #973, opened 2017-02-18
Scoped `alias_matcher` and `define_negated_matcher`.


reitermarkus:
Is there a way to create scoped alias matchers and negated matchers without rewriting it using `matcher do`?

I couldn't find anything about this in the documentation, so I assume this is not (yet) possible.

It would be great being able to do something like this:

```ruby
describe "something" do
  alias_matcher :have_failed, :be_failed
  define_negated_matcher :be_a_failure, :be_a_success

  ...
end
```

PR #974, opened 2017-02-18
Scoped `alias_matcher` and `define_negated_matcher`

reitermarkus:
This moves `alias_matcher` and `define_negated_matcher` to `RSpec::Matchers::DSL` in order to support defining scoped matchers.

Closes https://github.com/rspec/rspec-expectations/issues/973.


myronmarston:
Thanks for the PR! The change looks solid in general but I do have a few concerns:

* Most of us who work on RSpec regularly only run the specs locally (since they are so fast compared to the cukes), leaving the cukes to run on travis, unless we know they might be involved in the change we are making.  Thus, it's important to us to have spec coverage for this change, so it's covered by what we tend to run locally all the time.
* I'm not sure that this necessarily warrants a cuke at all; we don't document every last API in our cukes (that's what the API docs are for) and I'm not yet convinced this is a common enough need to warrant the cuke.  Thus, I think you can delete the cuke, and replace it with a spec instead.  (If you need help writing the spec, let us know).
* I'm a bit concerned with what this does to our docs.  I still consider `RSpec::Matchers.alias_matcher` and `RSpec::Matchers.define_negated_matcher` to be the "primary" APIs for this, with the scoped forms as a secondary thing.  (And our [new book](http://rspec.info/blog/2017/02/now-in-beta-effective-testing-with-rspec-3/) covers the `RSpec::Matchers` form but won't cover the scoped form).  Our yard API docs need to still show this API being available off of `RSpec::Matchers`.  I'm not sure how YARD will deal with your change here.  Can you try building the docs locally (`bundle exec yard server --reload`) and make sure these APIs are still discoverable in the docs for `RSpec::Matchers`, like they [currently do](http://rspec.info/documentation/3.5/rspec-expectations/RSpec/Matchers.html)?  You may need to add some YARD tags or directives to get this to do what we want.

Thanks!

reitermarkus:
> I still consider `RSpec::Matchers.alias_matcher` and `RSpec::Matchers.define_negated_matcher` to be the "primary" APIs for this, with the scoped forms as a secondary thing.

Is `RSpec::Matchers.define` also considered the “primary“ API?

> You may need to add some YARD tags or directives to get this to do what we want.

I have never used YARD, but I tried a bunch of things, and I couldn't get the methods added by `extend DSL` to be recognized as class methods for `RSpec::Matchers`. Not sure if there is a way to do that. As an alternative, I moved the examples for the methods back to `RSpec::Matchers`.

myronmarston:
> Is `RSpec::Matchers.define` also considered the “primary“ API?

I'd say so.  It's the only way I've ever defined a custom matcher.

Anyhow, to fix this, I think you can just move the `extend` from the [end of dsl.rb](https://github.com/rspec/rspec-expectations/blob/f1d80314ddb37933d49b096cf72d1834d8006e31/lib/rspec/matchers/dsl.rb#L466) and put it inline in the `Matchers` module in `matchers.rb`.  I believe YARD's parser will see the extension and will include the methods in its docs for us automatically.

reitermarkus:
> I believe YARD's parser will see the extension and will include the methods in its docs for us automatically.

I already tried that. It doesn't seem to change anything. The docs say `Extended by: DSL`, as before, but the extended methods are not shown as class methods.

lib/rspec/matchers/dsl.rb
```diff
@@ -2,7 +2,49 @@ module RSpec
...
+      # @see RSpec::Matchers
+      def define_negated_matcher(negated_name, base_name, &description_override)
+        alias_matcher(negated_name, base_name, :klass => AliasedNegatedMatcher, &description_override)
+      end
+
       # Defines a custom matcher.
+      #
+      # @param name [Symbol] the name for the matcher
+      # @yield [Object] block that is used to define the matcher.
+      #   The yielded arg, when specified, is the expected value.
```
myronmarston: Thanks for improving this!

reitermarkus: You're welcome. 😀

reitermarkus:
Do you have a rough ETA for 3.6.0, by any chance?

myronmarston:
> Do you have a rough ETA for 3.6.0, by any chance?

Not yet, sorry.

PR #977, opened 2017-03-05
Change the `warn_about_negative_false_positive` message

koic:
This PR put a blank for a sense of unity as a whole warning message.


xaviershay:
oh good catch, thank you!

PR #981, opened 2017-04-08
Fix regression introduced in #965.

myronmarston:


myronmarston:
It's not your fault! We didn't cover these cases with tests before. All you can do is keep the tests cases that do exist green, and you did that.

myronmarston:
Can someone from @rspec/rspec review this?

xaviershay:
looking

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -369,16 +363,18 @@ def initialize(*args)
...
         def matches?(block)
           @actual_formatted = []
+          @actual = []
           args_matched_when_yielded = true
           yield_count = 0
 
-          @probe = YieldProbe.probe(block) do |args|
-            @actual_formatted << RSpec::Support::ObjectFormatter.format(args)
-            args_matched_when_yielded &&= values_match?(@expected[yield_count], args)
+          @probe = YieldProbe.probe(block) do |*arg_array|
```
xaviershay: is adding the splat the key? Not immediately clear to me what in this change actually fixes it. (Inlining `successive_yield_args` seems good overall regardless though.)

Gemfile
```diff
@@ -12,6 +12,7 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
   end
 end
 
+gem 'coderay' # for syntax highlighting
```
xaviershay: not clear how this is related

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -563,6 +563,30 @@ def each_arg(*args, &block)
     expect(val).to be_nil
   end
 
+  it "works correctly when the method yield multiple args each time (passing case)" do
```
xaviershay: specs are clear, good.

myronmarston: But your comment helped me realize "yield" should be "yields" here so I fixed that :).

Gemfile
myronmarston: It's not related to the changes here other than I wanted syntax highlighting in the failure output while working on this, so I added it, and I think it's worth keeping :).

lib/rspec/matchers/built_in/yield.rb
myronmarston: The fix has two parts to it.  After the changes in #965, this block only accepted a single, non-splatted arg.  Ruby's semantics dictate that that makes the extra arguments get dropped on the floor.  As a result, it only compared the expected values against the _first_ yielded argument of each yield.

The splat is necessary so that we receive _all_ the arguments on each yield, so we can compare them to what is expected.  However, doing that is not enough.  It fixed the new specs I added but broke most (or all) of the other specs for this matcher.

That's because the splat causes the `arg_array` variable to _always_ be an array, even if only one value was yielded (in which case, it's an array of one value).  The `successive_yield_args` thing that I inlined was what was handling that before, but w/o inlining the logic, it was only involved in determining `@actual` (since that's where it was called from) and not actually involved in the match logic.  By in-lining that method's logic, it ensures that if only one value is yielded, we deal with individual values, but if multiple values are yielded, we deal with an array for each yield.

Does that make sense?

Gemfile
xaviershay: fair

lib/rspec/matchers/built_in/yield.rb
xaviershay: it does. Ship it.

PR #983, opened 2017-04-21
use be truthy in be matcher

KevinColemanInc:
I noticed that there was some code duplication here and that it would make more sense for be to explicitly call `be_truthy` instead of returning the underlying object that `be_truthy` returns.

Should I also update the docs to mention that you can do `to be` instead of `to be_truthy`?

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -341,7 +341,7 @@ module RSpec
     # (e.g. be_empty), letting you choose the prefix that best suits the
     # predicate.
     def be(*args)
-      args.empty? ? Matchers::BuiltIn::Be.new : equal(*args)
+      args.empty? ? be_truthy : equal(*args)
     end
     alias_matcher :a_value, :be, :klass => AliasedMatcherWithOperatorSupport
 


JonRowe:
These aren't quite the same, `be` and `be_truthy` return different matchers.

KevinColemanInc:
There are rspec tests saying that this should have the same behavior
https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/matchers/built_in/be_spec.rb#L700

and here in the codebase, I see this comment:
https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/be.rb#L100

but it is being instantiated here.  Do you see the issue?

myronmarston:
> There are rspec tests saying that this should have the same behavior

Right, they have the same matching logic.  However, their failures messages and description are a bit different -- they correspond to the method used (`be` vs `be_truthy`).

> and here in the codebase, I see this comment:
> https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/be.rb#L100
> but it is being instantiated here. Do you see the issue?

That note is directed at end-users, not at ourselves.  The point of that note is that the end-user API is `be` instead of `RSpec::Matchers::Be.new`.  If we didn't instantiate the matchers we couldn't provide any of them to end-users.

KevinColemanInc:
Sorry, I misread that comment about instantiation.  that makes sense.  When I first saw it, I thought you were trying to say its an abstract class, but then I noticed `be_truthy` had a similar note.

The failure message for `be` is `"expected #{actual_formatted} to evaluate to true"` but this test is passing:
https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/matchers/built_in/be_spec.rb#L703

if the `be` class is supposed to test `if value == true` and not `truthy`, then that should evaluate to false, because `1 != true`?

Can I change the error message to `evaluate to truthy?` or fix the code to match the error message?

myronmarston:
> Sorry, I misread that comment about instantiation. that makes sense. When I first saw it, I thought you were trying to say its an abstract class, but then I noticed be_truthy had a similar note.

Right, all of our matcher classes have that note!

 > The failure message for be is `"expected #{actual_formatted} to evaluate to true"` but this test is passing:

You're right; the message isn't very clear we should improve that.

Although, part of me wonders if we should deprecate raw `be`; it's not very descriptive, IMO, and `be_truthy` has the same logic.  Still, we're generally waiting to put deprecation warnings in place until 3.99 so for now it would be good to improve the message.

KevinColemanInc:
the reason this came up was because I prefer the shorter syntax of `expect(1).to be`.  It is shorter and it reminds me of Shakespeare's "To be or not to be".

myronmarston:
OK.  Is there something about that that wasn't working for you?  Or is there an improvement to the failure message you have in mind?

KevinColemanInc:
I liked doing `to be` instead of `to be_truthy` because its shorter and easier to read.  

If your point is that `to be` has a different behavior than `to be_truthy`, then that test should fail.

if you're saying it has the same behavior as `to be_truthy`, then I think the underlying code for `to be` should call `to be_truthy` to make it more clear.

it seems like the only difference is the error message, but the `to be` error message doesn't correctly describe the behavior.

JonRowe:
>If your point is that to be has a different behavior than to be_truthy, then that test should fail.

> If you're saying it has the same behavior as to be_truthy, then I think the underlying code for to be should call to be_truthy to make it more clear.

It does and it doesn't, it behaves like `be_truthy` but `be_truthy` doesn't behave it `be`. e.g. `be_truthy <= 10` makes no sense but `be <= 10` does.

KevinColemanInc:
Currently, `be` behaves like `be_truthy` when there are no arguments passed in, but the error message does not agree with that.  It says it must be "true" not "truthy"

if you see in my PR when there are no arguments, then it behaves like `be_truthy` displaying the correct error message that matches the expectations (it must be "truthy").

To me, `be` and `be_truthy` are the exact same if no params are passed into `be`.  If params are passed into `be` then it needs to evaluate to true.

```
# current behavior
expect(nil).to be       # to evaluate to true
# Wrong, because it expects truthy, not true

expect(0).to be == 1 # to evaluate to true
# Correct, because it is expecting true

# correct behavior in my PR
expect(nil).to be       # expected: truthy value
expect(0).to be == 1 # should be true
```

myronmarston:
The failure message for raw `be` should be improved, I think.

JonRowe:
@KevinColemanInc if you'd like to improve the message, we'd welcome a PR for that.

PR #986, opened 2017-04-27
Fix incorrect comment in docs.

xaviershay:
It was contradicted by the expected output for the feature.


JonRowe:
LGTM, just kicked the build to see if it will go green

PR #987, opened 2017-05-03
Improve failure message of #change matcher with block and #satisfy matcher by including the block snippet

### LINKED ISSUE #937: Consider improving failure message of #change matcher by extracting block snippets
Currently failure message of the `change` matcher _with a block_ is sometimes unclear:

``` ruby
RSpec.describe '#do_something' do
  def do_something
    @foo = true
    @bar = false
  end

  before do
    @foo = false
    @bar = false
  end

  it 'changes @foo and @bar' do
    expect { do_something }.to change { @foo }
                          .and change { @bar }
  end
end
```

```
Failures:

  1) #do_something changes @foo and @bar
     Failure/Error:
       expect { do_something }.to change { @foo }
                             .and change { @bar }

       expected result to have changed, but is still false
```

It's not clear which the cause is `@foo` or `@bar` from the failure message. This is because RSpec does not know the snippets of the blocks and can only express them as `result`.

However, as we have introduced snippet extraction in https://github.com/rspec/rspec-core/pull/2083, extracting the block snippet may be possible in some cases.

I [tried implementing a prototype](https://github.com/rspec/rspec-expectations/commit/extract-block-snippet) and it seems to be working well. With the prototype, the above spec is reported as:

```
Failures:

  1) #do_something changes @foo and @bar
     Failure/Error:
       expect { do_something }.to change { @foo }
                             .and change { @bar }

       expected `@bar` to have changed, but is still false
```

@rspec/rspec What do you think?

yujinakayama:
This closes #937.

This PR depends on https://github.com/rspec/rspec-core/pull/2420 and https://github.com/rspec/rspec-support/pull/315.

### DIFF
--- a/features/support/ruby_features.rb
+++ b/features/support/ruby_features.rb
@@ -27,3 +27,13 @@ Around "@skip-when-required-keyword-args-unsupported" do |scenario, block|
     warn "Skipping scenario #{scenario.title} because required keyword arguments are not supported"
   end
 end
+
+Around "@skip-when-ripper-unsupported" do |scenario, block|
+  require 'rspec/support/ruby_features'
+
+  if ::RSpec::Support::RubyFeatures.ripper_supported?
+    block.call
+  else
+    warn "Skipping scenario #{scenario.title} because Ripper is not supported"
+  end
+end
--- a/lib/rspec/expectations/block_snippet_extractor.rb
+++ b/lib/rspec/expectations/block_snippet_extractor.rb
@@ -1,0 +1,248 @@
+module RSpec
+  module Expectations
+    # @private
+    class BlockSnippetExtractor # rubocop:disable Style/ClassLength
+      # rubocop should properly handle `Struct.new {}` as an inner class definition.
+
+      attr_reader :proc, :method_name
+
+      def self.try_extracting_single_line_body_of(proc, method_name)
+        lines = new(proc, method_name).body_content_lines
+        return nil unless lines.count == 1
+        lines.first
+      rescue Error
+        nil
+      end
+
+      def initialize(proc, method_name)
+        @proc = proc
+        @method_name = method_name.to_s.freeze
+      end
+
+      # Ideally we should properly handle indentations of multiline snippet,
+      # but it's not implemented yet since because we use result of this method only when it's a
+      # single line and implementing the logic introduces additional complexity.
+      def body_content_lines
+        raw_body_lines.map(&:strip).reject(&:empty?)
+      end
+
+    private
+
+      def raw_body_lines
+        raw_body_snippet.split("\n")
+      end
+
+      def raw_body_snippet
+        block_token_extractor.body_tokens.map(&:string).join
+      end
+
+      def block_token_extractor
+        @block_token_extractor ||= BlockTokenExtractor.new(method_name, source, beginning_line_number)
+      end
+
+      if RSpec.respond_to?(:world)
+        def source
+          RSpec.world.source_from_file(file_path)
+        end
+      else
+        RSpec::Support.require_rspec_support 'source'
+        def source
+          @source ||= RSpec::Support::Source.from_file(file_path)
+        end
+      end
+
+      def file_path
+        proc.source_location.first
+      end
+
+      def beginning_line_number
+        proc.source_location.last
+      end
+
+      Error = Class.new(StandardError)
+      TargetNotFoundError = Class.new(Error)
+      AmbiguousTargetError = Class.new(Error)
+
+      # @private
+      # Performs extraction of block body snippet using tokens,
+      # which cannot be done with node information.
+      BlockTokenExtractor = Struct.new(:method_name, :source, :beginning_line_number) do
+        attr_reader :state, :body_tokens
+
+        def initialize(*)
+          super
+          parse!
+        end
+
+        private
+
+        def parse!
+          @state = :initial
+
+          catch(:finish) do
+            source.tokens.each do |token|
+              invoke_state_handler(token)
+            end
+            raise TargetNotFoundError, 'Could not find the target block'
+          end
+        end
+
+        def finish!
+          throw :finish
+        end
+
+        def invoke_state_handler(token)
+          __send__("#{state}_state", token)
+        end
+
+        def initial_state(token)
+          @state = :after_method_call if token.location == block_locator.method_call_location
+        end
+
+        def after_method_call_state(token)
+          @state = :after_opener if handle_opener_token(token)
+        end
+
+        def after_opener_state(token)
+          if handle_closer_token(token)
+            finish_or_find_next_block_if_incorrect!
+          elsif pipe_token?(token)
+            finalize_pending_tokens!
+            @state = :after_beginning_of_args
+          else
+            pending_tokens << token
+            handle_opener_token(token)
+            @state = :after_beginning_of_body unless token.type == :on_sp
+          end
+        end
+
+        def after_beginning_of_args_state(token)
+          @state = :after_beginning_of_body if pipe_token?(token)
+        end
+
+        def after_beginning_of_body_state(token)
+          if handle_closer_token(token)
+            finish_or_find_next_block_if_incorrect!
+          else
+            pending_tokens << token
+            handle_opener_token(token)
+          end
+        end
+
+        def pending_tokens
+          @pending_tokens ||= []
+        end
+
+        def finalize_pending_tokens!
+          pending_tokens.freeze.tap do
+            @pending_tokens = nil
+          end
+        end
+
+        def finish_or_find_next_block_if_incorrect!
+          body_tokens = finalize_pending_tokens!
+
+          if correct_block?(body_tokens)
+            @body_tokens = body_tokens
+            finish!
+          else
+            @state = :after_method_call
+          end
+        end
+
+        def handle_opener_token(token)
+          opener_token?(token).tap do |boolean|
+            opener_token_stack.push(token) if boolean
+          end
+        end
+
+        def opener_token?(token)
+          token.type == :on_lbrace || (token.type == :on_kw && token.string == 'do')
+        end
+
+        def handle_closer_token(token)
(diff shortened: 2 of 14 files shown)


features/support/ruby_features.rb
```diff
@@ -27,3 +27,13 @@
...
+
+Around "@skip-when-ripper-unsupported" do |scenario, block|
+  require 'rspec/support/ruby_features'
+
+  if ::RSpec::Support::RubyFeatures.ripper_supported?
+    block.call
+  else
+    warn "Skipping scenario #{scenario.title} because Ripper is not supported"
+  end
+end
```
myronmarston: :+1: Nice and simple!

lib/rspec/expectations/block_snippet_extractor.rb
```diff
@@ -0,0 +1,241 @@
...
+      end
+
+    private
+
+      def raw_body_lines
+        raw_body_snippet.split("\n")
+      end
+
+      def raw_body_snippet
+        block_token_extractor.body_tokens.map(&:string).join('')
```
myronmarston: The `''` arg is unnecessary, I believe. `join` (no args) works just fine.

lib/rspec/expectations/block_snippet_extractor.rb
```diff
@@ -0,0 +1,241 @@
...
+      if RSpec.respond_to?(:world)
+        def source
+          RSpec.world.source_from_file(file_path)
+        end
+      else
+        RSpec::Support.require_rspec_support 'source'
+        def source
+          @source ||= RSpec::Support::Source.from_file(file_path)
+        end
+      end
```
myronmarston: :+1: Nice to see this leverage rspec-core if available.

lib/rspec/expectations/block_snippet_extractor.rb
```diff
@@ -0,0 +1,241 @@
...
+        private
+
+        def parse!
+          @state = :initial
+
+          catch(:finish) do
+            source.tokens.each do |token|
+              invoke_state_handler(token)
+            end
+            raise TargetNotFoundError, 'Could not find the target block'
```
myronmarston: I don't see anything that rescues this error--so does this mean the user will get this error instead of the `change` failure output they normally would get?  That seems problematic.

lib/rspec/expectations/block_snippet_extractor.rb
```diff
@@ -0,0 +1,241 @@
...
+          super
+          parse!
+        end
+
+        private
+
+        def parse!
+          @state = :initial
+
+          catch(:finish) do
```
myronmarston: It would be nice to see this implemented w/o `throw`/`catch`, but I can imagine that could be complicated, so if you considered that or tried it already let's stick with what you have here.

lib/rspec/expectations/block_snippet_extractor.rb
```diff
@@ -0,0 +1,241 @@
...
+
+        def opener_token_stack
+          @opener_token_stack ||= []
+        end
+
+        def pipe_token?(token)
+          token.op_type? && token.string == '|'
+        end
+
+        def correct_block?(body_tokens)
```
myronmarston: What is an "incorrect block"?

spec/rspec/expectations/block_snippet_extractor_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+            2.times do |index|
+              puts index
+            end
+          end
+        end
+
+        it 'does not confuse the inner `end` with the outer `end`' do
+          expect(body_content_lines).to eq([
+            '2.times do |index|',
+            'puts index',
```
myronmarston: Not sure if this matters or not, but I'm slightly surprised this line doesn't have two leading spaces for the indentation.  Is that important?

spec/rspec/expectations/block_snippet_extractor_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+        let(:expression) do
+          dummy_object.target_method { 2.negative? }; target_method { 1.positive? }
+        end
+
+        let(:dummy_object) do
+          double('dummy_object', :target_method => nil)
+        end
+
+        it 'raises AmbiguousTargetError' do
+          expect { body_content_lines }.to raise_error(BlockSnippetExtractor::AmbiguousTargetError)
```
myronmarston: Does this mean the user will get this error instead of a simpler failure message?  That would be bad, IMO.

spec/rspec/expectations/block_snippet_extractor_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+        let(:expression) do
+          target_method(another_method { 2.negative? }) { 1.positive? }
+        end
+
+        it 'correctly extracts the target snippet' do
+          expect(body_content_lines).to eq(['1.positive?'])
+        end
+      end
+
+      context "when wrong proc is given" do
```
myronmarston: What is a "wrong proc"?  I don't understand this example...

spec/rspec/expectations/block_snippet_extractor_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+          target_method(&block)
+        end
+
+        it 'raises TargetNotFoundError' do
+          expect { body_content_lines }.to raise_error(BlockSnippetExtractor::TargetNotFoundError)
+        end
+      end
+    end
+  end
+end
```
myronmarston: :+1: I love how thorough these spec are!

lib/rspec/expectations/block_snippet_extractor.rb
yujinakayama: It's rescued here https://github.com/rspec/rspec-expectations/pull/987/files#diff-16590f2e55b8f971bf68733232e5eca9R371.

lib/rspec/expectations/block_snippet_extractor.rb
yujinakayama: When parsing the following code:

```ruby
target_method(another_method { 2.negative? }) { 1.positive? }
#                                          ^
```

... BlockTokenExtractor iterates the following tokens:

```ruby
[
  [[1, 0], :on_ident, "target_method"],
  [[1, 13], :on_lparen, "("],
  [[1, 14], :on_ident, "another_method"],
  [[1, 28], :on_sp, " "],
  [[1, 29], :on_lbrace, "{"],
  [[1, 30], :on_sp, " "],
  [[1, 31], :on_int, "2"],
  [[1, 32], :on_period, "."],
  [[1, 33], :on_ident, "negative?"],
  [[1, 42], :on_sp, " "],
  [[1, 43], :on_rbrace, "}"], # <
  [[1, 44], :on_rparen, ")"],
  [[1, 45], :on_sp, " "],
  [[1, 46], :on_lbrace, "{"],
  [[1, 47], :on_sp, " "],
  [[1, 48], :on_int, "1"],
  [[1, 49], :on_period, "."],
  [[1, 50], :on_ident, "positive?"],
  [[1, 59], :on_sp, " "],
  [[1, 60], :on_rbrace, "}"]
]
```

Because there's no semantic information in tokens, BlockTokenExtractor does not know which block is the correct target. So, at the `^` point, BlockTokenExtractor asks BlockLocator (which utilizes semantic AST information) if the last parsed block (`{ 2.negative? }`) is the correct target or not. If it's incorrect, it continues further parsing.

spec/rspec/expectations/block_snippet_extractor_spec.rb
yujinakayama: Ideally it should properly handle indentations of multiline snippet, but for now we just use the result of BlockSnippetExtractor only when it's a single line. So I didn't implement the indentation handling since YAGNI.

spec/rspec/expectations/block_snippet_extractor_spec.rb
yujinakayama: I meant something "malicious", but the example is actually testing same behavior with the below example `context "when the block literal is described on different line with the method invocation"`. So I'll remove this example.

yujinakayama:
> On a side note: can this logic be applied to the `satisfy` matcher to provide a better description/failure message than "...satisfy block"?

Done 👍

spec/rspec/expectations/block_snippet_extractor_spec.rb
yujinakayama: Same as https://github.com/rspec/rspec-expectations/pull/987#discussion_r114675967.

yujinakayama:
Now this is ready for re-review.

lib/rspec/matchers/built_in/satisfy.rb
```diff
@@ -31,6 +33,28 @@ def failure_message
...
+          end
+
+          def extract_block_snippet
+            return nil unless @block
+            lines = Expectations::BlockSnippetExtractor.body_content_lines_of(@block, 'satisfy')
+            return nil unless lines.size == 1
+            "`#{lines.first}`"
+          rescue Expectations::BlockSnippetExtractor::Error
+            nil
+          end
```
myronmarston: This has a lot of duplicated knowledge between this method and the similar method in the `change` matcher.  Is there a way to extract something common?  Or is that premature?

spec/rspec/expectations/block_snippet_extractor_spec.rb
myronmarston: Can you document the fact that we don't deal with the indentation since we only ever use one line currently?  Ideally near that logic.

yujinakayama: Done 👍

lib/rspec/matchers/built_in/satisfy.rb
yujinakayama: Refactored them by adding `BlockSnippetExtractor.try_extracting_single_line_body_of`.

yujinakayama:
@myronmarston Thanks for the quick review 😺

PR #988, opened 2017-05-04
Support block snippet extraction with aliased matchers

yujinakayama:
Follow-up to https://github.com/rspec/rspec-expectations/pull/987.

The interface is not smart but we cannot pass matcher name to the constructor with the current matcher protocol.

### DIFF
--- a/lib/rspec/matchers/built_in/base_matcher.rb
+++ b/lib/rspec/matchers/built_in/base_matcher.rb
@@ -22,6 +22,9 @@ module RSpec
         # @private
         attr_reader :actual, :expected, :rescued_exception
 
+        # @private
+        attr_writer :matcher_name
+
         def initialize(expected=UNDEFINED)
           @expected = expected unless UNDEFINED.equal?(expected)
         end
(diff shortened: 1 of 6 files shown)


lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -96,6 +99,15 @@ def self.matcher_name
         end
 
         # @private
+        def matcher_name
+          if defined?(@matcher_name)
+            @matcher_name
+          else
+            self.class.matcher_name
```
myronmarston: Why does this fall back to the class method (and is there a `matcher_name` class method even defined?).

yujinakayama: Because `#matcher_name=` is set only when aliased (i.e. `@matcher_name` is nil when using matchers with their original name)..

The class method is defined [here](https://github.com/rspec/rspec-expectations/blob/41869a2abfbacede36391bd22f7257a5c840f7fd/lib/rspec/matchers/built_in/base_matcher.rb#L94-L96).

Issue #989, opened 2017-05-04
Improved `change` and `satisfy` failure messages not working


myronmarston:
``` ruby
RSpec.describe do
  example "`change` matcher" do
    a = b = 1

    expect {
      a += 1
      b += 2
    }.to change { a }.by(1).and change { b }.by(1)
  end

  example "`satisfy` matcher with symbol to_proc" do
    expect(2).to satisfy(&:odd?).and satisfy(&:positive?)
  end

  example "`satisfy` matcher with explicit blocks" do
    expect(2).to satisfy { |x| x.odd? }.and satisfy { |x| x.positive? }
  end
end
```

The recent changes to improve these failure messages are not working correctly.  Here's the output:

```
Failures:

  1) `satisfy` matcher with explicit blocks
     Failure/Error: expect(2).to satisfy { |x| x.odd? }.and satisfy { |x| x.positive? }
       expected 2 to satisfy block
     # /Users/myron/code/rspec-dev/repos/rspec-support/lib/rspec/support.rb:87:in `block in <module:Support>'
     # /Users/myron/code/rspec-dev/repos/rspec-support/lib/rspec/support.rb:96:in `notify_failure'
     # ./lib/rspec/expectations/fail_with.rb:35:in `fail_with'
     # ./lib/rspec/expectations/handler.rb:40:in `handle_failure'
     # ./lib/rspec/expectations/handler.rb:50:in `block in handle_matcher'
     # ./lib/rspec/expectations/handler.rb:27:in `with_matcher'
     # ./lib/rspec/expectations/handler.rb:48:in `handle_matcher'
     # ./lib/rspec/expectations/expectation_target.rb:65:in `to'
     # ./spec/foo_spec.rb:16:in `block (2 levels) in <top (required)>'
     # /Users/myron/code/rspec-dev/repos/rspec-core/lib/rspec/core/example.rb:254:in `instance_exec'
     # /Users/myron/code/rspec-dev/repos/rspec-core/lib/rspec/core/example.rb:254:in `block in run'
     # /Users/myron/code/rspec-dev/repos/rspec-core/lib/rspec/core/example.rb:500:in `block in with_around_and_singleton_context_hooks'
     # /Users/myron/code/rspec-dev/repos/rspec-core/lib/rspec/core/example.rb:457:in `block in with_around_example_hooks'
     # /Users/myron/code/rspec-dev/repos/rspec-core/lib/rspec/core/hooks.rb:464:in `block in run'
[...]

PR #990, opened 2017-05-04
Fix NoMethodError raised with &:symbol proc in #change and #satisfy

yujinakayama:
Ref: https://github.com/rspec/rspec-expectations/issues/989


myronmarston:
LGTM.

PR #999, opened 2017-07-06
Allow captured `$stdout/$stderr` to still return the actual`tty?` value.

reitermarkus:
This is useful for testing different behaviour with or without TTY.

### DIFF
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -129,13 +129,39 @@ module RSpec
       end
 
       # @private
+      class IOCapturer
+        def initialize(original)
+          @io = StringIO.new
+          @proxy = if defined?(::RSpec::Mocks)
+            ::RSpec::Mocks.space.proxy_for(original)
+          end
+        end
+
+        def respond_to?(method, *args)
+          if @proxy && @proxy.method_double_if_exists_for_message(method)
+            @proxy.object.respond_to?(method, *args)
+          else
+            @io.respond_to?(method, *args)
+          end
+        end
+
+        def method_missing(method, *args)
+          if @proxy && @proxy.method_double_if_exists_for_message(method)
+            @proxy.object.__send__(method, *args)
+          else
+            @io.__send__(method, *args)
+          end
+        end
+      end
+
+      # @private
       module CaptureStdout
         def self.name
           'stdout'
         end
 
         def self.capture(block)
-          captured_stream = StringIO.new
+          captured_stream = IOCapturer.new($stdout)
 
           original_stream = $stdout
           $stdout = captured_stream
@@ -155,7 +181,7 @@ module RSpec
         end
 
         def self.capture(block)
-          captured_stream = StringIO.new
+          captured_stream = IOCapturer.new($stderr)
 
           original_stream = $stderr
           $stderr = captured_stream
--- a/spec/rspec/matchers/built_in/output_spec.rb
+++ b/spec/rspec/matchers/built_in/output_spec.rb
@@ -144,6 +144,13 @@ module RSpec
           $stderr.print(msg)
         end
       }
+
+      it "forwards stubbed methods to the real $stderr" do
+        [true, false].each do |bool|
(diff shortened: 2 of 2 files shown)


myronmarston:
Thanks for your contribution!  I have a few thoughts on this:

* While I can see how this is useful, I'm worried that this may not be an appropriate default.  The IO object that is capturing the output is _not_ a TTY and it's a lie to make it claim that it is.  As such, I think it needs to be something that you opt-in to rather than the default behavior.  (I'm not yet sure what the opt-in API should be yet...).
* Your implementation loads and uses `forwardable` from the stdlib.  In RSpec we minimize how much of the stdlib we load and use.  If we load and use `forwardable`, it means a gem author can use `forwardable` but forget to `require` it, have passing specs (since RSpec loaded it) and release a broken gem.  So for this kind of thing we'd implement the delegation ourselves manually to avoid loading `forwadable`.

reitermarkus:
Is there an easy way to know if an object has been stubbed with `allow`, and know which methods have been stubbed? If so, the opt-in could be to only forward the methods that have been stubbed.

myronmarston:
If it's sufficient to let users opt-in to this by stubbing, we don't need to make any change...you could just do:

``` ruby
expect {
  allow($stdout).to receive(:tty?).and_return(true)
  do_something
}.to output(/something/).to_stdout
```

Does that work for you?

reitermarkus:
I think this should behave the same way, whether or not is is stubbed inside or outside of the expectation block. Also, this gets really redundant if you have multiple tests which rely on this. In my case about ten specs in the same file would have to be stubbed, instead of only once in a `before` block.

myronmarston:
> I think this should behave the same way, whether or not is is stubbed inside or outside of the expectation block.

I think that would be extremely implicit and surprising for users, and violates what users would expect from reading our docs:

* Our docs explain that stubbing an object affects only that object
* Our docs explain that the `output` matcher replaces `$stdout`/`$stderr` with a different object

Given that those IO objects get replaced with a new object -- there's no reason to expect from our docs that it carries stubs over.  And if we carried stubs over, it would be odd not to carry over message expectations for consistency (they are two sides of the same coin).

> Also, this gets really redundant if you have multiple tests which rely on this. In my case about ten specs in the same file would have to be stubbed, instead of only once in a before block.

You can create your own matcher or helper method that wraps `output` to provide this in the context of your project so you don't have to repeat it in many tests.

Now that I've thought about this some more, I think that this does not belong in RSpec.  I believe it's rarely needed, and would bear a high maintenance cost to the core team relative to the amount of use it will get.  We designed RSpec to be extensible (using custom matchers, etc) so that rarely needed things like this can be solved in individual projects (or in extension gems) rather than in RSpec itself.

I'm going to close this, but let us know if you want any help creating your own custom matcher or helper method or whatever to abstract this away for your projects.

reitermarkus:
I played around with this some more. Would something like this be acceptable to include upstream?

```ruby
expect { print "test" }.to output("test").to_stdout.as_tty
```

A custom matcher in this case would not be as nice to read, unless I'm missing something. The best I could come up with would be `output_to_stdout_as_tty("")`, which isn't really grammatically correct.

myronmarston:
If more users requested this feature, then yes, I think it would make sense to include `as_tty`.  But as it stands you're the only user who has asked, so I'm reticent to take on the maintenance burden of that.

Regarding a custom matcher, I was thinking you could simply wrap our built-in matcher using a decorator, and getting the exact same `.as_tty` syntax you are asking for.  Just include a module that overrides `output` and instantiates your decorator:

``` ruby
class OutputAsTTY < SimpleDelegator
  def matches?(block)
    if @as_tty
      # do what you need to do
    end

    super
  end

  def as_tty
    @as_tty = true
    self
  end

  module ExampleMethods
    def output(*args)
      core_matcher = super(*args)
      OutputAsTTY.new(core_matcher)
    end
  end
end

RSpec.configure do |rspec|
  rspec.include OutputAsTTY::ExampleMethods
end
```

reitermarkus:
Thank you very much, now I finally know what a `SimpleDelegator` is used for. 😄

PR #1005, opened 2017-07-25
Add receiver info to ChangeDetails#value_representation

joker1007:
This patch makes failure messages better if user uses compound matcher.

example

```ruby
expect { do_something }.to \
  change(ModelA, :count).from(0).to(1)
  .and(change(ModelB, :count).from(0).to(1))
  .and(change(ModelC, :count).from(0).to(1))
```

## previous message

```
          expected #count to have initially been 0, but was 2

       ...and:

          expected #count to have initially been 0, but was 1
```

This message is not user friendly because user cannot detect failed matchers.

## after patched

```
          expected `ModelB.count` to have initially been 0, but was 2

       ...and:

          expected `ModelC.count` to have initially been 0, but was 1
```

In this case, user can detect failed matchers easily

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -374,13 +374,13 @@ module RSpec
         def inspect_message_with_receiver(message, receiver)
           case receiver
           when Module
-            "#{receiver}.#{message}"
+            "`#{receiver}.#{message}`"
           when NilClass
-            "nil##{message}"
+            "`nil##{message}`"
           else
             singleton_class = class << receiver; self; end
             klass = singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
-            "#{klass}##{message}"
+            "`#{klass}##{message}`"
           end
         end
 
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -372,10 +372,15 @@ module RSpec
         end
 
         def inspect_message_with_receiver(message, receiver)
-          if receiver.class == Class
+          case receiver
+          when Module
             "#{receiver}.#{message}"
+          when NilClass
+            "nil##{message}"
           else
-            "#{receiver.class}##{message}"
+            singleton_class = class << receiver; self; end
+            klass = singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
+            "#{klass}##{message}"
           end
         end
 
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -375,12 +375,8 @@ module RSpec
           case receiver
           when Module
             "`#{receiver}.#{message}`"
-          when NilClass
-            "`nil##{message}`"
           else
-            singleton_class = class << receiver; self; end
-            klass = singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
-            "`#{klass}##{message}`"
+            "`#{Support.class_of(receiver)}##{message}`"
           end
         end
 
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -334,7 +334,7 @@ module RSpec
         def value_representation
           @value_representation ||=
             if @message
-              "##{@message}"
+              inspect_message_with_receiver(@message, @receiver)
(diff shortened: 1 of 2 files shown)


yujinakayama:
Thank you for the improvement! Could you add/update failing specs accordingly?

yujinakayama:
By the way, on RSpec 3.6, you can benefit from the block snippet extraction by writing the change matchers as `change { ModelA.count }`:

http://rspec.info/blog/2017/05/rspec-3-6-has-been-released/#expectations-improved-failure-messages-for-the-change-and-satisfy-matchers

lib/rspec/matchers/built_in/change.rb
```diff
@@ -371,6 +371,14 @@ def evaluate_value_proc
           end
         end
 
+        def inspect_message_with_receiver(message, receiver)
+          if receiver.class == Class
```
yujinakayama: In RSpec codebase we prefer `Class === receiver` over `receiver.class == Class`, since the `receiver`, which is given by users, might be an object that does not respond to `#class` like BasicObject.

lib/rspec/matchers/built_in/change.rb
```diff
@@ -371,6 +371,14 @@ def evaluate_value_proc
...
+            "#{receiver}.#{message}"
+          else
+            "#{receiver.class}##{message}"
```
yujinakayama: We also need to handle objects that don't respond to `#class` here like the following:

```ruby
singleton_class = class << receiver; self; end
singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
```

https://github.com/rspec/rspec-support/blob/70a8b33fb194f5014e37d99dc3d9acca88ab8647/lib/rspec/support/object_formatter.rb#L202-L203

lib/rspec/matchers/built_in/change.rb
yujinakayama: Moreover, we should check if it's `Module` or not rather than `Class` so that we can handle both classes and modules 💁

joker1007:
Thank you for review comments.
I will fix soon.

joker1007:
@yujinakayama 
I have fixed points that you commented.

>> By the way, on RSpec 3.6, you can benefit from the block snippet extraction by writing the change matchers as change { ModelA.count }:

Thanks. I know this feature. But we already have so many `change(receiver, message)` style test codes. I prefer to improve change matcher.

lib/rspec/matchers/built_in/change.rb
```diff
@@ -372,10 +372,15 @@ def evaluate_value_proc
         end
 
         def inspect_message_with_receiver(message, receiver)
-          if receiver.class == Class
+          case receiver
+          when Module
             "#{receiver}.#{message}"
+          when NilClass
+            "nil##{message}"
```
yujinakayama: I think the special notation `nil#some_message` is a bit odd. Isn't `NilClass#some_message` good enough?

If you did this to handle the following issue in the class extraction:

```ruby
[1] pry(main)> receiver = nil
=> nil
[2] pry(main)> singleton_class = class << receiver; self; end
=> NilClass
[3] pry(main)> klass = singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
=> Object # This is wrong
```

... we can simplify it as follows:

```ruby
def inspect_message_with_receiver(message, receiver)
  case receiver
  when Module
    "#{receiver}.#{message}"
  else
    "#{class_of(receiver)}##{message}"
  end
end

def class_of(object)
  object.class
rescue NoMethodError
  singleton_class = class << object; self; end
  singleton_class.ancestors.find { |ancestor| !ancestor.equal?(singleton_class) }
end
```

joker1007: Thanks!!
I will use `Support.class_of` after merge https://github.com/rspec/rspec-support/issues/325.

yujinakayama: Thank you for waiting. https://github.com/rspec/rspec-support/pull/325 is now merged and `Support.class_of` is available 👌

myronmarston:
@yujinakayama / @joker1007 what's the status of this PR?  I might start working on a fix for #1030 which would probably conflict with the changes here so it'd be good to get this merged first.

yujinakayama:
I've just added a commit and this PR should be ready to be merged.

joker1007:
Thanks for your support! @yujinakayama 
It looks good to me!

yujinakayama:
@joker1007 Thanks for the improvement!

PR #1008, opened 2017-08-12
Improve predicate matcher message for `be_something?`

pocke:
I sometimes write `expect(actual).to be_something?` by mistake. Actually, `expect(actual).to be_something` is correct (without the question suffix).

Currently, RSpec tells me about this mistake, but the message is not easy to understand.

```
expected actual to respond to `something?`
```

By this change, RSpec tells me about this mistake with more descriptive message.

```
expected actual to respond to `something?` or perhaps you meant `be_something`(without question suffix)
```

### DIFF
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -278,6 +278,8 @@ module RSpec
             msg << " or perhaps you meant `be true` or `be_truthy`"
           elsif predicate == :false?
             msg << " or perhaps you meant `be false` or `be_falsey`"
+          elsif predicate.to_s.end_with?('??')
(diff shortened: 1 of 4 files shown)


yujinakayama:
The error message already indicates the mistake by the `??`:

```ruby
RSpec.describe 'dynamic predicate matchers' do
  example do
    expect(2).to be_even?
  end
end
```

```
Failures:

  1) dynamic predicate matchers should be even?
     Failure/Error: expect(2).to be_even?
       expected 2 to respond to `even??`
```

Actually I'm not sure if this patch worth maintaining for the particular mistake.

myronmarston:
I agree with @yujinakayama.  Closing.

PR #1009, opened 2017-08-12
Raise warning when matching raise_error(StandardError) without a message

mariokostelac:
Same as raise_error, raise_error(StandardError) is prone to catch
NoMethodError, NameError, ArgumentError, mistakenly allowing tests to
pass without even executing wanted codepath.

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -12,7 +12,7 @@ module RSpec
         def initialize(expected_error_or_message=nil, expected_message=nil, &block)
           @block = block
           @actual_error = nil
-          @warn_about_bare_error = expected_error_or_message.nil?
+          @warn_about_bare_error = expected_error_or_message.nil? || expected_error_or_message == StandardError && expected_message.nil?
 
           case expected_error_or_message
           when nil
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -13,6 +13,16 @@ RSpec.describe "expect { ... }.to raise_error" do
     expect { raise }.to raise_error
   end
 
+  it "issues a warning when used with StandardError without a message" do
+    expect_warning_with_call_site __FILE__, __LINE__+1, /without providing a specific error/
+    expect { raise StandardError.new }.to raise_error(StandardError)
+  end
+
+  it "does not issue a warning when used with StandardError with a message" do
+    expect_no_warnings
+    expect { raise StandardError.new("message") }.to raise_error(StandardError, "message")
+  end
+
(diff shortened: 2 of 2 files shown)


mariokostelac:
I see that some tests from mocks projects are failing, but interested to hear if this does not make sense. I've recently bumped into this problem (catching NoMethodError by mistake) and was surprised that rspec does not protect me from that mistake.

xaviershay:
Thank you! Tests only appear to be failing because we fail on warnings and this generates a warning :)

I _think_ this probably makes sense, though I can't remember our policy for introducing new warnings. (i.e. ok to merge straight away or wait until next minor). If no one else chips in in the next couple of days I'll make an executive decision and probably just recommend merge (... after fixing mocks)

myronmarston:
I lean against this change.  Bare `raise_error()` is the path of lease resistance, and people tend to use it as a first step, or when they haven't really thought about what the error should be.  Once they've passed some kind of argument to `raise_error`, they've made a decision about what error they want to specify, and I don't think we should nag them with a warning.

Consider also some kind like `raise StandardError` (with no message).  While it's a bad idea to raise a `StandardError` with no message, for every bit of valid ruby code I think there needs to be a way to specify it with RSpec w/o getting warnings from RSpec, and `raise_error(StandardError)` would be the normal way to do that.

Thoughts?

mariokostelac:
> While it's a bad idea to raise a StandardError with no message, for every bit of valid ruby code I think there needs to be a way to specify it with RSpec w/o getting warnings from RSpec, and raise_error(StandardError) would be the normal way to do that.

Definitely interesting point! I am pretty new in this, but I guess the decision whether to merge it or not depends on what are warnings used for in RSpec. I think about warnings (with raising) as a strict mode in some other languages. It protects you from doing dangerous things. Of course, you can always turn it off if you think you know better than the strict mode, or really need to do something that strict mode does not allow.
Maybe we should allow turning off this specific warning (just for standard error), but have it turned on by default?

mikegee:
I like the idea here. But since there is some debate, perhaps this fits better in rubocop-rspec (if it doesn't already enforce this).

mariokostelac:
The more I read the following sentence, the more I think it should be done by rspec but rubocop itself:
> Consider also some kind like raise StandardError (with no message). While it's a bad idea to raise a StandardError with no message, for every bit of valid ruby code I think there needs to be a way to specify it with RSpec w/o getting warnings from RSpec, and raise_error(StandardError) would be the normal way to do that.

I will take a look rubocop rules. Feel free to close this one if you find @myronmarston's sentence as strong as I do.

mariokostelac:
@mikegee thanks for pointing out to the rubocop-rspec project.

myronmarston:
I think rubocop-rspec is a better place for this, too.  It can be much more opinionated than RSpec itself can :).

I'm going to close this since you're going that direction.

Issue #978, opened 2017-03-13
Include matcher does not utilize duck typing


KevinSjoberg:
Consider the following code:

```ruby
class Foo < SimpleDelegator
end
```

Now, a test:

```ruby
RSpec.describe Foo do
  it "includes bar" do
    foo = described_class.new({bar: "baz"})
    expect(foo).to include(bar: "baz")
  end
end
```

The following code will fail when it really should not. The culprit can be found in [#comparing_hash_to_a_subset?](https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/include.rb#L103-L105). It currently just checks if `actual` is a `Hash`.

If we instead check if `actual` responds to `#fetch`, `#key?` and `#keys`, we can now properly test hash-like objects as well. If we allow hash-like objects, we need to expect that hash-like objects are being passed in an expectation as well. Is `#fetch`, `#key?` and `#keys` enough of an interface?

I'd be happy to take a stab at this. Before I do, it would be nice with some feedback.

Cheers

myronmarston:
I can see how nice it would be for the matcher to treat anything hash-like like a hash, but there are a number of "is a hash?" checks outside the method you identified:

* https://github.com/rspec/rspec-expectations/blob/82ab90b513dfcd0b52064f56be4b2c6218adeded/lib/rspec/matchers/built_in/include.rb#L54
* https://github.com/rspec/rspec-expectations/blob/82ab90b513dfcd0b52064f56be4b2c6218adeded/lib/rspec/matchers/built_in/include.rb#L73
* https://github.com/rspec/rspec-expectations/blob/82ab90b513dfcd0b52064f56be4b2c6218adeded/lib/rspec/matchers/built_in/include.rb#L113

Seems like we'd have to apply duck-typing to all of them. (Although we might be able to get by only applying that to the ones that check actual values: it seems less useful to support hash-like objects passed to the matcher itself as expected values).  In addition, we do similar checks for `Array` and `String`:

https://github.com/rspec/rspec-expectations/blob/82ab90b513dfcd0b52064f56be4b2c6218adeded/lib/rspec/matchers/built_in/include.rb#L133

...and if we're going to support duck typing for hashes, we should probably support it for those, too, for consistency.

> Is `#fetch?`, `#key?` and `#keys?` enough of an interface?

Should be `fetch` (not `fetch?`) and `keys` (not `keys?`).  And it looks like we use `merge`, as well.

[...]

PR #1012, opened 2017-08-22
Duck-typing support for the `include` matcher

pablobm:
Recently I was refactoring some code where a class inherited from `Hash`, like follows:

```
class PseudoHash < Hash
  # ...
end
```

This `PseudoHash` class was pretty much a hash (with some additional functionality of course) and passed for one where required. However, inheriting from core classes is discouraged, so I changed that inheritance to something else.

The specs for this class used the `include` matcher. When I made the change, these specs started failing because the matcher internally checks for `#is_a?(Hash)` in order to run the hash subset matching. My fix was just to use `#to_hash` in each failing spec and leave it at that.

However, this left me wondering whether this is something that RSpec should be doing internally: my type quacked and walked like a hash, so why would RSpec not treat it like a hash? Ruby puts an emphasis on duck typing, and asking if an object is of a specific class or not is generally considered a smell.

Thoughts?


lib/rspec/matchers/built_in/include.rb
```diff
@@ -15,12 +15,14 @@ def initialize(*expecteds)
         # @api private
         # @return [Boolean]
         def matches?(actual)
+          actual = actual.to_hash if actual.respond_to?(:to_hash)
```
myronmarston: Can you use `Hash.try_convert` instead?  Seems purpose-built for when you want to try to convert an object to a hash...

features/built_in_matchers/include.feature
```diff
@@ -142,3 +142,58 @@ Feature: `include` matcher
...
+      | expected {:a => 7, :b => 5} to include :c and :d              |
+      | expected {:a => 7, :b => 5} to include {:d => 2}              |
+      | expected {:a => 7, :b => 5} to include {:a => 5}              |
+      | expected {:a => 7, :b => 5} to include {:a => 5, :b => 7}     |
+      | expected {:a => 7, :b => 5} to include :d                     |
+      | expected {:a => 7, :b => 5} not to include :a                 |
+      | expected {:a => 7, :b => 5} to include {:d => 3}              |
+      | expected {:a => 7, :b => 5} not to include {:a => 7}          |
+
+
```
myronmarston: IMO, this isn't worth elevating to such a prominent place in our relish docs.  We don't show every last use case of the matchers in our relish docs, and there's a definite maintenance cost to adding to our cucumber suite.  Do you mind removing this scenario?

lib/rspec/matchers/built_in/include.rb
```diff
@@ -87,6 +89,7 @@ def excluded_from_actual
           return [] unless @actual.respond_to?(:include?)
 
           expecteds.inject([]) do |memo, expected_item|
+            expected_item = expected_item.to_hash if expected_item.respond_to?(:to_hash)
```
franzliedke: Nice PR!

The discussion in #978 concluded that the `expecteds` did not need to be converted to a hash. Is this necessary?

lib/rspec/matchers/built_in/include.rb
pablobm: I don't follow. `Hash.try_convert` returns `nil` on failure, which we don't want here, as `include` does accept things other than hashes.

Unless you mean to do the following?

```ruby
actual = Hash.try_convert(actual) || actual
```

features/built_in_matchers/include.feature
pablobm: Agreed. Removed now.

lib/rspec/matchers/built_in/include.rb
pablobm: Aha, I had missed that conversation. I agree with the conclusion, so I have removed this conversion now.

pablobm:
@myronmarston No worries at all. Thank you for looking into it, and my apologies for not digging a bit more, missing https://github.com/rspec/rspec-expectations/issues/978

Happy to tackle any other concerns.

lib/rspec/matchers/built_in/include.rb
myronmarston: > Unless you mean to do the following?

Yep, that's what I had in mind: `actual = Hash.try_convert(actual) || actual`.

pablobm: I made the change as advised, but I'm on the fence as to whether it is an improvement or not. It does look a bit neater (to me), but I don't know if it's as understandable. I mean this in the sense that I have no idea what people would make of either version of the code, what idiom is best known (if known at all) by Ruby devs at large.

I used `#to_hash` directly on my initial version following Avdi Grimm's advice on his book [_Confident Ruby_](http://www.confidentruby.com)[1]. My thinking is that perhaps this is better known and understood because I happen to have read a resource where it was mentioned. I never heard of `.try_convert`, but it might be that it's better known and I just happen not to have read resources on it.

In any case, from Ruby's source code (and quick test), I understand it's the same thing, internally using `#to_hash` where available, so it shouldn't matter functionally whether we use one or the other.

[1] Particularly sections 3.2 _Use built-in conversion protocols_ and 3.3 _Conditionally call conversion methods_.

myronmarston: Thanks.  I don't disagree with your reasoning, but weigh other factors.  The main one that makes me think `try_convert` is worth is that it's the official conversion method Ruby has provided.  In the future, it might do more than check `to_hash`.  IMO, it makes sense to use an official conversion method instead of doing it yourself.

myronmarston:
Thanks for making that change, @pablobm.  Unfortunately, it looks like `try_convert` isn't available on 1.8.7.  Given we support 1.8.7 I think we're going to have to revert back to your `to_hash` approach.  Sorry to make you switch and then switch back!  If you just remove the commit from your branch (e.g. `git reset --hard HEAD~1`) and force push, the prior commit had a green build so I'll be able to merge w/o waiting for a full build.

Thanks!

pablobm:
No worries; I think your reasoning is pretty good and I'll bear it in mind in my future work. Too bad it wasn't possible this time. I have pushed the shortened branch.

myronmarston:
Thanks @pablobm!

isqad:
It is bad PR IMO, becouse it breaks those examples:

```
expect({'a' => 1}.with_indifferent_access).to include(:a)
```

PR #1033, opened 2017-12-28
Upgrade cucumber to 3.0

Dorian:
Prevent Fixnum warnings from old version of gherkin

### DIFF
--- a/cucumber.yml
+++ b/cucumber.yml
@@ -1,10 +1,10 @@
 <%
 def tags(tag)
-  tags = [tag]
-  tags << "~@ruby-1.9" if RUBY_VERSION.to_f < 1.9
+  tags = ["'#{tag}'"]
+  tags << "'not @ruby-1.9'" if RUBY_VERSION.to_f < 1.9
   tags.join(" --tags ")
 end
 %>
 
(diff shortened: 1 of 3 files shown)


myronmarston:
Thanks for taking a stab at this! Unfortunately, cucumber 3.0 does not work on some of the ruby versions that we still support (as you can see by the build failure), so we can't upgrade until we drop support for some of the older Ruby versions (which we plan to do in RSpec 4.0 at some future point). For now, we are going to stick with cucumber 1.x.

Feel free to re-submit this once we've dropped support for the old versions :).

JonRowe:
@myronmarston could we do a ruby version sensitive gemfile here? And allow us to have the right dependencies on the right versions?

myronmarston:
> @myronmarston could we do a ruby version sensitive gemfile here? And allow us to have the right dependencies on the right versions?

Maybe, but what is the advantage?  Having to maintain a cuke suite that runs against different versions of cucumber (on top of running against different versions of Ruby) sounds like a maintenance nightmare.  IMO, there's no significant advantage to upgrading cucumber until we can use it for everything.  After all, the benefit to upgrading is to unlock new features, but we can't use new features and get that benefit if we still have to support old versions.

PR #1034, opened 2017-12-29
Use `hash` for `change` matcher to handle deeply nested objects.

### LINKED ISSUE #1030: 'change' matcher using incorrect initial value (with Vectors nested in arrays)
When using vectors nested inside a hash inside an array the change matcher appears to take the resulting value as the initial value.

For example;

```ruby
require 'matrix'

RSpec.describe 'vectors' do
  let(:vectors) { [{ a: Vector[1, 2] }] }

  it 'adds to the vector' do
    expect { vectors[0][:a] += Vector[1, 1] }
      .to change{vectors}
      .from( [{a: Vector[1, 2] }] )
      .to( [{a: Vector[2, 3]}] )
  end
end
```

When I run this test I get:

```
Failures:

  1) vectors adds to the vector
     Failure/Error:
       expect { vectors[0][:a] += Vector[1, 1] }
         .to change{vectors}
         .from( [{a: Vector[1, 2] }] )
         .to( [{a: Vector[2, 3]}] )

       expected `vectors` to have initially been [{:a=>Vector[1, 2]}], but was [{:a=>Vector[2, 3]}]
     # ./spec/test_spec.rb:7:in `block (2 levels) in <top (required)>'

Finished in 0.02426 seconds (files took 0.12668 seconds to load)
1 example, 1 failure
```

If I remove the `from` section I then get a failure saying that `vectors` did not change.

I am using rspec 3.7.0 and Ruby 2.4.1

myronmarston:
Fixes #1030.

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -43,16 +43,13 @@ module RSpec
 
         # @private
         def matches?(event_proc)
-          @event_proc = event_proc
-          return false unless Proc === event_proc
           raise_block_syntax_error if block_given?
-          change_details.perform_change(event_proc)
-          change_details.changed?
+          perform_change(event_proc) && change_details.changed?
         end
 
         def does_not_match?(event_proc)
           raise_block_syntax_error if block_given?
-          !matches?(event_proc) && Proc === event_proc
+          perform_change(event_proc) && !change_details.changed?
         end
 
         # @api private
@@ -92,6 +89,18 @@ module RSpec
           @change_details ||= ChangeDetails.new(matcher_name, @receiver, @message, &@block)
         end
 
+        def perform_change(event_proc)
+          @event_proc = event_proc
+          change_details.perform_change(event_proc) do |actual_before|
+            # pre-compute values derived from the `before` value before the
+            # mutation is applied, in case the specified mutation is mutation
+            # of a single object (rather than a changing what object a method
+            # returns). We need to cache these values before the `before` value
+            # they are based on potentially gets mutated.
+            @actual_before_description = description_of(actual_before)
+          end
+        end
+
         def raise_block_syntax_error
           raise SyntaxError, "Block not received by the `change` matcher. " \
           "Perhaps you want to use `{ ... }` instead of do/end?"
@@ -99,12 +108,12 @@ module RSpec
 
         def positive_failure_reason
           return "was not given a block" unless Proc === @event_proc
-          "is still #{description_of change_details.actual_before}"
+          "is still #{@actual_before_description}"
         end
 
         def negative_failure_reason
           return "was not given a block" unless Proc === @event_proc
-          "did change from #{description_of change_details.actual_before} " \
+          "did change from #{@actual_before_description} " \
           "to #{description_of change_details.actual_after}"
         end
       end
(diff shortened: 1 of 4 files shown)


myronmarston:
This is ready for review /cc @rspec/rspec.

The travis failure is only for 1.9.3, and is related to the rspec-rails build, due to a rake version mismatch.  I've attempted a fix in rspec/rspec-rails#1935.

myronmarston:
Not sure why appveyor isn't building, but it was green before I rebased against master, and now travis is green, too.  Anyone from @rspec/rspec up for reviewing this?

lib/rspec/matchers/built_in/change.rb
```diff
@@ -312,8 +322,20 @@ def change_description
...
+      #
+      # Note that this class exposes the `actual_after` value, to allow the
+      # matchers above to derive failure messages, etc from the value on demand
+      # as needed, but it intentionally does _not_ expose the `actual_before`
+      # value. Some usages of the `change` matcher mutate a specific object
+      # returned by the value proc, which means that failure message snippets,
+      # etc, which are derived from the `before` value may not be accurate if
+      # they are lazily computed as needed. We must pre-compute them before
+      # applying the change in the `expect` block. To ensure that all `change`
+      # matchers do that properly, we do not expeose the `actual_before` value.
```
xaviershay: nit: typo on expeose

lib/rspec/matchers/built_in/change.rb
```diff
@@ -344,12 +366,29 @@ def value_representation
...
-          @actual_before != @actual_after
+          # Consider it changed if either:
+          #
+          # - The before/after values are unequal
+          # - The before/after values have different hash values
+          #
+          # The latter case specifically handles the case when the value proc
+          # returns the exact same object, but it has been mutated.
+          #
+          # Note that is is not sufficient to only check the hashes; it is
```
xaviershay: nit "is is" typo

lib/rspec/matchers/built_in/change.rb
```diff
@@ -359,16 +398,7 @@ def actual_delta
       private
 
         def evaluate_value_proc
-          value_proc = @value_proc || lambda { @receiver.__send__(@message) }
-
-          case val = value_proc.call
```
xaviershay: glad we could get rid of this

lib/rspec/matchers/built_in/change.rb
```diff
@@ -344,12 +366,29 @@ def value_representation
 
         def perform_change(event_proc)
           @actual_before = evaluate_value_proc
+          @before_hash = @actual_before.hash
+          yield @actual_before if block_given?
+
+          return false unless Proc === event_proc
```
xaviershay: not obvious to me why this was added/needed? When is a non-`Proc` passed?

xaviershay: oh, it might be `nil` with some invocations, though in that case why not check for that directly?

myronmarston: It an be anything, if a user tries to use it as a value matcher instead of a block matcher, e.g.:

``` ruby
expect(3).to change { ... }
expect(:foo).to change { ... }
```

In these cases, `event_proc` will be `3` and `:foo`, respectively.

Anyhow, this check wasn't actually added; it was moved as part of the diff.

lib/rspec/matchers/built_in/change.rb
myronmarston: Me too.  The dup'ing always felt dirty :(.

lib/rspec/matchers/built_in/change.rb
myronmarston: Good catch, fixed.

lib/rspec/matchers/built_in/change.rb
myronmarston: Good catch, fixed.

myronmarston:
Thanks for the review, @xaviershay.  I pushed fixes for the typos and will merge once the build finishes.

PR #1037, opened 2017-12-31
Disable Ruby 2.5 feature to avoid test failures on 2.5.

myronmarston:


xaviershay:
LGTM

spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -216,6 +216,19 @@ def expect_error_included_in_aggregated_failure(error)
...
+      # On Ruby 2.5+, the new `report_on_exception` causes the errors in the threads
+      # to print warnings, which our rspec-support test harness converts into a test
+      # failure since we want to enforce warnings-free code. To prevent the warning,
+      # we need to disable the setting here.
+      if Thread.respond_to?(:report_on_exception)
+        around do |example|
+          orig = Thread.report_on_exception
+          Thread.report_on_exception = false
+          example.run
+          Thread.report_on_exception = orig
```
xaviershay: I double checked that `example.run` won't raise if the spec fails (we don't actually say this explicitly in [our docs](https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks), though it's implied by context)

myronmarston: Sounds like something we should improve in our docs :).

xaviershay: IN ACTUAL FACT I just have poor reading comprehension: https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception

PR #1038, opened 2017-12-31
Update yard to 0.9.12, motivated by CVE.

xaviershay:
* Added @private tag for nested class (no longer inherits).
* Included full method definition in generated should/should_not docs.
* Used const_set for constant redefinition, since using = confuses YARD.
* Added short descriptions to methods that are @api private. Could
  probably just @private the lot but didn't want to make that decision
  here.

### DIFF
--- a/lib/rspec/expectations/minitest_integration.rb
+++ b/lib/rspec/expectations/minitest_integration.rb
@@ -53,6 +53,6 @@ module RSpec
   module Expectations
     remove_const :ExpectationNotMetError
     # Exception raised when an expectation fails.
-    ExpectationNotMetError = ::Minitest::Assertion
+    const_set :ExpectationNotMetError, ::Minitest::Assertion
   end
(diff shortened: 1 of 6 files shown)


lib/rspec/expectations/minitest_integration.rb
```diff
@@ -53,6 +53,6 @@ module RSpec
   module Expectations
     remove_const :ExpectationNotMetError
     # Exception raised when an expectation fails.
-    ExpectationNotMetError = ::Minitest::Assertion
+    const_set('ExpectationNotMetError', ::Minitest::Assertion)
```
myronmarston: Can this use a symbol instead?  After all, we used on for `remove_const`, and a symbol won't allocate a new object like a string will.

xaviershay: it can!

PR #1048, opened 2018-03-02
Don't use RSpec.configure to use the matcher DSL

### LINKED ISSUE #122: Matcher DSL isn't available inside example groups if rspec-expectations is loaded before rspec-core.
Not sure if this is the right place to report this, but it seemed the most sensible.

The rspec-expectations matcher DSL checks at load time whether `RSpec.configuration` is defined, and only includes itself if it is:

https://github.com/rspec/rspec-expectations/blob/a81b0678bd504c10da5475d73a676f6f60c1e9d7/lib/rspec/matchers/dsl.rb#L97

This causes problems if, e.g., you have a Gemfile like this:

```ruby
source "https://rubygems.org"
gem "rspec-rails"
gem "rspec-collection_matchers"
```

When rspec-collection_matchers is loaded, it loads rspec-expectations, but rspec-core has not yet been loaded. This means that the matcher DSL will not be available in example groups, and so scoped aliased and negated matchers cannot be used without manually extending `RSpec::Matchers::DSL`.

alyssais:
rspec-expectations shouldn't have knowledge of how to hook itself into rspec-core. rspec-core already includes the `RSpec::Matchers` module in example groups, so just including the module there (in addition to extending it) is enough to provide it to rspec-core, even if rspec-core hasn't been loaded yet and is loaded after rspec-expectations.

Fixes https://github.com/rspec/rspec/issues/122.

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -240,6 +240,7 @@ module RSpec
   #
   module Matchers
     extend ::RSpec::Matchers::DSL
+    include ::RSpec::Matchers::DSL
 
     # @!macro [attach] alias_matcher
     #   @!parse
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -94,8 +94,6 @@ module RSpec
         # :nocov:
       end
 
-      RSpec.configure { |c| c.extend self } if RSpec.respond_to?(:configure)
-
       # Contains the methods that are available from within the
       # `RSpec::Matchers.define` DSL for creating custom matchers.
       module Macros


myronmarston:
As the failing build indicates, this is not quite sufficient.  In rspec-core, we `include RSpec::Matchers` but we do not extend it (nor should we: that would add a ton of methods to a context where we do not want those methods added!).

I think the only solution here is to update rspec-core to `extend RSpec::Matchers::DSL`.  This can be done in the `config.expect_with` method, as part of [the branch for rspec](https://github.com/rspec/rspec-core/blob/6fdcbfec431a918afedeabcae17532fb0b1b2c2d/lib/rspec/core/configuration.rb#L795-L803).

Once that's been done, I think we can remove the `extend` call being done in rspec-expectations and everything should continue working.

pirj:
Are you still up to fix this @alyssais ?

JonRowe:
I'm closing this due to inactivity during the monorepo migration, the issue will be transferred

PR #1052, opened 2018-03-18
Pin ffi on RBX / 1.8

JonRowe:
FFI changed their minimum supported ruby version, so we'll have to pin it.

See https://travis-ci.org/rspec/rspec-expectations/jobs/354893396


myronmarston:
Why do we depend upon ffi?  I didn't think we did any native stuff...

JonRowe:
AFAIK it's a downstream dependency, I know that both our windows builds and RBX require it.

myronmarston:
LGTM

acdha:
Just wanted to thank you for actually documenting this in a searchable manner — I had to fix an old RHEL / Centos 6 build script and this issue was easier to find than spelunking through the main ffi issue tracker.

PR #1056, opened 2018-04-25
Fix tests for 951 + rdoc first word needs to match the variable

benoittgt:
Tests were leaking config changes.

It should fix #951

### DIFF
--- a/lib/rspec/expectations/configuration.rb
+++ b/lib/rspec/expectations/configuration.rb
@@ -56,6 +56,19 @@ module RSpec
         end
       end
 
+      # Configures the maximum character length that RSpec will print while
+      # formatting an object
+      # @param [Fixnum] the length to format to
+      # @example
+      #   RSpec.configure do |rspec|
+      #     rspec.expect_with :rspec do |c|
+      #       c.max_formatted_output_length = 200
+      #     end
+      #   end
+      def max_formatted_output_length=(length)
+        RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = length
+      end
+
       # The list of configured syntaxes.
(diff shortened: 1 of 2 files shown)


lib/rspec/expectations/configuration.rb
```diff
@@ -56,6 +56,19 @@ def syntax=(values)
         end
       end
 
+      # Configures the maximum character length that RSpec will print while
+      # formatting an object
+      # @param [Fixnum] the length to format to
```
JonRowe: @benoittgt this rdoc is wrong too :/ It needs to be `@param [Fixnum] length the number of characters to limit the formatted output to` (As the first word needs to match the variable).

benoittgt: Today I learn this. Thanks

benoittgt:
Should we add changelog entry?

JonRowe:
Thanks @benoittgt !

lib/rspec/expectations/configuration.rb
```diff
@@ -56,6 +56,19 @@ def syntax=(values)
...
 
+      # Configures the maximum character length that RSpec will print while
+      # formatting an object
+      # @param [Fixnum] length the number of characters to limit the formatted output to
+      # @example
+      #   RSpec.configure do |rspec|
+      #     rspec.expect_with :rspec do |c|
+      #       c.max_formatted_output_length = 200
+      #     end
+      #   end
```
myronmarston: I think the docs should mention that you can set the length to `nil` to prevent RSpec from doing any truncation.

benoittgt: Good idea @myronmarston. https://github.com/rspec/rspec-expectations/pull/1057

PR #1057, opened 2018-04-25
Add doc to inform that max_formatted_output_length can be set to nil

benoittgt:
See : https://github.com/rspec/rspec-expectations/pull/1056#discussion_r184113835


myronmarston:
Thanks @benoittgt!

PR #1060, opened 2018-05-22
Add documentation about testing array equality IN ORDER

ianfixes:
After a brief and fruitless search, it was unclear to me how to verify that an `actual` array in rspec would have not only the same elements as the `expected` array, but have them in the same order.

Notably, this concerning line:

```ruby
expect([1, 2, 3]).to     match_array([3, 2, 1])
```

This documentation fix reflects the fact that `eq` seems to properly compare array ordering and membership.  Should it also be referenced in the section on `eq`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,30 +175,41 @@ expect(1..10).to cover(3)
 ### Collection membership
 
 ```ruby
-expect(actual).to include(expected)
+# exact order, entire range
+expect(actual).to eq(expected)
+
+# exact order, partial range (based on an exact position)
 expect(actual).to start_with(expected)
 expect(actual).to end_with(expected)
 
-expect(actual).to contain_exactly(individual, items)
-# ...which is the same as:
-expect(actual).to match_array(expected_array)
+# any order, entire range
+expect(actual).to match_array(expected)
+# Assuming that:
+#   expected = [expected_element1, expected_element2]
+# You can also express this by passing the expected elements
+# as individual arguments (i.e. a variadic function or splat):
+expect(actual).to contain_exactly(expected_element1, expected_element2)
+
+# any order, partial range
+expect(actual).to include(expected)
 ```
 
 #### Examples
 
 ```ruby
-expect([1, 2, 3]).to include(1)
-expect([1, 2, 3]).to include(1, 2)
-expect([1, 2, 3]).to start_with(1)
-expect([1, 2, 3]).to start_with(1, 2)
-expect([1, 2, 3]).to end_with(3)
-expect([1, 2, 3]).to end_with(2, 3)
-expect({:a => 'b'}).to include(:a => 'b')
-expect("this string").to include("is str")
-expect("this string").to start_with("this")
-expect("this string").to end_with("ring")
-expect([1, 2, 3]).to contain_exactly(2, 3, 1)
-expect([1, 2, 3]).to match_array([3, 2, 1])
+expect([1, 2, 3]).to eq([1, 2, 3])            # Order dependent match
+expect([1, 2, 3]).to include(1)               # Exact order, partial range matches
+expect([1, 2, 3]).to include(1, 2)            #
+expect([1, 2, 3]).to start_with(1)            # Exact order, partial range, anchored position
+expect([1, 2, 3]).to start_with(1, 2)         #
+expect([1, 2, 3]).to end_with(3)              #
+expect([1, 2, 3]).to end_with(2, 3)           #
+expect({:a => 'b'}).to include(:a => 'b')     # Hash example
+expect("this string").to include("is str")    # String examples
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -199,6 +199,8 @@ expect("this string").to start_with("this")
 expect("this string").to end_with("ring")
 expect([1, 2, 3]).to contain_exactly(2, 3, 1)
 expect([1, 2, 3]).to match_array([3, 2, 1])
```
xaviershay: maybe add a trailing comment clarifying what this actually does? Similar to other documentation:

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers

It's kind of confusing that we have both `contain_exactly` and `match_array`, when they do the same thing just with different argument conventions.

WDYT about something like:

```
 expect([1, 2, 3]).to contain_exactly(2, 3, 1) # Order independent match
 expect([1, 2, 3]).to match_array([3, 2, 1])   # Equivalent to above
expect([1,2,3]).to eq([1,2,3]) # Order dependent match can use standard eq matcher
```

We should also consider adding a link to https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers "for complete documentation", maybe just below the `## Built-in Matchers` subhead.

ianfixes: I had assumed that the relishapp.com docs were pulled from this README.  If those are stored somewhere else in this repo (a different branch? hidden in some subdir?) then I'd be glad to modify that as well.

README.md
```diff
@@ -175,30 +175,40 @@ expect(1..10).to cover(3)
...
 expect(actual).to end_with(expected)
 
-expect(actual).to contain_exactly(individual, items)
-# ...which is the same as:
-expect(actual).to match_array(expected_array)
+# any order, entire range
+expect(actual).to match_array([expected_element1, expected_element2])
+# Assuming that:
+#   expected = [expected_element1, expected_element2]
+# You can also express this in a variadic function form:
```
myronmarston: I've never heard the term "variadic function form".  How about "You can also express this by passing the expected elements as individual arguments:"

README.md
```diff
@@ -175,30 +175,40 @@ expect(1..10).to cover(3)
...
+expect(actual).to contain_exactly(expected_element1, expected_element2)
+
+# any order, partial range
```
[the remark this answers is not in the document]
ianfixes: OK, so would you say that my "any order" comment is correct?  I'm just going for accuracy here.

README.md
```diff
@@ -175,30 +175,41 @@ expect(1..10).to cover(3)
...
 
-expect(actual).to contain_exactly(individual, items)
-# ...which is the same as:
-expect(actual).to match_array(expected_array)
+# any order, entire range
+expect(actual).to match_array([expected_element1, expected_element2])
+# Assuming that:
+#   expected = [expected_element1, expected_element2]
+# You can also express this as a variadic function,
+#  passing the expected elements as individual arguments:
```
JonRowe: Seconding @myronmarston here, can you remove the reference to variadic function, its overly complicated and confusing to beginners, how about just `You can express this by passing the expected elements as individual arguments:`

JonRowe: It's not common at all, two maintainers of this project both who have over 10 years experience of software development in a variety of languages have never heard of it. Whilst splat is commonly used in Ruby I think `You can express this by passing the expected elements as individual arguments:` is succinct, unambiguous, and doesn't need a technical term.

ianfixes: Maintainers of this project are not the target audience for the documentation, but then neither of us are representative sample sizes.  

I ran these searches to check my assumptions:
* https://www.google.com/search?q=%22variadic%22&tbm=bks
* https://trends.google.com/trends/explore?geo=US&q=variadic,ruby%20splat,kwargs,%22variable-length%20argument%22,varargs which does show a pretty region-specific bias

JonRowe: If I were proposing we use a different term that research would be a good indicator, but in order to be more inclusive (a senior dev can understand that sentence as well as a junior) Im asking for removal of as much technical jargon as possible, this also helps across language barriers.

ianfixes: > technical jargon

It's not jargon, it's literally the name of what you're talking about.
http://rosettacode.org/wiki/Variadic_function


> across language barriers

* https://de.wikipedia.org/wiki/Variadische_Funktion
* https://fr.wikipedia.org/wiki/Fonction_variadique
* https://it.wikipedia.org/wiki/Funzione_variadica
* https://ru.wikipedia.org/wiki/Вариативная_функция (phonetically, "variativnaya funkcia"


It's possible that your perspective is the only valid one here, and that you are paradoxically being inclusive by rejecting all the other ones.  It's possible that my edits only make rspec's documentation worse.  But the purpose of this PR is to "call a spade a spade, not a gardening tool" and your suggestions run contrary to that.

Issue #994, opened 2017-06-21
Is implicit should still the recommended way? Are we favoring `is_expected.to`?


lunks:
According to [Should.md](https://github.com/rspec/rspec-expectations/blob/master/Should.md#one-liners) implicit should reads better. I personally disagree and looking on the documentation itself it seems we are favoring `is_expected.to`.

Is a PR welcomed to remove that section? Or is the statement still valid?

myronmarston:
I think we should keep the section, but update it to mention there's also `is_expected`.  A PR is welcome!

pirj:
This seems to have been addressed.

PR #1061, opened 2018-06-06
Update Should.md to suggest is_expected in one-liners

adimasuhid:
### Description
This change updates `Should.md` to include a description about `is_expected` as suggested in #994. 

### Affected section
screen shot 2018-06-06 at 2 12 31 pm


myronmarston:
Thanks!

PR #1069, opened 2018-08-06
Call #to_hash only when object is not a Hash in RSpec::Matcher::BuiltIn::Include

### LINKED ISSUE #1068: expect(HashWithIndifferentAccess).to include(Symbol) behaves differently in 3.8.0
### Subject of the issue
Today I upgraded to the latest 3.8.0 rspec release and found that a specific test breaks now. After tinkering with it a bit I could isolated the problem. See the code snippet below.

### Your environment
* Ruby version: tested with 2.3, 2.5, 2.6rc2
* rspec-expectations version: 3.7.0 and 3.8.0

### Steps to reproduce
```ruby
require 'active_support/core_ext/hash/indifferent_access'

RSpec.describe 'include matcher difference between 3.7.0 and 3.8.0' do
  it 'should include the keywords' do
    hsh = HashWithIndifferentAccess.new
    hsh['foo'] = 'bar'

    expect(hsh).to include(:foo)
  end
end
```

### Expected behavior

Testsuite passes as in 3.7.0

### Actual behavior

Testsuite fails in 3.8.0

EDIT: Fixed the versions...

eitoball:
This PR should fix #1068.

With the change in #1012, `RSpec::Matchers::BuiltIn::Include` calls `#to_hash` to actual value. If actual value is a`ActiveSupport::HashWithIndifferentAccess`, `#to_hash` converts to ["a hash with string keys"](https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html#method-i-to_hash). This behavior causes expectations like `expect(hsh).to include(:foo)` fail as reported in #1068.

This PR change to call `#to_hash` only when object is not a hash. Because `ActiveSupport::HashWithIndifferentAccess` inherits `Hash`, it will not call `#to_hash`, and `expect(hsh).to include(:foo)` should pass.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -15,14 +15,14 @@ module RSpec
         # @api private
         # @return [Boolean]
         def matches?(actual)
-          actual = actual.to_hash if actual.respond_to?(:to_hash)
+          actual = actual.to_hash if convert_to_hash?(actual)
           perform_match(actual) { |v| v }
         end
 
         # @api private
         # @return [Boolean]
         def does_not_match?(actual)
-          actual = actual.to_hash if actual.respond_to?(:to_hash)
+          actual = actual.to_hash if convert_to_hash?(actual)
           perform_match(actual) { |v| !v }
         end
 
@@ -138,6 +138,10 @@ module RSpec
           expected.any? do |str|
             actual.include?(str) && lines.none? { |line| line == str }
           end
+        end
+
+        def convert_to_hash?(obj)
+          !(::Hash === obj) && obj.respond_to?(:to_hash)
         end
       end
     end
--- a/spec/rspec/matchers/built_in/include_spec.rb
+++ b/spec/rspec/matchers/built_in/include_spec.rb
@@ -1,3 +1,38 @@
+# This class fakes some behavior of
+# ActiveSupport::HashWithIndifferentAccess.
+# It dosen't convert recursively.
+class FakeHashWithIndifferentAccess < Hash
+  class << self
+    def from_hash(hsh)
+      new_hsh = new
+      hsh.each do |key, value|
+        new_hsh[key] = value
+      end
+      new_hsh
+    end
+  end
+
+  def [](key)
+    super(key.to_s)
+  end
+
+  def []=(key, value)
+    super(key.to_s, value)
+  end
+
+  def key?(key)
+    super(key.to_s)
+  end
+
+  def to_hash
+    new_hsh = ::Hash.new
+    each do |key, value|
+      new_hsh[key] = value
+    end
+    new_hsh
+  end
+end
+
 RSpec.describe "#include matcher" do
   it "is diffable" do
     expect(include("a")).to be_diffable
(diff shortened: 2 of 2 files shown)


lib/rspec/matchers/built_in/include.rb
```diff
@@ -139,6 +139,10 @@ def diff_would_wrongly_highlight_matched_item?
             actual.include?(str) && lines.none? { |line| line == str }
           end
         end
+
+        def convert_to_hash?(obj)
+          !obj.is_a?(::Hash) && obj.respond_to?(:to_hash)
```
myronmarston: Please use `!(::Hash === obj)` instead of `!obj.is_a?(::Hash)`.   The reason is that `is_a?` can be defined on any object to do anything -- or can even be undefined on some objects.  For example, `BasicObject` does not define `is_a?`:

```
2.4.4 :003 > BasicObject.new.is_a?(::Hash)
NoMethodError: undefined method `is_a?' for #<BasicObject:0x007fa910e077f0>
	from (irb):3
	from /usr/bin/irb:11:in `<main>'
```

In contrast, `!(::Hash === obj)` will work with _any_ object and we can count on its behavior being consistent unless someone monkey patches `Hash.===`.  Generally speaking we don't support environments where core types are monkey patched to misbehave but we do need to support users passing us arbitrary objects that define their own behavior.

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -131,6 +188,18 @@ def hash.send; :sent; end
       it_behaves_like "a Hash target"
     end
 
+    context "for a target that inherits from Hash" do
```
myronmarston: This doc string doesn't describe this situation as accurately as it could.  How about:

``` ruby
context "for a target that subclasses Hash to treat string/symbol keys interchangeably, but returns a raw hash from #to_hash"
```

Pretty long, but describes the exact situation we're dealing with much more accurately.

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -8,14 +43,23 @@ def build_target(hsh)
       hsh
     end
 
+    def use_string_keys?
```
myronmarston: IMO this method name is confusing: from the name it's not clear in which context this controls string keys being used.  There are still lots of uses of symbol keys even when this is set to `true`.

How about `use_string_keys_in_failure_message?`

eitoball: I totally agree with you, so I changed to `use_string_keys_in_failure_message?`.

lib/rspec/matchers/built_in/include.rb
eitoball: I see. Thank you. I changed it.

spec/rspec/matchers/built_in/include_spec.rb
eitoball: Much better. I change description as you proposed. Thank you.

eitoball:
@myronmarston Thank you for review. I updated where you commented. Also, I added 'undef's to avoid "method redefined" warrings that causes builds on Trivis CI fail.

myronmarston:
Thanks @eitoball!  Merging in spite of the appveyor failure; it's some problem installing the bundle on 1.9.3.

morgoth:
So the regression still persists in 3.8.1 for `ActiveModel::Errors` object.
Having following test:
```ruby
some_model = SomeModel.new
expect { some_model.validate }
  .to change { some_model.errors }
  .to include :fatal_error
```
it was passing before 3.8.0

myronmarston:
@morgoth please open a new issue with a clear repro snippet (ideally w/o using the `change` matcher, as I don't think this problem is isolated to the composition of `change` and `include`, right?).

namusyaka:
Unfortunately, the patch seems not to be enough for fixing sinatra's spec.
- https://github.com/sinatra/sinatra/blob/b801c6b76b89eb071e2b01700273054bbe8aae05/sinatra-contrib/lib/sinatra/cookies.rb
- https://github.com/sinatra/sinatra/blob/b801c6b76b89eb071e2b01700273054bbe8aae05/sinatra-contrib/spec/cookies_spec.rb#L562

Because `Sinatra::Cookie::Jar` doesn't inherit `Hash`. Any ideas?

jkowens:
I've opened PR #1073 which should fix the regressions mentioned by @namusyaka and @morgoth.

Issue #1071, opened 2018-09-06
Rspec respond_to not functional with required keyword in call to new


liamelliott:
I have some code using rspec version 3.8 that doesn't produce the result expected [at this link](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/respond-to-matcher#specify-required-keywords). If I create another class method with the same keywords and behaviour, it works without any issues.

```
class Visit

  Facility = Struct.new(:code, :name) do
    def to_facility
      self
    end
  end

  attr_reader :facility, :start_date, :number_of_nights

  def initialize(facility:, start_date: nil, number_of_nights: 1)
    @facility = facility
    @start_date = start_date
    @number_of_nights = number_of_nights
  end

  def self.other(facility:, start_date: nil, number_of_nights: 1)
    @facility = facility
    @start_date = start_date
    @number_of_nights = number_of_nights
  end

end
```

```
describe Visit do

  describe ".initialize" do
    it "correctly responds to :new" do
      expect(described_class).to respond_to(:other).with_keywords(:facility) # Works!
      expect(described_class).to respond_to(:new).with_keywords(:facility) # Doesn't work!
    end
  end
end
```

This test fails with "expected Visit to respond to :new with keyword :facility," but I don't see how my usage is any different from the relishapp example docs (1 required and 2 other keyword arguments). Only in their case they're using `plant(seed:, fertilizer: nil, water: 'daily)`.

myronmarston:
The underlying issue is that ruby defines `Visit.new` to accept _all_ arguments and pass them through to `#initialize`.  As a result, the metadata available about `Visit.new` does not indicate anything about the `#initialize` arguments:

``` irb
2.4.4 :018 > Visit.method(:new).parameters
 => [[:rest]]
2.4.4 :019 > Visit.instance_method(:initialize).parameters
 => [[:keyreq, :facility], [:key, :start_date], [:key, :number_of_nights]]
```

We solved this for rspec-mocks verifying doubles by looking at the signature of `#initialize` when dealing with `new`:

https://github.com/rspec/rspec-mocks/pull/886

We could theoretically do the same for `respond_to` but someone will have to invest the effort to do that.

liamelliott:
@myronmarston Interesting, thanks. Instead of a direct workaround of the ruby functionality on sending :new, would you support adding something like `expect(described_class).to respond_to(:new).and_defer_to(:initialize).with_keywords(:facility)`?

[...]

PR #1072, opened 2018-09-13
Support checking `initialize` when checking `respond_to(:new)`

JonRowe:
Following on from #1071 when we're checking the arity of a method for `respond_to` and the method being checked is a standard library `new` method, then we should actually check the method signature for the instance method `initialize`. This is similar to what we do in `rspec-mocks` for verifying doubles.

/cc @myronmarston as you wrote the inspiration!

### DIFF
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -125,9 +125,18 @@ module RSpec
 
           return true if expectation.empty?
 
-          signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
+          Support::StrictSignatureVerifier.new(method_signature_for(actual, name)).
+            with_expectation(expectation).valid?
+        end
 
-          Support::StrictSignatureVerifier.new(signature).with_expectation(expectation).valid?
+        def method_signature_for(actual, name)
+          method_handle = Support.method_handle_for(actual, name)
+
+          if name == :new && method_handle.owner === ::Class && ::Class === actual
+            Support::MethodSignature.new(actual.instance_method(:initialize))
+          else
+            Support::MethodSignature.new(method_handle)
+          end
         end
 
         def with_arity
--- a/spec/rspec/matchers/built_in/respond_to_spec.rb
+++ b/spec/rspec/matchers/built_in/respond_to_spec.rb
@@ -33,6 +33,11 @@ RSpec.describe "expect(...).to respond_to(:sym).with(1).argument" do
     expect(obj).to respond_to(:foo).with(1).argument
   end
 
+  it "verifes the method signature of new as if it was initialize" do
+    klass = Class.new { def initialize(a, b); end; }
+    expect(klass).to respond_to(:new).with(2).arguments
+  end
+
   it "fails if target does not respond to :sym" do
     obj = Object.new
     expect {
@@ -64,11 +69,63 @@ RSpec.describe "expect(...).to respond_to(:sym).with(1).argument" do
     }.to fail_with(/expected #<Object.*> to respond to :foo with 1 argument/)
   end
 
+  it "fails if the method signature of initialize does not match" do
+    klass = Class.new { def initialize(a, b); end; }
+    expect {
+      expect(klass).to respond_to(:new).with(1).arguments
+    }.to fail_with(/expected #<Class.*> to respond to :new with 1 argument/)
+  end
+
   it "still works if target has overridden the method method" do
     obj = Object.new
     def obj.method; end
     def obj.other_method(arg); end
     expect(obj).to respond_to(:other_method).with(1).argument
+  end
+end
+
+RSpec.describe "expect(...).to respond_to(:new)" do
+  context "with no tampering" do
+    it "will validate new as if it was initialize" do
+      klass = Class.new { def initialize(a, b, c); end }
+      expect(klass).not_to respond_to(:new).with(2).arguments
+      expect(klass).to respond_to(:new).with(3).arguments
+    end
+  end
+
+  context "on a class that has redefined `new`" do
+    it "uses the method signature of the redefined `new` for arg verification" do
+      klass = Class.new { def self.new(a); end }
+      expect(klass).to respond_to(:new).with(1).argument
+      expect {
+        expect(klass).to respond_to(:new).with(2).arguments
+      }.to fail_with(/expected #<Class.*> to respond to :new with 2 arguments/)
+      expect {
+        expect(klass).to_not respond_to(:new).with(1).argument
+      }.to fail_with(/expected #<Class.*> not to respond to :new with 1 argument/)
+    end
+  end
+
+  context "on a class that has undefined `new`" do
+    it "will not respond to new" do
+      klass =
+        Class.new do
+          class << self
+            undef new
+          end
+        end
+
+      expect {
+        expect(klass).to respond_to(:new)
+      }.to fail_with(/expected .* to respond to :new/)
+    end
+  end
+
+  context "on a class with a private `new`" do
+    it "will not respond to new" do
+      klass = Class.new { private_class_method :new; def initialize(a, b, c); end }
+      expect(klass).not_to respond_to(:new)
+    end
   end
 end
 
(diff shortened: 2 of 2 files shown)


spec/rspec/matchers/built_in/respond_to_spec.rb
```diff
@@ -33,6 +33,12 @@ def obj.foo(a, *args); end
     expect(obj).to respond_to(:foo).with(1).argument
   end
 
+  it "passes if target is new" do
```
myronmarston: This doc string is overly general.  

> expect(...).to respond_to(:sym).with(1).argument passes if target is new

That reads like the expectation will always pass if the target is `new`, regardless of how `initialize` is defined.

spec/rspec/matchers/built_in/respond_to_spec.rb
```diff
@@ -493,6 +515,16 @@ def obj.some_method(arg, arg2, arg3, *args); end
...
+    it 'passes if target is :new with keywords' do
+      # note we can't use the metaobject definition for initialize
+      klass = eval %{Class.new { def initialize(a: nil, b: nil); end}}
+      expect(klass).to respond_to(:new).with_keywords(:a, :b)
+
+      # note we can't use the metaobject definition for initialize
+      klass_2 = eval %{Class.new { def initialize(**kw_args); end}}
+      expect(klass_2).to respond_to(:new).with_keywords(:a, :b)
+    end
+
```
myronmarston: We need some specs for when `new` has been monkeyed with.  See the 3 contexts starting here:

https://github.com/rspec/rspec-mocks/pull/886/files#diff-7a6cccf0d2c37e75e4affc0a24501df6R426

It looks like your implementation handles this but you haven't written specs for it.

spec/rspec/matchers/built_in/respond_to_spec.rb
```diff
@@ -493,6 +515,16 @@ def obj.some_method(arg, arg2, arg3, *args); end
...
     end
 
+    it 'passes if target is :new with keywords' do
+      # note we can't use the metaobject definition for initialize
+      klass = eval %{Class.new { def initialize(a: nil, b: nil); end}}
+      expect(klass).to respond_to(:new).with_keywords(:a, :b)
+
+      # note we can't use the metaobject definition for initialize
+      klass_2 = eval %{Class.new { def initialize(**kw_args); end}}
+      expect(klass_2).to respond_to(:new).with_keywords(:a, :b)
```
myronmarston: Seems like we need some examples of the expectations failing. All the ones you've added to these specs are of passing expectations.

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -125,7 +125,14 @@ def matches_arity?(actual, name)
...
 
-          signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
+          method_handle = Support.method_handle_for(actual, name)
+
+          signature =
+            if name == :new && method_handle.owner === ::Class && Class === actual
+              Support::MethodSignature.new(actual.instance_method(:initialize))
+            else
+              Support::MethodSignature.new(method_handle)
+            end
```
myronmarston: Nice to see how simple this is!

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -125,7 +125,14 @@ def matches_arity?(actual, name)
 
           return true if expectation.empty?
 
-          signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
+          method_handle = Support.method_handle_for(actual, name)
+
+          signature =
+            if name == :new && method_handle.owner === ::Class && Class === actual
```
myronmarston: Why use `::Class` in one spot and `Class` (no `::` in another)?

spec/rspec/matchers/built_in/respond_to_spec.rb
JonRowe: I agree, reworded

spec/rspec/matchers/built_in/respond_to_spec.rb
JonRowe: Theres one failing, but you're right it might not be enough

lib/rspec/matchers/built_in/respond_to.rb
JonRowe: I'm wondering if we need to do the same hoops for an instance_method as a method

lib/rspec/matchers/built_in/respond_to.rb
JonRowe: Oversight, fixed

spec/rspec/matchers/built_in/respond_to_spec.rb
benoittgt: Something like?
```ruby
klass_3 = eval %{Class.new { def initialize(a: nil, b: nil); end}}
expect {
  expect(klass_3).to respond_to(:new).with_keywords(1..2)
}.to fail_with(/expected #<Class.*> to respond to :new with keyword 1..2/)
```

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -125,7 +125,14 @@ def matches_arity?(actual, name)
 
           return true if expectation.empty?
 
-          signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
+          method_handle = Support.method_handle_for(actual, name)
+
+          signature =
```
benoittgt: CI is broken on linting

```
lib/rspec/matchers/built_in/respond_to.rb:112:9:C: Metrics/MethodLength: Method has too many lines. [19/15]
        def matches_arity?(actual, name)...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
48 files inspected1 offense detected
```

Maybe `signature` assignment can be extracted as method?

JonRowe:
Added a bunch more examples and some of the edge cases for tampering with new, probably worth a re-review

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -125,9 +125,20 @@ def matches_arity?(actual, name)
...
 
-          signature = Support::MethodSignature.new(Support.method_handle_for(actual, name))
 
-          Support::StrictSignatureVerifier.new(signature).with_expectation(expectation).valid?
+          Support::StrictSignatureVerifier.new(
+            build_method_signature(actual, name)
+          ).with_expectation(expectation).valid?
+        end
+
+        def build_method_signature(actual, name)
```
myronmarston: nit: this method doesn't do much "building" -- that's what `Support::MethodSignature` does.  Maybe call it `method_signature_for(actual, name)`?

spec/rspec/matchers/built_in/respond_to_spec.rb
```diff
@@ -72,6 +84,45 @@ def obj.other_method(arg); end
...
+      klass = Class.new { def initialize(a, b, c); end }
+      expect(klass).not_to respond_to(:new).with(2).arguments
+      expect(klass).to respond_to(:new).with(3).arguments
+    end
+  end
+
+  context "on a class that has redefined `new`" do
+    it "uses the method signature of the redefined `new` for arg verification" do
+      klass = Class.new { def self.new(a); end }
+      expect(klass).to respond_to(:new).with(1).argument
```
myronmarston: It would be good to show a failure for `with(2).arguments` to demonstrate that it is actually using the signature of the redefined `new`; otherwise the implementation could just always pass when it has been redefined.

myronmarston:
LGTM outside the build failure

lib/rspec/matchers/built_in/respond_to.rb
JonRowe: Ha I think I had that first then changed it, happy to oblige.

spec/rspec/matchers/built_in/respond_to_spec.rb
JonRowe: Done

JonRowe:
Merging as appveyor is playing up, have at least 1 green windows build.

PR #1073, opened 2018-09-15
Only call to_hash when actual object does not respond to include?

jkowens:
According to the docs for the built-in include matcher, if an object responds to `#include?` that method will be called. Currently, if an object implements `#to_hash`, the `#include?` method on the returned `Hash` is called instead (unless the object is already a Hash).

This change in behavior was introduced in https://github.com/rspec/rspec-expectations/commit/a0c39589979befc8279c2d4f78579a504d7dfcff.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -141,7 +141,7 @@ module RSpec
         end
 
         def convert_to_hash?(obj)
-          !(::Hash === obj) && obj.respond_to?(:to_hash)
+          !obj.respond_to?(:include?) && obj.respond_to?(:to_hash)
         end
       end
     end
--- a/spec/rspec/matchers/built_in/include_spec.rb
+++ b/spec/rspec/matchers/built_in/include_spec.rb
@@ -128,6 +128,14 @@ RSpec.describe "#include matcher" do
       end
     end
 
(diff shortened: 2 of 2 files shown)


xaviershay:
thank you! Want to try including a spec so we don't regress in the future?

jkowens:
@xaviershay sure thing 👍

jkowens:
@xaviershay I added a spec, hopefully it looks ok.

jkowens:
@JonRowe #1069 came with spec coverage. I don't think this should regress that change...but this is my first time sending up a PR for rspec. Is there something more you think should be considered?

https://github.com/rspec/rspec-expectations/blob/31f108eae68e31ee65f0622c9ef4a488a1602943/spec/rspec/matchers/built_in/include_spec.rb#L191

JonRowe:
@jkowens I'd accept a manual test and a note in the commit that it's been checked with a real version of active support

jkowens:
@JonRowe I carried out a manual test and updated the commit message as suggested.

betesh:
Thanks for this fix.  @JonRowe Can I trouble you to publish this fix by releasing rspec-expectations 3.8.2?

JonRowe:
Released as 3.8.2

betesh:
Thank you!

PR #1076, opened 2018-09-19
Handle unexpected Index / KeyError when looking up match results

### LINKED ISSUE #1074: Uninformative `KeyError` with custom matchers
### Subject of the issue
When I create a custom matcher which expects an error and forget implementing `expect_call_stack_jump?`, I get `KeyError` in a certain condition.
This error seems uninformative for users, so some other error should be raised.

### Your environment
* Ruby version: 2.5.0
* rspec-expectations version: 3.8.1

### Steps to reproduce
```ruby
RSpec::Matchers.define :exit_script do
  supports_block_expectations

  # def expects_call_stack_jump?
  #   true
  # end

  match do |block|
    begin
      block.call
    rescue SystemExit
      true
    else
      false
    end
  end
end

RSpec.describe 'exit_script' do
  context 'when not composed' do
    it 'matches if Kernel.exit called' do
      expect { exit }.to exit_script
    end
  end

  context 'when composed using and' do
    it 'matches if Kernel.exit called' do
      x = 0
      expect { x += 1; exit }.to change { x }.and exit_script
    end
  end

  context 'when composed using or' do
    it 'matches if Kernel.exit called' do
      x = 0
      expect { x += 1; exit }.to change { x }.or exit_script
    end
  end
end
```

### Expected behavior
An error with more helpful information

### Actual behavior
```
Failures:

  1) exit_script when composed using and matches if Kernel.exit called
     Failure/Error: expect { x += 1; exit }.to change { x }.and exit_script
     
     KeyError:
       key not found: #<RSpec::Matchers::BuiltIn::Change:0x000055bdf61c9440 @receive...
       Did you mean?  #<RSpec::Matchers::DSL::Matcher exit_script>
     # ./spec/sample_spec.rb:29:in `block (3 levels) in <top (required)>'

  2) exit_script when composed using or matches if Kernel.exit called
     Failure/Error: expect { x += 1; exit }.to change { x }.or exit_script
     
     KeyError:
       key not found: #<RSpec::Matchers::BuiltIn::Change:0x000055bdf6199da8 @receive...
       Did you mean?  #<RSpec::Matchers::DSL::Matcher exit_script>
     # ./spec/sample_spec.rb:36:in `block (3 levels) in <top (required)>'
```

KeyError itself is raised at [`RSpec::Matchers::BuiltIn::Compound::NestedEvaluator#matcher_matches?`](https://github.com/rspec/rspec-expectations/blob/89472853437828e6460cd4bc69708df7ff81e0c5/lib/rspec/matchers/built_in/compound.rb#L156).

JonRowe:
When an unexpected local jump (and probably some other conditions) happens in some composed matchers, this result in an unexpected KeyError.

Fixes #1074

### DIFF
--- a/lib/rspec/matchers/built_in/compound.rb
+++ b/lib/rspec/matchers/built_in/compound.rb
@@ -154,7 +154,12 @@ module RSpec
           end
 
           def matcher_matches?(matcher)
-            @match_results.fetch(matcher)
+            @match_results.fetch(matcher) do
+              raise ArgumentError, "Your #{matcher.description} has no match " \
+               "results, this can occur when an unexpected call stack or " \
+               "local jump occurs. Prehaps one of your matchers needs to " \
+               "declare `expects_call_stack_jump?` as `true`?"
+            end
           end
 
         private
--- a/spec/rspec/matchers/composable_spec.rb
+++ b/spec/rspec/matchers/composable_spec.rb
@@ -95,6 +95,28 @@ module RSpec
           expect(with_matchers_cloned([stdout]).first).to equal(stdout)
         end
       end
+
+      describe "when an unexpected call stack jump occurs" do
+        RSpec::Matchers.define :cause_call_stack_jump do
+          supports_block_expectations
+
+          match do |block|
+            begin
+              block.call
+              false
+            rescue Exception
+              true
+            end
+          end
+        end
+
+        it "issue a warning suggesting `expects_call_stack_jump?` has been improperly declared" do
+          expect {
+            x = 0
+            expect { x += 1; exit }.to change { x }.and cause_call_stack_jump
+          }.to raise_error(/no match results, [\.\w\s]+ declare `expects_call_stack_jump\?`/)
+        end
+      end
     end
   end
 end


lib/rspec/matchers/built_in/compound.rb
```diff
@@ -155,6 +155,13 @@ def initialize(actual, matcher_1, matcher_2)
 
           def matcher_matches?(matcher)
             @match_results.fetch(matcher)
+          rescue IndexError
```
myronmarston: Instead of rescuing this, I believe you can just pass this `raise` expression in a block to `fetch`.  The block is called only if the provided key is absent.  It's an explicit, built-in way with `fetch` to run custom logic when the key is missing.

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -155,6 +155,13 @@ def initialize(actual, matcher_1, matcher_2)
...
+            raise ArgumentError,
+              """
+                A matcher has no match results, this can occur when an unexpected
```
myronmarston: "A matcher has no match results" is kinda vague.  Would be good to use `matcher.description` in here somewhere to be more specific.  Maybe: 

```
"Your '#{matcher.description}' matcher was unable to produce a match result. " \
"Usually this happens when one of the matchers in your compound expression " \
"experiences a call stack jump (e.g. via a `throw` or `raise`) without defining " \
"`expects_call_stack_jump?` to return `true`.
```

Also, notice my use of `\`; that concats the multi-line string at parse time.  I think it's better to have a single long string that wraps in the user's terminal based on whatever their terminal width is rather than hard-wrapping it for them based on the width you decided on here.

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -155,6 +155,13 @@ def initialize(actual, matcher_1, matcher_2)
...
+                call stack or local jump occurs. Prehaps one of your matchers needs
+                to declare `expects_call_stack_jump?` as true?
+              """.gsub(/^\s+/,'').tr("\n"," ")
```
myronmarston: Haha, I just realized you are removing the wrapping here....I hadn't noticed.  I still think `\` is simpler.

spec/rspec/matchers/composable_spec.rb
```diff
@@ -95,6 +95,26 @@ module Matchers
...
+
+          match do |block|
+            block.call
+            false
+          rescue Exception
+            true
+          end
+        end
+
+        it "will not raise an index or key error" do
```
myronmarston: It's generally better to positively express what the behavior _is_ rather than negatively express what the behavior _is not_.  The index or key error was entirely due to an internal implementation detail (our use of a hash and `Hash#fetch`).  Instead, I think it's better to say something like:

```
it "provides a clear directions to the user about the need for `expects_call_stack_jump?`" do
  # ...
end
```

Issue #995, opened 2017-06-21
have_attributes matcher does not show diff of mismatched attributes


segiddins:
Running the following file, I get the following output for the two different expectations

```ruby
HasString = Struct.new(:string)

RSpec.describe "foo" do
  it "bar" do
    hs = HasString.new("abc\ndef\nghij\n\n493207428")

    # expect(hs.string).to eq "def"
=begin
Failures:

  1) foo bar
     Failure/Error: expect(hs.string).to eq "def"

       expected: "def"
            got: "abc\ndef\nghij\n\n493207428"

       (compared using ==)

       Diff:

       @@ -1,2 +1,6 @@
       +abc
        def
       +ghij
       +
       +493207428

     # ./spec/foo_spec.rb:6:in `block (2 levels) in <top (required)>'
=end
    expect(hs).to have_attributes(string: "def")
=begin
Failures:

  1) foo bar
     Failure/Error: expect(hs).to have_attributes(string: "def")

       expected #<struct HasString string="abc\ndef\nghij\n\n493207428"> to have attributes {:string => "def"} but had attributes {:string => "abc\ndef\nghij\n\n493207428"}
       Diff:
       @@ -1,2 +1,2 @@
       -:string => "def",
       +:string => "abc\ndef\nghij\n\n493207428",

     # ./spec/foo_spec.rb:30:in `block (2 levels) in <top (required)>'
=end
  end
end
```

I want a way to get the diff from the `eq` matcher, but still printing the `to_s`/`inspect` from the object in question, as `has_attributes` does.

My guess is that it's showing the hash diff, but if it showed something like:

```
:string =>
          +abc
           def
          +ghji
          +493207428
```

I think that'd be much more helpful, as diffs make failing tests _much_ more informative

phylor:
For people coming from search engines, https://github.com/mcmire/super_diff improves a lot on the output of `have_attributes` and other matchers.

JonRowe:
Closing during triage of issues during the monorepo migration, we have a bunch of similar issues around improving / adding swapable differs.

PR #1078, opened 2018-09-23
Add ability for raise_error to create expectations on error attributes

jcouball:
I could not find an easy way to to define an expectation for an error that has multiple attributes (like `SystemExit`.  Please let me know if I missed something.

This change adds a `with_attributes` chain to the `raise_error` matcher (like the existing `with_message` chain).

Internally, `with_attributes` uses the `have_attributes` matcher, so all functionality of that matcher is exposed with `with_attributes`.

Here is how it works:

```Ruby
expect { raise SystemExit.new(1, 'File not found') }.to(
  raise_error(SystemExit).with_attributes(message: 'File not found', status: 1)
)
```

or

```Ruby
expect { raise SystemExit.new(1, 'FILE NOT FOUND') }.to(
  raise_error(SystemExit).with_attributes(message: /not found/i, status: 1..255)
)
```

This change does not change any existing functionality.

I have followed all the directions and the checklist from https://github.com/rspec/rspec-expectations/blob/master/DEVELOPMENT.md

### DIFF
--- a/features/built_in_matchers/raise_error.feature
+++ b/features/built_in_matchers/raise_error.feature
@@ -26,7 +26,11 @@ Feature: `raise_error` matcher
     expect { raise "oops" }.to raise_error(/op/)
     expect { raise "oops" }.to raise_error(RuntimeError, "oops")
     expect { raise "oops" }.to raise_error(RuntimeError, /op/)
+    expect { raise SystemExit.new(1, 'oops') } to raise_error(SystemExit).with_attributes(status: 1, message: 'oops')
+    expect { raise SystemExit.new(1, 'oops') } to raise_error(SystemExit).with_attributes(status: 1, message: /op/)
     ```
+
+  with_attributes uses the have_attributes matcher internally.  See the have_attributes matcher documentation for more details.
 
   Scenario: expect any error
     Given a file named "example_spec" with:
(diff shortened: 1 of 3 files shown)


myronmarston:
Thanks for taking a stab at implementing this yourself, @jcouball!  As it turns out, RSpec already supports what you're looking for, but with a slightly different syntax.

Since RSpec 3.0, we've supported [composable matchers](http://rspec.info/blog/2014/01/new-in-rspec-3-composable-matchers/) which allow you to pass a matcher as an argument to another matcher -- such as passing `have_attributes` (or it's noun-phrased alias, `an_object_having_attributes`) to `raise_error`:

``` ruby
expect {
  raise SystemExit.new(1, 'File not found')
}.to raise_error(an_object_having_attributes(message: 'File not found', status: 1))
```

This doesn't check the exception class like your example does, but that's easily rectified by passing a compound matcher expression:

``` ruby
expect {
  raise SystemExit.new(1, 'File not found')
}.to raise_error(an_instance_of(SystemExit).and
                 an_object_having_attributes(message: 'File not found', status: 1))
```

...or, since `class` is ultimately just an attribute as well, you can specify the class of the exception as another attribute:

``` ruby
expect {
  raise SystemExit.new(1, 'File not found')
}.to raise_error(an_object_having_attributes(message: 'File not found', status: 1, class: SystemExit))
```

Our composable matcher support is ultimately more flexible than what your PR adds here (your PR adds support for matching on attributes but the composable matchers allow you to match on _anything_), so I'm going to close this.

Thanks again!

jcouball:
Thank you, @myronmarston.  Learning this was worth the price of the work I did.  I appreciate you taking the time to give this feedback.

I notice that the failure message is lacking when a composable matcher is used.  The output does not take into account the composed matcher:

```Ruby
RSpec.describe 'expect SystemExit' do
  let(:actual_status) { 99 }
  it 'should be able to expect SystemExit' do
    expect {
      raise SystemExit.new(actual_status, 'File not found')
    }.to raise_error(an_instance_of(SystemExit).and an_object_having_attributes(message: 'File not found', status: 1))
  end
end
```

Outputs:

```
...
expected Exception with an instance of SystemExit and an object having attributes {:message => "File not found", :status => 1}, got #<SystemExit: File not found> with backtrace:
...
```

I would expect the `got #<SystemExit: File not found>` to include information about the composed matchers.  In this case, I'd expect it to say something like, `got an instance of SystemExit with attributes {:message => "File not found", :status => 99} with backtrace: ...`

Is this something that I could improve?  Do you know of an example where failure message includes actual values from the composable matchers?

myronmarston:
@jcouball that's definitely something that can be improved -- see #759 and #995.

PR #1090, opened 2018-12-14
Improve documentation surrounding array matchers

JonRowe:
Replacement of #1060 since it was closed, builds on @ianfixes work with some language tweaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,30 +175,45 @@ expect(1..10).to cover(3)
 ### Collection membership
 
 ```ruby
-expect(actual).to include(expected)
+# exact order, entire collection
+expect(actual).to eq(expected)
+
+# exact order, partial collection (based on an exact position)
 expect(actual).to start_with(expected)
 expect(actual).to end_with(expected)
 
-expect(actual).to contain_exactly(individual, items)
-# ...which is the same as:
-expect(actual).to match_array(expected_array)
+# any order, entire collection
+expect(actual).to match_array(expected)
+
+# You can also express this by passing the expected elements
+# as individual arguments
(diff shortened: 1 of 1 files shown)


JonRowe:
@myronmarston @xaviershay as you helped review the original PR care to comment?

README.md
```diff
@@ -175,30 +175,47 @@ expect(1..10).to cover(3)
 ### Collection membership
 
 ```ruby
-expect(actual).to include(expected)
+# exact order, entire range
```
myronmarston: I think "range" is potentially confusing since ruby has a range class and this isn't referring to that.  How about using "collection" instead of "range" here and below?

README.md
```diff
@@ -175,30 +175,47 @@ expect(1..10).to cover(3)
...
 expect(actual).to end_with(expected)
 
-expect(actual).to contain_exactly(individual, items)
-# ...which is the same as:
-expect(actual).to match_array(expected_array)
+# any order, entire range
+expect(actual).to match_array(expected)
+
+# Assuming that:
+#   expected = [expected_element1, expected_element2]
```
myronmarston: This comment is confusing.  Initially I read it to be saying, "this line assumes the array is in the exact order `[expected_element1, expected_element2]` but it does not assume any order.  Now I'm realizing that `expected` here refers to `expected` in the `expect(actual).to match_array(expected)` line above.  But that was totally not obvious and took me reading this a few times to realize.

I don't think these 2 lines are necessary and I think they confuse more than they clarify.

JonRowe:
Thanks for the initial work on this @ianfixes

Issue #627, opened 2014-08-15
Can I custom diffable's behaviour by just comparing obj's attributes?


tony612:
I have a custom matcher for rails's AR like this:

```ruby
RSpec::Matchers.define :has_same_attributes_as do |expected|
  ignored = %w[id updated_at created_at]
  match do |actual|
    actual.attributes.except(*ignored) == expected.attributes.except(*ignored)
  end

  failure_message do |actual|
    "Expected strength of #{expected.attributes.except(*ignored)}, but had #{actual.attributes.except(*ignored)}"
  end

  diffable
end
```

I just want to compare the attributes except `id created_at updated_at`, but when I use diffable it still returns the diff about `excepted` obj and `actual` obj instead of their attributes.

Is there any way to solve this? thx

PR #1096, opened 2019-01-16
Allow to define custom differ object in matchers

sponomarev:
The questions about custom differ already appeared in the past in rspec/rspec-expectations#627, rspec/rspec#76. I think it could be valuable to allow custom differ objects in matchers.

The proposed solution should be completely backward compatible. And, I hope, will provide a way to extend custom RSpec matchers libraries.

### DIFF
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -19,18 +19,21 @@ module RSpec
 
       # Raises an RSpec::Expectations::ExpectationNotMetError with message.
       # @param [String] message
-      # @param [Object] expected
-      # @param [Object] actual
+      # @param [Object] matcher
       #
-      # Adds a diff to the failure message when `expected` and `actual` are
+      # Adds a diff to the failure message when `matcher.expected` and `matcher.actual` are
       # both present.
-      def fail_with(message, expected=nil, actual=nil)
+      def fail_with(message, matcher=nil)
         unless message
           raise ArgumentError, "Failure message is nil. Does your matcher define the " \
                                "appropriate failure_message[_when_negated] method to return a string?"
         end
 
-        message = ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(expected).message_with_diff(message, differ, actual)
+        message = if !matcher.nil?
+          ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(matcher.expected).message_with_diff(message, differ, matcher.actual)
+        else
+          ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(nil).message_with_diff(message, differ, nil)
+        end
 
         RSpec::Support.notify_failure(RSpec::Expectations::ExpectationNotMetError.new message)
       end
--- a/lib/rspec/expectations/fail_with.rb
+++ b/lib/rspec/expectations/fail_with.rb
@@ -30,10 +30,12 @@ module RSpec
         end
 
         message = if !matcher.nil?
-          ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(matcher.expected).message_with_diff(message, differ, matcher.actual)
-        else
-          ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(nil).message_with_diff(message, differ, nil)
-        end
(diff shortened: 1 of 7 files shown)


lib/rspec/expectations/fail_with.rb
```diff
@@ -19,18 +19,23 @@ def differ
...
       # @param [String] message
-      # @param [Object] expected
-      # @param [Object] actual
+      # @param [Object] matcher
       #
-      # Adds a diff to the failure message when `expected` and `actual` are
+      # Adds a diff to the failure message when `matcher.expected` and `matcher.actual` are
       # both present.
-      def fail_with(message, expected=nil, actual=nil)
+      def fail_with(message, matcher=nil)
```
JonRowe: I'm afraid you can't do this, it changes a public api.

spec/rspec/expectations/fail_with_spec.rb
```diff
@@ -45,29 +48,49 @@
...
+
   before do
     allow(RSpec::Matchers.configuration).to receive_messages(:color? => true)
   end
 
   it "tells the differ to use color" do
-    expected = "foo bar baz\n"
-    actual = "foo bang baz\n"
+    expected = "foo bang baz\n"
+    actual = "foo bar baz\n"
```
JonRowe: To prove your new implementation works, these should have been left alone.

JonRowe (changes requested):
As this stands this changes a public API so cannot be merged at this point. I'm also not sure having a matcher level differ makes any sense, we're interested in building a new built in differ for RSpec 4, and I'm open to providing configurable differs globally, but per matcher has problems.

lib/rspec/expectations/fail_with.rb
```diff
@@ -19,18 +19,23 @@ def differ
...
         unless message
           raise ArgumentError, "Failure message is nil. Does your matcher define the " \
                                "appropriate failure_message[_when_negated] method to return a string?"
         end
 
-        message = ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(expected).message_with_diff(message, differ, actual)
+        message = if !matcher.nil?
```
JonRowe: As you can see above we already inject the differ, it just can't come from a param here, or at least, it'd have to be an extra param.

lib/rspec/expectations/fail_with.rb
```diff
@@ -19,18 +19,23 @@ def differ
...
+                    differ_in_use = matcher.respond_to?(:differ) ? matcher.differ : differ
+
+                    ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(matcher.expected).message_with_diff(message, differ_in_use, matcher.actual)
+                  else
+                    ::RSpec::Matchers::ExpectedsForMultipleDiffs.from(nil).message_with_diff(message, differ, nil)
```
JonRowe: I'm yeah nahing this, expected and actual need to be passed in, differ could be an optional extra, this would eliminate the need for this if.

spec/rspec/expectations/fail_with_spec.rb
```diff
@@ -45,29 +48,49 @@
     expected_diff = dedent(<<-EOS)
       |
       |@@ -1,2 +1,2 @@
-      |-["poo", "car"]
-      |+[(a string matching /foo/), (a string matching /bar/)]
+      |-[(a string matching /foo/), (a string matching /bar/)]
+      |+["poo", "car"]
```
JonRowe: To be backwards compatible, existing specs shouldn't change...

pirj:
Would you be interested in taking another approach to the problem @sponomarev ?

benoittgt:
I am interested to work on this if @sponomarev can not, after I finish the task on documentation and yard. https://github.com/rspec/rspec-dev/pull/233

sponomarev:
@pirj Yeah, for sure. I'll take care of it this week. @benoittgt thanks for the offer, I think I'll deal with it on my own =)

pirj:
@sponomarev A friendly reminder

benoittgt:
@sponomarev Little bump on this one 😊

JonRowe:
I'm closing this as part of the monorepo migration, we are interested in a pluggable differ overall but this isn't the right approach.

Issue #1070, opened 2018-09-06
Rspec 3.8 .change breaks object comparison (Stripe::Charge.list)


KelseyDH:
### Subject of the issue
I have a working rspec test for `Stripe::Charge.list` that works in rspec 3.7.1 but breaks in rspec 3.8.x

### Your environment
* Ruby version: 2.5.1
* rspec-expectations version: 3.8.1

### Steps to reproduce
The Stripe::Charge.list object id changes every time it is called, even if the data it gives back is the same.  Meaning a call like this now fails:

```
            it "does not create a charge in stripe" do
              expect{ true }.to_not change{ Stripe::Charge.list }
            end
```

### Expected behavior
The expected behaviour is that calls to two Stripe::Charge.list objects should consider not changed if the attributes within that object remain unchanged.

E.g.:

```
#<Stripe::ListObject:0x3feae3af47b4> JSON: {
  "object": "list",
  "data": [
[...]

PR #1100, opened 2019-02-07
Handle sudden subject hash change

pirj:
#1034 introduced a hash comparison in addition to equality comparison, and it broke the following examples (using `rspec-rails`):

    it { expect { 'noop' }.not_to change { country.reload.cities } }

a workaround to that was to use the `.to_a` on the relation:

    it { expect { 'noop' }.not_to change { country.reload.cities.to_a } }

and the underlying problem is with the objects that are changing their hash during the comparison.

[Notable example](https://github.com/rails/rails/blob/788eb51df3007d94b266965433b0be88a23b84c5/activerecord/lib/active_record/associations/collection_proxy.rb#L959):

    def ==(other)
      load_target == other
    end

The problem is originally described [here](https://github.com/rspec/rspec-expectations/issues/1030#issuecomment-450823945), and the proposed solution was to raise the question against Rails team to fix the `CollectionProxy#hash` behaviour, but I believe this wouldn't work for the following reasons:

1. The issue is specific to RSpec, while Rails comes with its own testing framework that does the assertions using `==`, and is not affected by the problem in question.

2. Fixing the issue would require to either:
 - stick to the original hash that the object was assigned that breaks the 'dynamic hashcode value depending on the contents of the object itself' that mutable Ruby objects seem to adhere to
 - pre-load the relation that may be a huge performance hit

*The change swaps conditions in a way that the hashcode of the objects being compared is not yet updated by the other sneaky condition.*

This might also fix the issue described in #1070

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -373,6 +373,7 @@ module RSpec
           event_proc.call
 
           @actual_after = evaluate_value_proc
+          @actual_hash = @actual_after.hash
           true
         end
 
@@ -387,8 +388,9 @@ module RSpec
           #
           # Note that it is not sufficient to only check the hashes; it is
           # possible for two values to be unequal (and of different classes)
-          # but to return the same hash value.
-          @actual_before != @actual_after || @before_hash != @actual_after.hash
+          # but to return the same hash value. Also, some objects may change
+          # their hash after being compared with `==`/`!=`.
+          @actual_before != @actual_after || @before_hash != @actual_hash
         end
 
         def actual_delta
--- a/spec/rspec/matchers/built_in/change_spec.rb
+++ b/spec/rspec/matchers/built_in/change_spec.rb
@@ -101,6 +101,20 @@ RSpec.describe "expect { ... }.to change ..." do
     }.to change { obj.x }
   end
 
+  it 'does not detect changes in an object that updates its hash upon comparison' do
+    obj = Class.new do
+      def ==(another)
+        @hash = rand # (^ °=°)^ #
+        self.object_id == another.object_id
+      end
+      def hash
+        @hash ||= super
+      end
+    end.new
+
+    expect { }.not_to change { obj }
+  end
+
   context "with nil value" do
     before(:example) do
       @instance = SomethingExpected.new


lib/rspec/matchers/built_in/change.rb
```diff
@@ -387,8 +387,9 @@ def changed?
           #
           # Note that it is not sufficient to only check the hashes; it is
           # possible for two values to be unequal (and of different classes)
-          # but to return the same hash value.
-          @actual_before != @actual_after || @before_hash != @actual_after.hash
+          # but to return the same hash value. Also, some objects may change
+          # their hash after being compared with `==`/`!=`.
+          @before_hash != @actual_after.hash || @actual_before != @actual_after
```
JonRowe: Can we restore this line to its original ordering, but cache the hash value just after line 375
e.g.
```
@actual_before != @actual_after || @before_hash != @actual_hash
```

pirj: That works as well, will update shortly.
Any objections to using a local variable for `actual_hash`?

JonRowe: I'd prefer it if we captured the hash just after the variable, (to minimise the chance of it changing) so it needs to be an ivar from that perspective

lib/rspec/matchers/built_in/change.rb
```diff
@@ -387,8 +387,10 @@ def changed?
           #
           # Note that it is not sufficient to only check the hashes; it is
           # possible for two values to be unequal (and of different classes)
-          # but to return the same hash value.
-          @actual_before != @actual_after || @before_hash != @actual_after.hash
+          # but to return the same hash value. Also, some objects may change
+          # their hash after being compared with `==`/`!=`.
+          actual_hash = @actual_after.hash
```
JonRowe: As indicated in my last comment on the old code, I'd prefer this be moved up to just after `@actual_after = evaluate_value_proc` and be an `@actual_hash` definition.

lib/rspec/matchers/built_in/change.rb
pirj: Works either way. But all ivars is more visually appealing. Changed.

JonRowe: Sorry to be a pain, but as per the other comment I would like it moved up to the `perform_change` method, as this would be the first chance to grab the hash and guards against mutation happening elsewhere!

lib/rspec/matchers/built_in/change.rb
pirj: Ah, sorry, I initially got your idea wrong. Fixed.
Does it look up to standards now?

pirj:
Awesome. Thanks for your patience and accepting this blazingly fast, Jon!

stephaniefiddy:
When is this change scheduled to be released? I'm experiencing the same issue when using the change matcher on associations.

JonRowe:
We don't have a release schedule, things tend to go out as warranted. However I'm happy to announce I've just shipped 3.8.3 containing this patch.

PR #1104, opened 2019-03-08
Fix RuboCop offences

pirj:
I've kept the commits separate for easier review, with descriptive commit messages where I had to make a decision that needs explanation.

Tried to keep the changes minimal, picking the style matching the code the most (specifically with %-literals and spaces inside brackets).

RuboCop version update is better to be done simultaneously with the other rspec-* repositories since a number of rules are changed/renamed/removed, so I decided not to do that in the scope of this pull request.


.rubocop.yml
```diff
@@ -42,6 +42,10 @@ AccessModifierIndentation:
   Exclude:
     - 'lib/rspec/expectations/syntax.rb' # Too much diff to fix
 
+Layout/SpaceInsideBlockBraces:
+  EnforcedStyle: space
+  EnforcedStyleForEmptyBraces: space
```
JonRowe: Instead of enforcing a space, can you enforce none? Would reduce churn

.rubocop.yml
```diff
@@ -50,6 +50,9 @@ AccessModifierIndentation:
   Exclude:
     - 'lib/rspec/expectations/syntax.rb' # Too much diff to fix
 
+Layout/AlignParameters:
+  EnforcedStyle: with_first_parameter
```
JonRowe: Can we enforce a style that leaves our preferred spacing? Then we can revert the rest of this commits changes.

.rubocop.yml
```diff
@@ -55,6 +55,9 @@ AccessModifierIndentation:
 Layout/AlignParameters:
   EnforcedStyle: with_first_parameter
 
+Layout/SpaceInsideHashLiteralBraces:
+  EnforcedStyle: no_space
```
JonRowe: Is there an enforced space style? Which changes more?

spec/rspec/matchers/built_in/change_spec.rb
```diff
@@ -104,7 +104,7 @@ class SomethingExpected
   it 'does not detect changes in an object that updates its hash upon comparison' do
     obj = Class.new do
       def ==(another)
-        @hash = rand # (^ °=°)^ #
+        @hash = rand # (^ '=')^ #
```
JonRowe: You can just remove the comment

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -708,11 +708,8 @@ def foo
             end
           end
 
-          begin
-            aggregate_failures do
-              expect(1).to rescue_failure
-            end
-          rescue RSpec::Expectations::ExpectationNotMetError
+          aggregate_failures do
+            expect(1).to rescue_failure
```
JonRowe: You can disable the cop here, and restore this, its to prevent a failure causing the suite problems

.rubocop.yml
```diff
@@ -13,6 +13,9 @@ LineLength:
 Style/BlockComments:
   Enabled: false
 
+Style/Semicolon:
+  Enabled: false
+
```
JonRowe: The `x = 0`, `x += 1` is a wanted part of proving a stack jump

REPORT_TEMPLATE.md
```diff
@@ -6,8 +6,6 @@ DO NOT modify it by hand as your changes will get lost the next time it is gener
 # Report template
 
 ```ruby
-# frozen_string_literal: true
-
```
JonRowe: Can we leave this in whilst turning off the cop? We don't have it for "reasons" but agree with it in principle

spec/rspec/matchers/composable_spec.rb
```diff
@@ -104,7 +104,7 @@ module Matchers
             begin
               block.call
               false
-            rescue Exception
+            rescue SystemExit
```
JonRowe: Just disable the cop, don't change this

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -709,7 +709,7 @@ class PseudoHash < SimpleDelegator
 
           domain = Struct.new(:domain) do
             def matches?(url)
-              URI(url).host == self.domain
```
JonRowe: If you're going to remove this change the shadowing above.

JonRowe: Actually honestly I prefer the self, can we disable the cop instead?

.rubocop.yml
```diff
@@ -4,6 +4,9 @@ inherit_from:
 AllCops:
   TargetRubyVersion: 2.1
   DisplayCopNames: true
+  Exclude:
+    - bin/*
```
JonRowe: Bin is bundler binstubs FYI

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -301,7 +301,7 @@ def to_s
 
   it "fails if another error is raised (NameError)" do
     expect {
-      expect { raise RuntimeError, "example message" }.to raise_error(NameError)
+      expect { raise "example message" }.to raise_error(NameError)
```
JonRowe: You can suppress this cop, I think its better explicit

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -31,9 +31,9 @@ def ok
 
   it "clears user instance variables between invocations" do
     RSpec::Matchers.define(:be_just_like) do |expected|
-      match do |actual_value|
+      match do |actual|
         @foo ||= expected
-        @foo == actual_value
+        @foo == actual
```
JonRowe: do we need to revert this?

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -591,11 +591,11 @@ def foo
         end
 
         specify "`match?` returns true if the wrapped expectation passes" do
-          expect(self.matcher.matches?('value')).to be_truthy
+          expect(matcher.matches?('value')).to be_truthy
         end
 
         specify "`match?` returns false if the wrapped expectation fails" do
-          expect(self.matcher.matches?('other value')).to be_falsey
+          expect(matcher.matches?('other value')).to be_falsey
```
JonRowe: Again I feel `self.` is more suitable, can we supress the rule instead

spec/rspec/matchers/dsl_spec.rb
pirj: The [impact on `rspec-mocks`](https://github.com/rspec/rspec-mocks/commit/3b909ed1a951bbca340ea98c27ab65da7f43881c) was definitely more significant than [here](https://github.com/rspec/rspec-expectations/pull/703/commits/fdd9ef3862c4fa09acef0b369ecad8eb5de72dee).
`actual` is the original naming, and Myron [mentioned](https://bugs.ruby-lang.org/issues/10661) that "changed many variable and method names to names I do not like as much". Since the Ruby warning is no more, I believe it's ok to bring the code to its original state.

I've changed the names along with the removal of redundant `self`s that were introduced in the same commit. WDYT if I revert the changes to variable names in this pull request and address this in a follow-up?

spec/rspec/matchers/dsl_spec.rb
pirj: Sorry, not exactly sure what you mean. Is `matcher` being referred to in `self.matcher` the one defined in a `let`?

spec/rspec/matchers/dsl_spec.rb
pirj: Got it, makes total sense.

.rubocop.yml
pirj: Will open an issue.

.rubocop.yml
JonRowe: Ah ok, I think I would prefer none.

.rubocop.yml
JonRowe: Honestly I think I would prefer this rule turned off then.

spec/rspec/matchers/built_in/change_spec.rb
JonRowe: Ha ok

REPORT_TEMPLATE.md
pirj: Went this way:
 - enforced to never add `frozen_string_literal`
 - excluded this doc file from the check

pirj:
@benoittgt Thanks for the review.
Unfortunately, `Style/RedundantSelf` is not fixed in latest RuboCop, I temporarily updated to 0.65, it still detects where it shouldn't, and doesn't detect where it should, so I hope to re-evaluate and remove local disabling when it's fixed in RuboCop.

.rubocop.yml
pirj: Oh, just noticed that this setting is coming [from `rspec-dev`'s `.rubocop_rspec_base.yml`](https://github.com/rspec/rspec-dev/blob/master/travis/.rubocop_rspec_base.yml#L13).
Disabled this cop for this repository.

pirj:
@JonRowe Addressed a couple more remarks.

Appreciate if you check out a few of the remaining questions that I have: [1],(https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263993069), [2](https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263993910), [3](https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263994098).

pirj:
@JonRowe I'm up for changing the code either way, please let me know what you think.

JonRowe:
@pirj Sorry I'm very busy at the moment, I promise this is high on my list to re-review I very much appreciate your time taken here!

pirj:
@JonRowe A friendly reminder. There are just three questions left ([1](https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263993069), [2](https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263993910), [3](https://github.com/rspec/rspec-expectations/pull/1104#discussion_r263994098)), I'm up to change it the way you see fit.

My further plan is to:
- perform fix the offences in the other RSpec repositories
- update RuboCop version to the latest
- extract common parts to `rspec-dev`'s common `.rubocop.yml`

I'll stick to the practice I used here with the other repositories - reduce the code churn and disable the rules that would require massive codebase change.
This might take a while considering the effort to review those changes, I hope for your support in those changes.

spec/rspec/matchers/built_in/include_spec.rb
JonRowe: I revert to my original then, remove the shadowing!

spec/rspec/matchers/dsl_spec.rb
JonRowe: I trust your judgement here

spec/rspec/matchers/dsl_spec.rb
JonRowe: No, not sure what I meant here.

JonRowe:
Just a fix required to your "1" and a green build and this can be merged. Thanks so much for your patience here @pirj

spec/rspec/matchers/built_in/include_spec.rb
pirj: This is an interesting example. I thought of renaming one of the variables, but then my OCD chimed in, and I decided to check if that object *is* treated as a matcher when `failure_message` is defined.

A balancing example:
```
      it 'does treats an object that implements #matches? and failure_message as a matcher' do
        matcher = Struct.new(:value) do
          def matches?(_)
            fail "`matches?` should never be called"
          end

          def failure_message
            'oops'
          end
        end

        expect([matcher.new("rspec.info")]).to include(matcher.new("rspec.info"))

        expect {
          expect([matcher.new("rspec.info")]).to include(matcher.new("foo.com"))
        }.to fail_matching("expected [#{matcher.new("rspec.info").inspect}] to include")
      end
    end
```
and it was still green. This led me to a thought that its `matches?` was never called.
The trail led me to [`values_match?`](https://github.com/rspec/rspec-expectations/blob/a85a42c/lib/rspec/matchers/composable.rb#L67), then to `Matchers.is_a_matcher?`, which returns `true`, and then to [`values_match?`](https://github.com/rspec/rspec-support/blob/cb038de/lib/rspec/support/fuzzy_matcher.rb#L18) that does the matching using `===`, that in turn is only [aliased to `matches?`](https://github.com/rspec/rspec-expectations/blob/a85a42c/lib/rspec/matchers/composable.rb#L45) when `Composable` is included to the matcher.

I played around with [`all_of` argument matcher](https://github.com/toptal/rspec-any_of) recently, tried to implement it as a class rather than to use `RSpec::Matchers.define` DSL, and failed miserably. Just like here, `matches?` never gets called, but `===` does. It felt that there are some differences between how generic and composable matchers can be defined, and I didn't completely understand it. Neither I do now, but at least I've found that bit that `Composable` should be included in argument matchers.

From my point of view, this example is misleading. Am I missing something?
Do you mind if I remove it altogether?

JonRowe: Its a regression check of sorts, so it needs to remain, I don't mind if you want to refactor it to be less confusing. This is testing composability of the include matcher and that the object is not treated as a further argument matcher in a nested format, not the matcher definition of the object itself.

pirj: Done 👍

pirj:
@JonRowe [Addressed](https://github.com/rspec/rspec-expectations/pull/1104/commits/8d877a8cb5cb426c3252959186c6a4f2706bd563) the remaining issue. Please take another look.

benoittgt:
I reviewed last commits. LGTM. Thanks for the hard work @pirj

pirj:
@JonRowe Please take another look.

JonRowe:
Thanks @pirj !

pirj:
Thanks for accepting @JonRowe !

PR #1107, opened 2019-04-01
Added warning message for #1105

taichi-ishitani:
Hi,

I added warning message for #1105 as a special failure message.

PR #1112, opened 2019-04-23
"KindOf" matchers: Rescue `NoMethodError` instead of `respond_to?` check

### LINKED ISSUE #1111: #1107 Somehow breaks my specs
### Subject of the issue
I'm getting the following error on perfectly normal object:
```
ArgumentError:
       The be_a_kind_of matcher requires that the actual object responds to either #kind_of? or #is_a? methods but it responds to neigher of two methods.
```

### Environment
* Ruby version: 2.5.1
* rspec-expectations version: 3.8.3

### Steps to reproduce
I do not really know yet, I will investigate more on it and provide details.

### Actual behavior
```ruby
value.respond_to?(:is_a?)
#=> true
::Kernel.instance_method(:respond_to?).bind(value).call(:is_a?)
#=> false
```

P.S an error message has a typo - `neigher`.

FunkyloverOne:
Resolves #1111

### DIFF
--- a/lib/rspec/matchers/built_in/be_instance_of.rb
+++ b/lib/rspec/matchers/built_in/be_instance_of.rb
@@ -14,8 +14,9 @@ module RSpec
       private
 
         def match(expected, actual)
-          if actual_object_respond_to?(actual, :instance_of?)
-            actual.instance_of?(expected)
+          result = nil
+          if executable? { result = actual.instance_of?(expected) }
+            result
           else
             raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
                                    "the actual object responds to #instance_of? method " \
@@ -23,8 +24,11 @@ module RSpec
           end
         end
 
-        def actual_object_respond_to?(actual, method)
-          ::Kernel.instance_method(:respond_to?).bind(actual).call(method)
+        def executable?
+          yield
+          true
+        rescue NoMethodError
+          false
         end
       end
     end
--- a/lib/rspec/matchers/built_in/be_instance_of.rb
+++ b/lib/rspec/matchers/built_in/be_instance_of.rb
@@ -14,21 +14,11 @@ module RSpec
       private
 
         def match(expected, actual)
-          result = nil
-          if executable? { result = actual.instance_of?(expected) }
-            result
-          else
-            raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
-                                   "the actual object responds to #instance_of? method " \
-                                   "but it does not respond to the method."
-          end
-        end
-
-        def executable?
-          yield
-          true
+          actual.instance_of?(expected)
         rescue NoMethodError
-          false
+          raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
+                                 "the actual object responds to #instance_of? method " \
+                                 "but a `NoMethodError` was encountered instead."
         end
       end
     end
--- a/lib/rspec/matchers/built_in/be_kind_of.rb
+++ b/lib/rspec/matchers/built_in/be_kind_of.rb
@@ -8,10 +8,11 @@ module RSpec
       private
 
         def match(expected, actual)
-          if executable?(actual, :kind_of?, expected)
-            actual.kind_of?(expected)
-          elsif executable?(actual, :is_a?, expected)
-            actual.is_a?(expected)
+          result = nil
+          if executable? { result = actual.kind_of?(expected) }
+            result
+          elsif executable? { result = actual.is_a?(expected) }
+            result
           else
             raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
                                    "the actual object responds to either #kind_of? or #is_a? methods " \
@@ -19,8 +20,8 @@ module RSpec
           end
         end
 
-        def executable?(object, method, *args)
-          object.public_send(method, *args)
+        def executable?
+          yield
           true
         rescue NoMethodError
           false
--- a/lib/rspec/matchers/built_in/be_kind_of.rb
+++ b/lib/rspec/matchers/built_in/be_kind_of.rb
@@ -8,19 +8,23 @@ module RSpec
       private
 
         def match(expected, actual)
-          if actual_object_respond_to?(actual, :kind_of?)
-            actual.kind_of?(expected)
-          elsif actual_object_respond_to?(actual, :is_a?)
-            actual.is_a?(expected)
+          result = nil
+          if executable? { result = actual.kind_of?(expected) }
+            result
+          elsif executable? { result = actual.is_a?(expected) }
+            result
           else
             raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
                                    "the actual object responds to either #kind_of? or #is_a? methods "\
-                                   "but it responds to neigher of two methods."
+                                   "but it responds to neither of two methods."
           end
         end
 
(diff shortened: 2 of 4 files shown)


FunkyloverOne:
CI checks for JRuby seem broken.

benoittgt:
Thanks for the PR @FunkyloverOne 

Sorry @FunkyloverOne it is probably related to https://github.com/rvm/rvm/issues/4675

I just re-ran JRuby builds

lib/rspec/matchers/built_in/be_instance_of.rb
```diff
@@ -14,17 +14,21 @@ def description
       private
 
         def match(expected, actual)
-          if actual_object_respond_to?(actual, :instance_of?)
-            actual.instance_of?(expected)
+          result = nil
+          if executable? { result = actual.instance_of?(expected) }
```
benoittgt: I am **personally** not a huge fan of using a block in that case. What about passing args to the method and simply rescue `NoMethodError`.
 
This a not a blocker to accept the PR. I am waiting for @JonRowe's final review. 🙌🏻

benoittgt:
CI seems better. Thanks for the PR. 🙂

LGTM even with my comment.

lib/rspec/matchers/built_in/be_instance_of.rb
FunkyloverOne: Yeah, I thought about it, but it just seemed like I basically have to pass everything to that method. Because the method will end up being something like this:
```ruby
def executable?(object, method, *args)
  object.public_send(method, *args)
  true
rescue
  false
end
```

I don't know, I just liked that block trick better :)

JonRowe: This can be simplified to:

```
def match(expected, actual)
  actual.instance_of?(expected)
rescue NoMethodError
  raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
                         "the actual object responds to #instance_of? method " \
                         "but a `NoMethodError` was encountered instead."
end
```

The block trick adds execution time

JonRowe: Remember we're just trying to surface a friendlier message to those objects not supporting a standard Ruby interface.

FunkyloverOne: @JonRowe  are we talking about **only** that case with `#instance_of?` check, right? So I'm leaving the block trick in `#is_a?`/`#kind_of?` check, right? Or it's still better to replace it with a method there?

FunkyloverOne:
OK guys, I realize that performance is a priority here, so I decided to run a benchmark, here:
```ruby
require 'benchmark/ips'

def blocky(expected, actual)
  def match(expected, actual)
    result = nil
    if blocky_executable? { result = actual.kind_of?(expected) }
      result
    elsif blocky_executable? { result = actual.is_a?(expected) }
      result
    else
      raise ArgumentError
    end
  end
end

def non_blocky(expected, actual)
  if executable?(actual, :kind_of?, expected)
    actual.kind_of?(expected)
  elsif executable?(actual, :is_a?, expected)
    actual.is_a?(expected)
  else
    raise ArgumentError
  end
end

def blocky_executable?
  yield
  true
rescue NoMethodError
  false
end

def executable?(object, method, *args)
  object.public_send(method, *args)
  true
rescue NoMethodError
  false
end

Benchmark.ips do |x|
  x.report('blocky') { blocky(Integer, 1) }
  x.report('non_blocky') { non_blocky(Integer, 1) }
  x.compare!
end
```
```
Warming up --------------------------------------
              blocky   144.418k i/100ms
          non_blocky   244.396k i/100ms
Calculating -------------------------------------
              blocky      2.538M (± 1.0%) i/s -     12.709M in   5.007300s
          non_blocky      3.845M (± 3.5%) i/s -     19.307M in   5.027727s

Comparison:
          non_blocky:  3844958.7 i/s
              blocky:  2538323.9 i/s - 1.51x  slower
```
So I will also update that `BeAKindOf` matcher as well.

FunkyloverOne:
@benoittgt, @JonRowe, it's ready now.

FunkyloverOne:
Oh, it's not, this "non blocky version"  breaks some checks. I guess I have to revert that.

FunkyloverOne:
@JonRowe, @benoittgt, ping

p-mongo:
This PR fixes the Mongoid test suite broken by #1107.

benoittgt:
Maybe this could be addressed too in the same PR? https://github.com/rspec/rspec-expectations/issues/1114

lib/rspec/matchers/built_in/be_kind_of.rb
```diff
@@ -8,19 +8,23 @@ class BeAKindOf < BaseMatcher
...
+          elsif executable? { result = actual.is_a?(expected) }
+            result
           else
             raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
-                                   "the actual object responds to either #kind_of? or #is_a? methods "\
-                                   "but it responds to neigher of two methods."
+                                   "the actual object responds to either #kind_of? or #is_a? methods " \
+                                   "but a `NoMethodError` was encountered instead."
           end
         end
```
JonRowe: The original use of this matcher just supported `kind_of` so I propose we go back to that with a simple rescue.

```
def match(expected, actual)
  actual.kind_of? expected
rescue NoMethodError
  raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
                         "the actual object responds to #kind_of? method " \
                         "but a `NoMethodError` was encountered instead."
end
```

FunkyloverOne:
Done.

benoittgt:
Thanks. LGTM

pirj:
I might be mistaken or miss something obvious, but it seems to me that there's a mistake in the benchmark, a nested method definition:
```
def blocky(expected, actual)
  def match(expected, actual)
```

After removing this nesting `blocky` is 2.x faster instead of being 1.77x slower.

FunkyloverOne:
@pirj OMG, how could I miss that :laughing:

JonRowe:
To be honest you also need to consider the happy path is likely to never invoke the rescue, where as the `if` would always be called.

nijikon:
@JonRowe can we release a new version? The previous implementation breaks our specs.

pirj:
@nijikon Would you accept [installing the gem from git](https://bundler.io/v2.0/guides/git.html) as a temporary solution?

```
group :test do
 gem 'rspec-expectations', git: 'https://github.com/rspec/rspec-expectations', branch: 'master'
end
```

nijikon:
It will, but I consider this a bug, so there should be a bugfix release 😄

benoittgt:
@nijikon We will. @JonRowe I can do it if you want.

danielpietzsch:
Would be great to get this released! I’m using this with the “rspec-rails” gem and could get the specs working again by (temporarily) installing all the gems from the master branch, [as described in the “rspec-rails” readme](https://github.com/rspec/rspec-rails#installation):

```ruby
%w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
  gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => 'master'
end
```

JonRowe:
Theres another PR I'd like to get merged before releasing, thanks for your patience!

JonRowe:
Released as 3.8.4 due to inactivity on aforementioned other PR.

PR #1116, opened 2019-05-24
Prevent Aliased Matchers from Overriding Expected Data

lnestor:
## Previous Behavior

The aliased matcher implementation of a description currently replaces all instance of the old matcher's name in the new matcher's name. If it happens that the expected values include the old matcher's name, these also get overwritten. In the examples below, the string `"include"` is replaced in both the outputs by the new matcher's name.

```ruby
it 'overwrites data for a_string_including' do
  my_string = "a string with 'include'"
  expect("some random string").to match(a_string_including(my_string))
end

# => expected "some random string" to match 
#       (a string including "a string with 'a string including'")
```

```ruby
it 'overwrites data for a_hash_including' do
  my_hash = { my_string: "a string with 'include'" }
  expect({}).to match(a_hash_including(my_hash))
end

# => expected {} to match 
#       (a hash including {:my_string => "a string with 'a hash including'"})
```

## New Behavior

This change makes it so it only replaces the first instance of the original name so none of the expected data is also overwritten. The two above error outputs would be the following:

`expected {} to match (a hash including {:my_string => "a string with 'include'"})`
`expected "some random string" to match (a string including "a string with 'include'")`

I saw this in https://github.com/rspec/rspec-rails/issues/1835.

### DIFF
--- a/lib/rspec/matchers/aliased_matcher.rb
+++ b/lib/rspec/matchers/aliased_matcher.rb
@@ -44,7 +44,7 @@ module RSpec
       #
       # @api private
       def failure_message
-        @description_block.call(super, replace_after_start: true)
+        @description_block.call(super, true)
       end
 
       # Provides the failure_message_when_negated of the aliased matcher. Aliased matchers
@@ -54,7 +54,7 @@ module RSpec
       #
       # @api private
       def failure_message_when_negated
-        @description_block.call(super, replace_after_start: true)
+        @description_block.call(super, true)
       end
     end
 
@@ -105,7 +105,7 @@ module RSpec
       def optimal_failure_message(same, inverted)
         if DefaultFailureMessages.has_default_failure_messages?(@base_matcher)
           base_message = @base_matcher.__send__(same)
-          overriden    = @description_block.call(base_message, replace_after_start: true)
+          overriden    = @description_block.call(base_message, true)
           return overriden if overriden != base_message
         end
 
@@ -127,7 +127,7 @@ module RSpec
       # Creates the overriden description.
       #
       # @api private
-      def call(old_desc, replace_after_start: false)
+      def call(old_desc, replace_after_start=false)
         if @override_block
           @override_block.call(old_desc)
         else
--- a/lib/rspec/matchers/aliased_matcher.rb
+++ b/lib/rspec/matchers/aliased_matcher.rb
@@ -44,7 +44,7 @@ module RSpec
       #
       # @api private
       def failure_message
-        @description_block.call(super)
+        @description_block.call(super, replace_after_start: true)
       end
 
       # Provides the failure_message_when_negated of the aliased matcher. Aliased matchers
@@ -54,7 +54,7 @@ module RSpec
       #
       # @api private
       def failure_message_when_negated
-        @description_block.call(super)
+        @description_block.call(super, replace_after_start: true)
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/rspec/matchers/dsl.rb
```diff
@@ -29,7 +29,7 @@ module DSL
       # @see RSpec::Matchers
       def alias_matcher(new_name, old_name, options={}, &description_override)
         description_override ||= lambda do |old_desc|
-          old_desc.gsub(EnglishPhrasing.split_words(old_name), EnglishPhrasing.split_words(new_name))
+          old_desc.sub(EnglishPhrasing.split_words(old_name), EnglishPhrasing.split_words(new_name))
```
JonRowe: I think this needs to only match the start of the string, this might still accidentally override terms if the original description is not generated.

spec/rspec/matchers/aliased_matcher_spec.rb
```diff
@@ -79,6 +79,19 @@ def description
...
+        def description
+          "my repeating base matcher my repeating base matcher"
+        end
+      end
+
+      RSpec::Matchers.alias_matcher :my_repeating_override, :my_repeating_base_matcher
+
+      it 'does not override data in the description with the same name as the matcher' do
+        matcher = my_repeating_override
+        expect(matcher.description).to eq("my repeating override my repeating base matcher")
```
JonRowe: It'd prefer a practical example showing the problem using `fail_with`, even the one from rspec/rspec-rails#1835 would suffice.

JonRowe (changes requested):
Thanks for tackling this, a bit of feedback above!

lib/rspec/matchers/dsl.rb
lnestor: Thanks for the feedback! Can you clarify this a bit? `#sub` matches the first occurrence, so are you saying something like matching the first n characters of the string?

Also, what do you mean by if the original description is not generated? Since `BaseMatcher` implements description, I would think that they all would have a description. Unless a class implements its own description which does not contain the name of the matcher.

JonRowe: > Can you clarify this a bit? #sub matches the first occurrence, so are you saying something like matching the first n characters of the string?

`"hello_old_name".sub("old_name","new_name")` would produce `"hello_new_name"` but we don't want that. We only want to replace `old_name` as it was generated initially. 

> Also, what do you mean by if the original description is not generated? Since BaseMatcher implements description, I would think that they all would have a description. 

Similarly it is possible for matchers to implement `description` themselves and it might mean they don't contain the old_name at all, in this case we shouldn't be replacing the first instance blindly as it might not be in the description but in the expected.

lib/rspec/matchers/aliased_matcher.rb
```diff
@@ -105,12 +105,53 @@ def failure_message_when_negated
...
+      def default_description(old_desc, replace_after_start)
+        old_name_split = EnglishPhrasing.split_words(@old_name)
+
+        if old_desc.start_with?(old_name_split) || replace_after_start
+          old_desc.sub(old_name_split, EnglishPhrasing.split_words(@new_name))
+        else
+          old_desc
+        end
+      end
+    end
```
JonRowe: This is a lot of added complication, and still doesn't remove the global sub in favour of a "start of string only" replace.

pirj:
Do you have any questions on how to proceed with this @lnestor ?

pirj:
@lnestor Are you interested in wrapping this up?

JonRowe:
Migrated to the monorepo as rspec/rspec#123

PR #1121, opened 2019-06-24
Add Expectations namespace to ExpectationTarget tags.

castwide:
The `@!method` tag for `RSpec::Matchers#expect` references `ExpectationTarget`, but that class is not exposed there. `Expectations::ExpectationTarget` resolves correctly.


JonRowe:
Thanks!

Issue #805, opened 2015-06-15
is_expected with block expectations


xaviershay:
The following situation just confused someone here at work:

```
describe 'is_expected with block expectation' do
  subject { raise }

  it { is_expected.to raise_error }
end
```

That reads like it should work, but doesn't because `is_expected` is defined as `expect(subject)`.

I don't typically use `is_expected`, so not sure if there is a canonical way to do this already. It isn't listed at https://www.relishapp.com/rspec/rspec-core/docs/subject/one-liner-syntax

1. Is there a "correct way" to do this?
2. Is one-liner syntax something that we (as a core team) use much, or is it in the `expect_any_instance_of` bucket of extras?
3. Should we add one-liner Relish documentation about using block matchers? I'm thinking yes.
4. Is there a way to fix this transparently, and if yes, should we? My gut says no because it would be switching `is_expected` implementation based on matcher used and that seems gnarly.

Tangentially related discussion thread on separating block vs value expectations, in which we didn't discuss this case: https://github.com/rspec/rspec-expectations/issues/526

cc @rspec/rspec

JonRowe:
> 1. Is there a "correct way" to do this?

Not with the one liner syntax that I'm aware of.

> 2. Is one-liner syntax something that we (as a core team) use much, or is it in the expect_any_instance_of bucket of extras?

It's in the same bucket, I certainly don't recommend it.

> 3. Should we add one-liner Relish documentation about using block matchers? I'm thinking yes.

We should mention they're not supported.

> 4. Is there a way to fix this transparently, and if yes, should we? My gut says no because it would be switching is_expected implementation based on matcher used and that seems gnarly.

this has come up before and we shied away from it because the only way to really do it is to pass a block out of the subject block and then conditionally detect what to do with it, the last conversation we had around this resulted in the original issue raiser producing puyo/rspec-subject_call

myronmarston:
> Not with the one liner syntax that I'm aware of.

There is, if you're willing to make your `subject` a lambda:

``` ruby
describe 'is_expected with block expectation' do
  subject { -> { raise } }

  it { is_expected.to raise_error }
end
```

...that's pretty obtuse and not something I'd recommend, generally.

> Is one-liner syntax something that we (as a core team) use much, or is it in the `expect_any_instance_of` bucket of extras?

[...]

PR #1125, opened 2019-08-05
Prevent block-only matchers from being used with value expectation target

pirj:
The purpose of this (rather experimental) change is to restrict the usage of:

```ruby
subject { -> { do_something } }
it { is_expected.to change(something).to(new_value) }
```

This syntax was referenced to as:

> pretty obtuse and not something I’d recommend, generally.
https://github.com/rspec/rspec-expectations/issues/805#issuecomment-112239820

> We should mention they’re not supported.
https://github.com/rspec/rspec-expectations/issues/805#issuecomment-112238533

### Problems with this approach

Specs fail quite badly mostly due to the use of this syntax in RSpec Expectations' own specs:

```ruby
RSpec.describe RSpec::Matchers::BuiltIn::ChangeToValue do
  k = 0
  before { k = 0 }
  it_behaves_like "an RSpec matcher", :valid_value => lambda { k = 2 },
                                      :invalid_value => lambda { k = 3 },
                                      :disallows_negation => true do
    let(:matcher) { change { k }.to(2) }
  end
end
```

Would you accept this if I change those specs to something more canonical? E.g. like:

```ruby
RSpec.describe RSpec::Matchers::BuiltIn::ChangeToValue do
  let(:k) { 0 }
  it_behaves_like "an RSpec matcher" do
    let(:valid_value) { k += 1 }
    let(:invalid_value) { }

    def matcher
      change { k }.from(0)
    end
  end
end
```

The above change didn't work from the first time, but I'll keep trying if that's the right direction.

### Related Links

https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html
https://www.reddit.com/r/ruby/comments/cejl3q/call_for_discussion_rspec_implicit_block
https://github.com/rubocop-hq/rspec-style-guide/issues/76


lib/rspec/matchers/built_in/change.rb
```diff
@@ -72,11 +72,14 @@ def description
           "change #{change_details.value_representation}"
         end
 
-        # @private
         def supports_block_expectations?
           true
         end
 
+        def supports_value_expectations?
```
pirj: It's possible to tell if a matcher supports block expectations, but it's not possible to detect that it doesn't support value expectations, and will be sending `call` on a value target.

lib/rspec/matchers/built_in/change.rb
```diff
@@ -72,11 +72,14 @@ def description
           "change #{change_details.value_representation}"
         end
 
-        # @private
```
JonRowe: If you want to make a method public you must document it.

lib/rspec/matchers/built_in/change.rb
JonRowe: Seems reasonable!

spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -262,13 +262,19 @@ def expect_block
 
     describe "expect(...).to matcher.and(other_matcher)" do
 
-      it_behaves_like "an RSpec matcher", :valid_value => 3, :invalid_value => 4, :disallows_negation => true do
+      it_behaves_like "an RSpec value matcher" do
```
pirj: Now that value and blocks specs are separate I think that for "an RSpec value matcher" it's possible to keep passing the arguments via shared example, not using `let`'s.
It will also reduce the size of this pull request.

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def invalid_expectation
+    expect { |block| invalid_block(&block) }
+  end
+
+  include_examples "an RSpec matcher"
+
+  it 'preserves the symmetric property of `==`' do
+    expect(matcher).to eq(matcher)
+    # TODO: the following expectations fail due to
```
pirj: @JonRowe I have two open questions on how to deal with the following examples.

Apart from this one and the pair below everything is green.

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,102 @@
...
+    #    RuntimeError:
+    #      Warnings were generated: foo
+    #
+    # expect(matcher).not_to eq(valid_block)
+    # expect(valid_block).not_to eq(matcher)
+    #
+    # Does it make sense for block matchers?
+  end
+
+  # TODO: Two options here.
```
pirj: The second set of examples I some have doubts.

pirj:
I'm still up to adjust this the way you see fit but need some guidance on specs.

spec/spec_helper.rb
```diff
@@ -5,8 +5,10 @@
   minimum_coverage 97
 end
 
-Dir['./spec/support/**/*'].each do |f|
-  require f.sub(%r|\./spec/|, '')
+Dir.chdir('spec') do
```
JonRowe: Why?

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,102 @@
...
+  include_examples "an RSpec matcher"
+
+  it 'preserves the symmetric property of `==`' do
+    expect(matcher).to eq(matcher)
+    # TODO: the following expectations fail due to
+    #
+    # 1) output.to_stderr matcher behaves like an RSpec block-only matcher preserves the symmetric property of `==`
+    #    Failure/Error: raise "Warnings were generated: #{output}" if has_output?
+    #    RuntimeError:
+    #      Warnings were generated: foo
```
JonRowe: So this is because warnings are generated by the specs, it should be possible to rewrite the relevant specs warning free...

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,102 @@
...
+  #   expect(matcher).not_to be === ->(block) { invalid_block(&block) }
+  # end
+  #
+  # I see two options:
+  # 1. Introduce another conditional flag, and run the former two examples
+  #    for all specs except yield_spec.rb. Also pull those two latter to
+  #    yield_spec.rb itself.
+  # 2. Remove this. I'm not too confident with the code base to tell if it's
+  #    absolutely necessary to check this, and why `===` is vital for
+  #    composing matchers, and if it's applicable to block matchers.
```
JonRowe: Its applicable to all matchers, its the protocol for them.

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,102 @@
...
+  # end
+  #
+  # it 'does not match an invalid block when using #=== so it can be composed' do
+  #   expect(matcher).not_to be === ->(block) { invalid_block(&block) }
+  # end
+  #
+  # I see two options:
+  # 1. Introduce another conditional flag, and run the former two examples
+  #    for all specs except yield_spec.rb. Also pull those two latter to
+  #    yield_spec.rb itself.
```
JonRowe: How about `->(*args) {  args.length == 1 && args[0] === Proc ? invalid_block(&args[0]) : invalid_block }` with a note explaining why? How did these pass before? Different inputs?

spec/spec_helper.rb
pirj: I've split shared examples into several files and placed them under `spec/support/shared_examples` directory. `Dir` matched both files and the directory itself.
I couldn't resist the temptation for this change but will minimize it so `Dir` only yields `.rb` files

spec/support/shared_examples/block_matcher.rb
pirj: Ended up using `proc` that works in both cases.
Added a note.

spec/support/shared_examples/block_matcher.rb
pirj: Exactly the same problem as below it seems. `proc` fixed it.

spec/rspec/matchers/built_in/compound_spec.rb
pirj: +512 -> +444, less churn.

spec/support/shared_examples/block_matcher.rb
pirj: Previously, valid values were defined as e.g.:
```ruby
  it_behaves_like "an RSpec matcher", :valid_value => lambda { k += 1 },
```
and that were supposed to check that
```ruby
expect { k += 1 }.to change { k }
```
The `matcher` here is initialized by `change { k }`, and the target is a block.

So the whole point of those two expectations is to make sure that `==` on both the lambda and the matcher return `false` when compared with each other.

After this change the target block is defined using a method, and the reason for warnings to appear is that those `output`/`throw_symbol`/`raise_error` targets are actually get called when target is compared, since they are not a block.

Do you think that writing those examples as the following will achieve the same goal?
```ruby
    expect(matcher).not_to eq(-> { valid_block })
    expect(-> { valid_block }).not_to eq(matcher)
```

pirj:
@JonRowe Builds are green, all specs are in place.
It took a while to work around Ruby 1.8.7 block signature strictness, moved stuff around.
No rush, please take another look when you have a moment.

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,80 @@
+RSpec.shared_examples "an RSpec block-only matcher" do |*options|
+  # Note: Ruby 1.8 expects you to call a block with arguments if it is
+  # declared that accept arguments. In this case, some of the specs
+  # that include examples from this shared example group do not pass
+  # arguments. A workaround is to use splat and pick the first argument
+  # if it was passed.
+  options = options.first || {}
```
pirj: Another approach would be to always add an empty hash as an argument when including this shared example group, both are workarounds, but this one is less scattered.

Had to use the same trick in `yield_spec.rb`.

spec/support/shared_examples/block_matcher.rb
```diff
@@ -39,8 +58,8 @@ def invalid_expectation
 
   # NOTE: The four examples below make use of `proc` due to its tolerance
   # to missing arguments. This is required to work with both matchers that
-  # pass arguments. E.g. `yield` passes a probe to the block, while other
-  # matchers don't.
+  # pass arguments and not, e.g. `yield` passes a probe to the block,
```
JonRowe: `and not`?

spec/support/shared_examples/block_matcher.rb
JonRowe: Can you not just do `|options = {}|` or does that not work on 1.8.7?

pirj: Yes, the only reason for this is 1.8.7. Couldn't figure out a better approach.

pirj:
Squashed commits.

pirj:
@JonRowe Do you think this needs some adjustment?

JonRowe:
> @JonRowe Do you think this needs some adjustment?
I promise this is high on my list for a re-review!

pirj:
Thanks!

I'm holding off from merging this, found a problem with `rspec-collection_matchers`, still getting to the bottom of it.
```
     NoMethodError:
       undefined method `supports_value_expectations?' for #<#<Class:0x00007fb20c0a7128>:0x00007fb20c0a7088>
     # ./lib/rspec/collection_matchers/have.rb:71:in `determine_collection'
     # ./lib/rspec/collection_matchers/have.rb:41:in `matches?'
     # ./spec/rspec/collection_matchers/have_spec.rb:532:in `block (4 levels) in <top (required)>'
```

lib/rspec/expectations/expectation_target.rb
```diff
@@ -118,9 +152,11 @@ def enforce_block_expectation(matcher)
       end
 
       def supports_block_expectations?(matcher)
-        matcher.supports_block_expectations?
-      rescue NoMethodError
-        false
+        if matcher.respond_to?(:supports_block_expectations?)
```
pirj: This change is due to `have` matcher from `rspec-collection_matchers` that is sensitive to methods called on it, i.e. if we call `supports_value_expectations?` on `have(3).players`, it starts to think that it should match `supports_value_expectations?` association, not `players`.

The problem is with `supports_value_expectations?`, changing `supports_block_expectations?` for consistency.

mvz:
Just found this change and it also breaks the following code. Not sure if that was intended:

```ruby
      call = lambda do
        Reek::CLI::Silencer.silently do
          described_class.new ['--foo']
        end
      end
      expect(call).to raise_error(SystemExit) do |error|
        expect(error.status).to eq Reek::CLI::Status::DEFAULT_ERROR_EXIT_CODE
      end
```
```

**Comment by pirj** [comment]
@mvz Would you accept the code written as:
```ruby
     expect do
       Reek::CLI::Silencer.silently do
         described_class.new ['--foo']
       end
     end.to raise_error(SystemExit) do |error|
       expect(error.status).to eq Reek::CLI::Status::DEFAULT_ERROR_EXIT_CODE
     end
```

It's also possible to extract the lambda to a method and use `expect { silencer_configured_with_foo_option }.to raise_error ...`.

mvz:
@pirj Yes, both are fine options, although I would prefer the second one rather than chaining multi-line blocks. I was just surprised my existing specs started failing.

I fixed my code by changing `expect(call)` to `expect(&call)`.

JonRowe:
This PR has been reverted for now.

JonRowe:
I would like to include this again, but we need to address the following situations:

* Transitionally allow blocks to be in-explicitly passed in for future deprecation and warning. This is quite easy to achieve but has knock on effects for the next thing I'd like to address.
* Still allow passed in blocks to be introspected as values.

pirj:
@JonRowe Sounds good. I can tackle this in the near future.
What I intend to do is to warn in `enforce_value_expectation` instead of `fail`ing.
Not sure I completely understand what you mean by block introspection, can you please elaborate a bit?

JonRowe:
Things like `expect(-> { }).to be_a(Proc)`

PR #1126, opened 2019-08-19
Notice when object implement inspect but == is different

benoittgt:
In few cases the object tested implement `inspect` but the eq matcher
will see a difference when doing ==.

This can lead to a misleading output like this one:

```
Failures:

  1) Foo confuses users with an empty diff
     Failure/Error: Foo.something(Foo.new)

       #<Foo (class)> received :something with unexpected arguments
         expected: ("foobar")
              got: ("foobar")
       Diff:

     # ./spec/foo_spec.rb:13:in `block (2 levels) in <top (required)>'
```

This change add a notice to help to understand this failed expectation.

Fix: https://github.com/rspec/rspec-support/issues/274

### DIFF
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -69,7 +69,12 @@ module RSpec
         @expected_list.map do |(expected, diff_label)|
           diff = differ.diff(actual, expected)
           next if diff.strip.empty?
-          "#{diff_label}#{diff}"
+          if diff == "\e[0m\n\e[0m"
+            "#{DEFAULT_DIFF_LABEL}\n" +
+              "  <diff is empty, do your objects produce identical inspect output?>"
+          else
+            "#{diff_label}#{diff}"
+          end
         end.compact.join("\n")
       end
     end
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -70,7 +70,7 @@ module RSpec
           diff = differ.diff(actual, expected)
           next if diff.strip.empty?
           if diff == "\e[0m\n\e[0m"
-            "#{diff_label}\n" +
+            "#{diff_label}\n" \
               "  <The diff is empty, are your objects producing identical `#inspect` output?>"
           else
             "#{diff_label}#{diff}"
--- a/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
+++ b/lib/rspec/matchers/expecteds_for_multiple_diffs.rb
@@ -70,8 +70,8 @@ module RSpec
           diff = differ.diff(actual, expected)
           next if diff.strip.empty?
           if diff == "\e[0m\n\e[0m"
-            "#{DEFAULT_DIFF_LABEL}\n" +
-              "  <diff is empty, do your objects produce identical inspect output?>"
+            "#{diff_label}\n" +
+              "  <The diff is empty, are your objects producing identical `#inspect` output?>"
           else
             "#{diff_label}#{diff}"
           end
--- a/spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
+++ b/spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
@@ -60,14 +60,14 @@ module RSpec
       end
 
       describe "#message_with_diff" do
-        it "returns message and warning if diff is empty" do
+        it "returns a message warning if the diff is empty" do
           allow(FakeDiffer).to receive(:diff) { "\e[0m\n\e[0m" }
           expect(wrapped_value.message_with_diff(
             message, differ, actual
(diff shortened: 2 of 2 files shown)


benoittgt:
I need to work on tests and probably on related tests on other RSpec lib.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -46,7 +46,12 @@ def self.for_many_matchers(matchers)
       # @return [String]
       def message_with_diff(message, differ, actual)
         diff = diffs(differ, actual)
-        message = "#{message}\n#{diff}" unless diff.empty?
+        if @expected_list[0].first.class != actual.class
```
JonRowe: Diff is only printed if the expected is not equal to the actual, hence my earlier suggestion of simply printing a warning message when the diff is empty, we don't need to detect anything else.

benoittgt: That's a good suggestion. I will simplify my PR. Thanks Jon

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
```diff
@@ -66,6 +66,9 @@ def create_matcher(stubs)
             message, differ, actual
           )).to eq(dedent <<-EOS)
             |a message
+            |Diff:
+    <diff is empty, do your objects produce identical inspect output?>
```
JonRowe: The `|` is used to remove whitespace before it (consider it the edge of the terminal) so these can be formatted correctly.

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -46,8 +46,14 @@ def self.for_many_matchers(matchers)
       # @return [String]
       def message_with_diff(message, differ, actual)
         diff = diffs(differ, actual)
-        message = "#{message}\n#{diff}" unless diff.empty?
-        message
+        if diff == "Diff:\e[0m\n\e[0m"
```
JonRowe: Prehaps this should actually be in `def diffs`

benoittgt: yep completely. Sorry I was working on it on the train step by step but didn't push my last change in `diffs`. :)

benoittgt:
For 

```ruby
class Foo
  def to_s; 'foobar'; end
  def inspect; to_s.inspect; end
end

RSpec.describe Foo do
  it do
    expect(Foo.new).to eq('foobar')
  end

  it do
    expect({ u: 'a'}).to eq({ b: 's'})
  end

  it { expect({ :a => "A" }).not_to have_key(:a) }
end
```

Screenshot 2019-09-19 at 19 14 29

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -69,7 +69,12 @@ def diffs(differ, actual)
         @expected_list.map do |(expected, diff_label)|
           diff = differ.diff(actual, expected)
           next if diff.strip.empty?
-          "#{diff_label}#{diff}"
+          if diff == "\e[0m\n\e[0m"
+            "#{DEFAULT_DIFF_LABEL}\n" +
```
JonRowe: Why not `diff_label`?

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
```diff
@@ -69,7 +69,12 @@ def diffs(differ, actual)
...
+              "  <diff is empty, do your objects produce identical inspect output?>"
```
JonRowe: Suggest:
```
<The diff is empty, are your objects producing identical `#inspect` output?>
```

spec/rspec/matchers/expecteds_for_multiple_diffs_spec.rb
```diff
@@ -60,12 +60,14 @@ def create_matcher(stubs)
       end
 
       describe "#message_with_diff" do
-        it "returns just provided message if diff is empty" do
-          allow(FakeDiffer).to receive(:diff) { "" }
+        it "returns message and warning if diff is empty" do
```
JonRowe: Suggest: `it "returns a message warning if the diff is empty" do`

benoittgt: Good idea!

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
benoittgt: Yes, it's better

lib/rspec/matchers/expecteds_for_multiple_diffs.rb
benoittgt: True. This is cleaner. Thanks for the review.

ivoanjo:
Thanks a lot @benoittgt :+1: :+1: :bowing_man: :bowing_man:

PR #1132, opened 2019-09-05
Improve `change` mather detection of objects that have changed.

JonRowe:
If an object is `==` and it's `hash` results are `==` it can still have changed, as it seems instance variable contents are not considered changes. One solution to this problem is to also compare the hashes of the instance attributes. This might have some performance problems so I'm not totally convinced this is the go but it does solve the problem.

### DIFF
--- a/benchmarks/hashing_large_objects.rb
+++ b/benchmarks/hashing_large_objects.rb
@@ -1,0 +1,50 @@
+require 'benchmark/ips'
+
+OBJECT_VARIABLES = Object.instance_method(:instance_variables)
+OBJECT_VARIABLE_GET = Object.instance_method(:instance_variable_get)
+
+def calculate_attribute_hashes(object)
+  OBJECT_VARIABLES.bind(object).call.map { |name| OBJECT_VARIABLE_GET.bind(object).call(name).hash }
+end
+
+[10, 100, 1000, 10000].each do |object_size|
+  klass = Class.new do
+    define_method(:initialize) do
+      object_size.times do |i|
+        instance_variable_set(:"@my_var_#{i}", 'a' * (rand * 100))
+      end
+    end
+  end
+  instance = klass.new
+
+  Benchmark.ips do |ips|
+    ips.report("Get hashes for object of size: #{instance.instance_variables.size}") do
+      calculate_attribute_hashes(instance)
+    end
+  end
+end
+
+# Warming up --------------------------------------
+# Get hashes for object of size: 10
+#                         14.412k i/100ms
+# Calculating -------------------------------------
+# Get hashes for object of size: 10
+#                         158.585k (± 5.2%) i/s -    792.660k in   5.012663s
+# Warming up --------------------------------------
+# Get hashes for object of size: 100
+#                          1.664k i/100ms
+# Calculating -------------------------------------
+# Get hashes for object of size: 100
+#                          17.041k (± 9.0%) i/s -     84.864k in   5.036735s
+# Warming up --------------------------------------
+# Get hashes for object of size: 1000
+#                        173.000  i/100ms
+# Calculating -------------------------------------
+# Get hashes for object of size: 1000
+#                           1.808k (± 4.6%) i/s -      9.169k in   5.082355s
+# Warming up --------------------------------------
+# Get hashes for object of size: 10000
+#                         16.000  i/100ms
+# Calculating -------------------------------------
+# Get hashes for object of size: 10000
+#                         176.298  (± 3.4%) i/s -    896.000  in   5.089070s
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -367,6 +367,8 @@ module RSpec
         def perform_change(event_proc)
           @actual_before = evaluate_value_proc
           @before_hash = @actual_before.hash
+          @before_attribute_hashes = calculate_attribute_hashes(@actual_before)
+
           yield @actual_before if block_given?
 
           return false unless Proc === event_proc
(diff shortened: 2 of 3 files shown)


benoittgt:
> This might have some performance problems so I'm not totally convinced this is the go but it does solve the problem.

Maybe we can write a benchmark?

If there is a cost. Maybe we should make this change available via configuration flag?

pirj:
I'm still trying to grasp `change` matcher implementation, and it seems that in the case when used with qualifiers (e.g. `to` or `from`), it's not really necessary to even compare for hashes or equality.

The following naïve change:
```diff
def matches?(event_proc)
-  perform_change(event_proc) && @change_details.changed? && @matches_before && matches_after?
+  perform_change(event_proc) && @matches_before && matches_after?
end
```

only breaks a few of the examples, that leads me to a thought it might be possible to use `@change_details` only when no qualifiers are specified.

Not sure if it's technically feasible though.

JonRowe:
The change matcher itself has the semantic of the result under test having had to have changed, there are a lot of qualifiers that could pass without the underlying object changing so I don't want to rely on them

JonRowe:
I ran benchmarks on the hash calculation method

```
Calculating -------------------------------------
Get hashes for object of size: 10
                        158.585k (± 5.2%) i/s -    792.660k in   5.012663s
Calculating -------------------------------------
Get hashes for object of size: 100
                         17.041k (± 9.0%) i/s -     84.864k in   5.036735s
Calculating -------------------------------------
Get hashes for object of size: 1000
                          1.808k (± 4.6%) i/s -      9.169k in   5.082355s
Calculating -------------------------------------
Get hashes for object of size: 10000
                        176.298  (± 3.4%) i/s -    896.000  in   5.089070s
```

So even with a really large object with 10'000 instance variables we can gather the hash in under 10 ms so I'm happy merging this if you are

benoittgt:
Thanks a lot for the benchmark! 

LGTM

pirj:
This seems to break some existing specs.
Debugging:

```ruby
[10] pry(#<RSpec::Matchers::BuiltIn::ChangeDetails>)> @actual_before.instance_variables.select { |v| @actual_before.instance_variable_get(v) != @actual_after.instance_variable_get(v) }
=> [:@attributes,
 :@association_cache,
 :@transaction_state,
 :@_new_record_before_last_commit,
 :@errors,
 :@_already_called,
 :@_database_validations_fallback,
 :@new_record_before_save,
 :@mutations_before_last_save,
 :@attributes_changed_by_setter]
[11] pry(#<RSpec::Matchers::BuiltIn::ChangeDetails>)> @actual_before.transaction_state    NoMethodError: undefined method `transaction_state' for #<MasterBookingPageConfiguration:0x00007fad8a5385b0>
Did you mean?  transaction
from /Users/pirj/.rvm/gems/ruby-2.5.5@platform/gems/activemodel-5.2.2.1/lib/active_model/attribute_methods.rb:430:in `method_missing'
[12] pry(#<RSpec::Matchers::BuiltIn::ChangeDetails>)> @actual_before.instance_variable_get :@transaction_state
=> #<ActiveRecord::ConnectionAdapters::TransactionState:0x00007fad88571100
 @children=
  [#<ActiveRecord::ConnectionAdapters::TransactionState:0x00007fad898d88d8
    @children=[],
    @state=:committed>],
 @state=:committed>
[13] pry(#<RSpec::Matchers::BuiltIn::ChangeDetails>)> @actual_after.instance_variable_get :@transaction_state
=> nil
```

Those are service instance variables that are semantically unrelated to the object itself and doesn't really change its state.

I'm not certain if it improves things, since now some of the negated change expectations start failing:
```ruby
expect { some_operation_that_might_or_might_not_change }.not_to change { user.profile }
```

benoittgt:
Do we miss regression tests?

JonRowe:
No, its just more accurate than it was before, I mean the object is changing, its just a question of if this is a semantically change object or an actually changed one

pirj:
@JonRowe For some reason, [this change](https://github.com/rspec/rspec-expectations/blob/91e5b65d5077ea8aeffbc98dbfe03b164ce89f19/lib/rspec/matchers/built_in/change.rb#L396):
```ruby
          @actual_before != @actual_after ||
          @before_hash != @actual_hash ||
          @before_attribute_hashes != @after_attribute_hashes
```
 didn't make it [into 3.9.2](https://github.com/rspec/rspec-expectations/blob/v3.9.2/lib/rspec/matchers/built_in/change.rb#L393):
```ruby
          @actual_before != @actual_after || @before_hash != @actual_hash
```
even though [it's there on `master`](https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/change.rb#L396).

I've tested, it does resolve https://github.com/rspec/rspec-rails/issues/1173.

There's an edge case - when no qualifier is chained, it passes.
Nothing in the object has changed apart `@inspection_filter`'s hash and `@attributes`'s hash.
`@attributes` for a model != `model.attributes`, the former is an `ActiveModel::AttributeSet`, while the latter is a hash.
```ruby
  specify do
    widget = Widget.create(name: 'widget', package: 1)

    expect {
      # widget.update(name: 'A new name', package: 2)
    }.to change { Widget.find(widget.id) }
  end
```

I'm still concerned if this change should be released, as it may break Rails `expect { Model.find(id) }` specs because of this.

JonRowe:
This whole PR is unreleased due to concerns about bugs / performance issues.

JonRowe:
Really I should revert it from master.

JonRowe:
This has been reverted due to performance concerns, if someone wants to tackle this again (even using this as a base) please do!

Issue #1135, opened 2019-10-15
Duplicate matcher does not create a warning


bsvin33t:
### Subject of the issue
When a duplicate matcher is introduced, a warning should be produced by RSpec.

### Your environment
* Ruby version: 2.5.3
* rspec-expectations version: 3.8.4

### Steps to reproduce
```ruby
RSpec::Matchers::DSL::Matcher.new(:duplicate_matcher, {}, self, nil)
RSpec::Matchers::DSL::Matcher.new(:duplicate_matcher, {}, self, nil)
```

### Expected behavior
It should create a warning

### Actual behavior
There is no warning for the user about the duplicate matcher.

bsvin33t:
I have created a PR that does just that. But, I'm afraid that the specs everywhere are breaking because RSpec raises an error when creating a warning. Is there an alternative method that I can use that will allow RSpec to gracefully inform the user about this.

bsvin33t:
I don't think using `config.raise_on_warning = false` would be a good idea as it would suppress all warnings.

JonRowe:
Closing in preference to discussion on #1136 itself.

PR #1136, opened 2019-10-15
Add warning when a duplicate matcher is defined

bsvin33t:
### Subject of the issue
When a duplicate matcher is introduced, a warning should be produced by RSpec.

### Your environment
* Ruby version: 2.5.3
* rspec-expectations version: 3.8.4

### Steps to reproduce
```ruby
RSpec::Matchers::DSL::Matcher.new(:duplicate_matcher, {}, self, nil)
RSpec::Matchers::DSL::Matcher.new(:duplicate_matcher, {}, self, nil)
```

### Expected behavior
It should create a warning

### Actual behavior
There is no warning for the user about the duplicate matcher.

### New Behavior
Introduces a warning when a duplicate matcher is introduced into the testing system

Issue link: https://github.com/rspec/rspec-expectations/issues/1135

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,5 @@
+# This file was generated on 2019-12-05T21:32:23+00:00 from the rspec-dev repo.
+# DO NOT modify it by hand as your changes will get lost the next time it is generated.
+
+github: [JonRowe]
+open_collective: rspec
--- a/.rubocop_rspec_base.yml
+++ b/.rubocop_rspec_base.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2019-07-24T15:33:50+02:00 from the rspec-dev repo.
+# This file was generated on 2019-10-21T09:35:11+01:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # This file contains defaults for RSpec projects. Individual projects
--- a/.rubocop_rspec_base.yml
+++ b/.rubocop_rspec_base.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2019-10-21T09:35:11+01:00 from the rspec-dev repo.
+# This file was generated on 2019-12-26T17:20:33+00:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # This file contains defaults for RSpec projects. Individual projects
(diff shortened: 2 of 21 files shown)


JonRowe:
Migrating conversation from the issue to the pr that actually has the code:

> I have created a PR that does just that. But, I'm afraid that the specs everywhere are breaking because RSpec raises an error when creating a warning. Is there an alternative method that I can use that will allow RSpec to gracefully inform the user about this.

This is probably because you've added a new warning and RSpec itself is triggering it, we don't want our test suite to produce unexpected warnings so you need to fix them if you'd like to continue with this. Your one spec that does trigger a warning shouldn't be the cause as you're using a stub to silence it, although you might like to check how we do this elsewhere as we do also have "expect warning" helpers.

> I don't think using config.raise_on_warning = false would be a good idea as it would suppress all warnings.

As above I'm afraid thats not an option for our test suite.

Overall I'm on the fence about this, I'm not sure it needs a warning but if we add one we've you've chosen to do so isn't the right place, its over in `lib/rspec/matchers/dsl.rb:72` the name of the matcher is not what can cause issues, its the definition of methods in the same place.

How I would implement this is thusly:

```
def define(name, &declarations)
  warn_about_block_args(name, declarations)
  if respond_to?(name)
    RSpec.warning "Matcher definition #{name} is overriding an existing method #{name}".
  end
  define_method name do |*expected, &block_arg|
    RSpec::Matchers::DSL::Matcher.new(name, declarations, self, *expected, &block_arg)
  end
end

```

JonRowe:
Any interest in continuing this @bsvin33t?

bsvin33t:
@JonRowe thanks for the reminder. I would like to work on this. 

But, I'm unable to understand to which object the methods are defined in this line.

https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/dsl.rb#L72

I'm unable to find them on the `self` object. 

I intend to use this information to implement that solution that you have provided.

Can you please point me to something(or code) that I can read/look at so that I can understand what is happening here.

Thanks!
Vineeth

JonRowe:
So example groups define classes, and examples are instances of those classes. So hence checking when you define the matcher method should sense any conflicts that will arise. There is no "one" class this can be mixed into, it needs to be within the method I highlighted.

bsvin33t:
Hey @JonRowe I encountered another roadblock when I added the spec and the implementation for this.

But when I define a new matcher, I am starting to encounter a warning from ruby itself that a method has already been defined(and thus a failure on spec because of an unexpected warning). 

This got me thinking, instead of creating a new warning coming from rspec, why not we just bubble this warning up and show it to us(rspec users)?

JonRowe:
You can undefine methods if you need to.. the current behaviour would bubble up already we don't rescue it...

pirj:
@bsvin33t A friendly reminder.

bsvin33t:
I would like to close this, as I feel the process of undefining methods might add to overall complexity.

Issue #536, opened 2014-04-21
Unexpected target/handler behavior with respect to blocks


laserlemon:
I would expect the following two assertions to behave similarly, at least to the point that they would both pass, but they don't.

``` ruby
expect("foo").to eq("foo")
expect { "foo" }.to eq("foo")
```

In the same vein, I would expect these two assertions to behave differently but they are identical.

``` ruby
expect { "foo" }.to eq("foo")
expect(proc { "foo" }).to eq("foo")
```

It seems that the handoff between the target and the handler could be improved by somehow providing context/insight into whether the incoming "actual" value is something dynamic to be evaluated (a given block) or a static value (even a proc) given as an argument.

I'm hoping it's worth a conversation. Thank you!

mixelpixel:
How does one get the rspec matcher to implement `supports_block_expectations?`?

myronmarston:
> How does one get the rspec matcher to implement `supports_block_expectations?`?

Just define it as a method inside your matcher definition, e.g.

``` ruby
class MyMatcher
  def supports_block_expectations?
    true
  end

  # ...
end
```

If you are using the matcher definition DSL, you can just call [`supports_block_expectations`](http://rspec.info/documentation/3.5/rspec-expectations/RSpec/Matchers/DSL/Macros.html#supports_block_expectations-instance_method):

``` ruby
RSpec::Matchers.define :my_matcher do
  supports_block_expectations

  # ..
end
```

mixelpixel:
hi @myronmarston , thanks for the reply - I'm fairly new to Ruby and rspec. I really appreciate the helpful info!

nicktime:
@myronmarston 
```
You must pass an argument rather than a block to use the provided matcher (eq "foo"),
or the matcher must implement `supports_block_expectations?`.
```
is a really confusing message. It made me think `eq` got a block but needs an argument. Couldn't figure it out until googled up this issue.

myronmarston:
@nicktime what is your code snippet that generated that message?

nicktime:
@myronmarston https://github.com/nicktime/tomograph/blob/add-json-support/spec/lib/tomograph/api_blueprint/yaml_spec.rb#L70
when put in a curly braces like this
```
      expect { subject.to_tomogram.map(&:to_hash) }.to eq(tomogram_hash)
```
So I was thinking the problem is with `tomogram_hash`.

JonRowe:
[...]

Issue #526, opened 2014-04-16
Provide a cleaner delineation between block vs value expectations


myronmarston:
I've occasionally heard confusion from users about when to use `expect {...}` vs `expect(...)`.  Most people I've talked to haven't been confused (the matchers that need a block obviously need a block when you stop and think about it) but I can understand the confusion, particularly for users who are new to ruby (or even programming) and don't understand the reasons why particular matchers require blocks but others don't.  One example of this confusion is in #525.

Currently, if you use the wrong form, you can get confusing errors or false positives/negatives.  For example, passing an arg rather than a block to `yield_control` raises a confusing error:

``` ruby
expect(nil).not_to yield_control # undefined method `arity' for nil:NilClass
```

...but passing an arg to `raise_error` can result in a passing expectation:

``` ruby
expect(nil).not_to raise_error # passes, but prints a warning: `raise_error` was called with non-proc object nil
```

Meanwhile, when you pass a block rather than an arg (intending for the expectation to be set on the return value of the block), you can get false positives:

``` ruby
expect { Foo.bar }.not_to be_nil # passes since a block is not nil, but `Foo.bar` was never called
```

So, I'd like to try to come up with a way to give the user good error messages that warn them they've done the wrong thing in these situations.  At a minimum, we should fix the block matchers so they fail with a consistent message when no block has been passed; then we could tell users "if you're confused, always pass an argument, and the matcher will tell you if you need a block instead".  It would be nice to also give the user an error if they used a block for a value matcher, but that's much more complicated.

For starters, the block form is just syntactic sugar; the following are equivalent:

``` ruby
expect(lambda { raise "boom" }).to raise_error("boom")
expect { raise "boom" }.to raise_error("boom")
```

And in fact, you can use the block form in place of passing a lambda object for matchers that do not normally expect blocks:

``` ruby
expect { }.to be_a(Proc)
```

...works just fine, although it's a bit of a tautology.

I think I'm OK with no longer allowing these to be used interchangeably (as you generally intend one or the other, and if you have a lambda or proc object you want to use with a block matcher, you can always prefix it with `&` to pass as a block), but it would be a bit of a breaking change that would need to go in 3.0.

Once we do that, it would be good to provide an error for when users wrongly pass a block...but I haven't figured out a good way to do that.  Ideas:
[...]

Issue #1134, opened 2019-10-03
Block matchers not detecting passed in block arguments


msxavi:
### Why introducing a breaking change in the latest patch?

Upgrading expectations from 3.8.4 to 3.8.5 breaks specs with: `You must pass a block rather than an argument to `expect` to use the provided block expectation matcher (change `Post.count` by 1).`

### Your environment
* Ruby version: 2.6.3
* rspec-expectations version: 3.8.5

### Steps to reproduce
```
  describe 'POST #create' do
    subject do
      -> { post :create, params: params }
    end

    let(:params) {}

    it 'creates a new post' do
      is_expected.to change { Post.count }.by(1)
    end
  end
```

### Expected behavior
`is_expected.to change { Post.count }.by(1)` should not raise an error.

Breaking changes should be introduced in major updates like 4.0. Is there any approach to keep the compatibility?

### Actual behavior
`is_expected.to change { Post.count }.by(1)` raises
`You must pass a block rather than an argument to `expect` to use the provided block expectation matcher (change `Post.count` by 1)`

highb:
I agree. This sort of major breaking change does not belong in a Z release.

JonRowe:
:wave: Hi, sorry about this, this is not a breaking change though, it's a bug. The change was intended as an improvement to the warning which happens when people incorrectly pass in values to block matchers. This is an edge case that was overlooked and we'll investigate a fix.

pirj:
@msxavi Can you please provide more details on how the subject is defined?
Related:
https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html
https://github.com/rubocop-hq/rspec-style-guide/issues/76

alexjfisher:
@pirj Hi! [Voxpupuli](https://voxpupuli.org/) has noticed build failures in several of our puppet modules this morning.
eg https://github.com/voxpupuli/puppet-cassandra/blob/9954256a6767d51c9dec30b7e7f755bb9734c8d8/spec/defines/schema/permission_spec.rb#L19
https://travis-ci.org/voxpupuli/puppet-cassandra/jobs/593000621#L594

Puppet module testing is based on [rspec-puppet](https://github.com/rodjek/rspec-puppet)

If it helps, I can provide some other examples, or drop by `#voxpupuli` on Freenode IRC.
Thanks

pirj:
[...]

PR #1139, opened 2019-10-29
Print deprecation message when implicit block expectation syntax is used

pirj:
The purpose of this (rather controversial) change is to discourage the usage of a so-called implicit block expectation syntax:

```ruby
subject { -> { do_something } }
it { is_expected.to change(something).to(new_value) }
```

In case this syntax is being used, a deprecation warning will be printed.

This syntax was referenced to as:

> pretty obtuse and not something I’d recommend, generally.
https://github.com/rspec/rspec-expectations/issues/805#issuecomment-112239820

> We should mention they’re not supported.
https://github.com/rspec/rspec-expectations/issues/805#issuecomment-112238533

From [`rspec-core` source](https://github.com/rspec/rspec-core/blob/725b0d2eadaa9ac9f816a6d2d7ef1e507a6705cd/features/subject/one_liner_syntax.feature#L25):

> The one-liner syntax only works with non-block expectations (e.g. `expect(obj).to eq`, etc) and it cannot be used with block expectations (e.g. `expect { object }`).

This is a less strict follow-up to #1125, which was released as part of a patch release 3.8.5, and later reverted in 3.8.6 due to a number of complaints.
Even though [workarounds are possible for the most desperate usages of this syntax](https://github.com/rodjek/rspec-puppet/pull/767), introducing a deprecation warning in pre-4.0 version of RSpec and removal of (coincidental, but still)  support for the implicit block syntax in a major release makes more sense than removal of it in a minor/patch version, since it's used quite often across different codebases I've seen.

### Example Deprecation Message

```
Deprecation Warnings:

You must pass a block rather than an argument to `expect` to use the provided block expectation matcher (change `1`), or the matcher must implement `supports_value_expectations?`.
```

I'm not quite certain it's the best one possible, since it does not refer to the code that led to this deprecation message.
Appreciate any hint how to make it better.

### Additional Things to Notice

According to https://github.com/rspec/rspec-expectations/pull/1125#issuecomment-539429318, the following works with no disruptions:

```ruby
RSpec.describe 'introspect' do
  it do
    expect(-> {}).to be_a(Proc)
  end
end
```

### Related Links

https://github.com/rspec/rspec-expectations/issues/526
https://github.com/rspec/rspec-expectations/pull/530
https://github.com/rspec/rspec-expectations/issues/536
https://github.com/rspec/rspec-expectations/pull/1066
https://github.com/rspec/rspec-expectations/pull/1125

https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html
https://www.reddit.com/r/ruby/comments/cejl3q/call_for_discussion_rspec_implicit_block
https://github.com/rubocop-hq/rspec-style-guide/issues/76
https://github.com/rubocop-hq/rubocop-rspec/pull/789

Example fix for matcher libraries that are forced to keep the interface to provide the syntax for their users https://github.com/rodjek/rspec-puppet/pull/767


Changelog.md
```diff
@@ -32,6 +32,11 @@ Bug Fixes:
 
 Bug Fixes:
 
+* Prevent unsupported implicit block expectation syntax from being used.
+  (Phil Pirozhkov, #1125)
+
+Bug Fixes:
+
```
JonRowe: This needs sorting out, its not right...

lib/rspec/expectations/expectation_target.rb
```diff
@@ -118,9 +157,11 @@ def enforce_block_expectation(matcher)
...
 
       def supports_block_expectations?(matcher)
-        matcher.supports_block_expectations?
-      rescue NoMethodError
-        false
+        if matcher.respond_to?(:supports_block_expectations?)
+          matcher.supports_block_expectations?
+        else
+          false
+        end
```
JonRowe: This can go back as it was

Changelog.md
```diff
@@ -51,7 +56,7 @@ Bug Fixes:
 * Prevent composed `all` matchers from leaking into their siblings leading to duplicate
   failures. (Jamie English, #1086)
 * Prevent objects which change their hash on comparison from failing change checks.
-  (Phil Pirozhkov, #1110)
+  (Phil Pirozhkov, #1100)
```
JonRowe: Is this right?

lib/rspec/expectations/expectation_target.rb
```diff
@@ -33,16 +33,16 @@ def initialize(value)
...
-      def self.for(value, block)
+      def self.for(value, &block)
         if UndefinedValue.equal?(value)
-          unless block
+          unless block_given?
             raise ArgumentError, "You must pass either an argument or a block to `expect`."
           end
           BlockExpectationTarget.new(block)
-        elsif block
+        elsif block_given?
```
JonRowe: Why is the above changing?

lib/rspec/expectations/expectation_target.rb
```diff
@@ -90,6 +90,45 @@ def prevent_operator_matchers(verb)
...
+
+    private
+
+      def enforce_value_expectation(matcher)
+        return if supports_value_expectations?(matcher)
+
+        RSpec.deprecate(
+          "expect(value).to #{RSpec::Support::ObjectFormatter.format(matcher)}",
+          :message =>
+            "You must pass a block rather than an argument to `expect` to use the provided " \
```
JonRowe: Does this message need rewording? It should probably talk about it being recommended rather than must.

lib/rspec/expectations/expectation_target.rb
```diff
@@ -90,6 +90,45 @@ def prevent_operator_matchers(verb)
...
+            "or the matcher must implement `supports_value_expectations?`."
+        )
+      end
+
+      def supports_value_expectations?(matcher)
+        if matcher.respond_to?(:supports_value_expectations?)
+          matcher.supports_value_expectations?
+        else
+          true
+        end
```
JonRowe: As per below, call the method, rescue the `NoMethodError` its better ruby

lib/rspec/matchers/built_in/change.rb
```diff
@@ -242,11 +253,6 @@ def did_change_failure
           "did change from #{@actual_before_description} " \
           "to #{description_of @change_details.actual_after}"
         end
-
-        def not_given_a_block_failure
-          "expected #{@change_details.value_representation} to have changed " \
-          "#{change_description}, but was not given a block"
-        end
```
JonRowe: Why are you removing this? Is this a revert artifact or ?

lib/rspec/matchers/built_in/change.rb
```diff
@@ -337,6 +342,8 @@ def change_description
       class ChangeDetails
         attr_reader :actual_after
 
+        UNDEFINED = Module.new
```
JonRowe: Freeze this

lib/rspec/matchers/built_in/output.rb
```diff
@@ -94,20 +94,25 @@ def supports_block_expectations?
...
+        end
+
       private
 
         def captured?
           @actual.length > 0
         end
 
         def positive_failure_reason
-          return "was not a block" unless Proc === @block
```
JonRowe: Why is this removed?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -199,7 +205,6 @@ def format_backtrace(backtrace)
         end
 
         def given_error
-          return " but was not given a block" unless Proc === @given_proc
```
JonRowe: Why is this removed?

lib/rspec/matchers/built_in/throw_symbol.rb
```diff
@@ -88,20 +88,23 @@ def description
...
+
+        # @api private
         def expects_call_stack_jump?
           true
         end
 
       private
 
         def actual_result
-          return "but was not a block" unless Proc === @block
```
JonRowe: Why is this removed?

spec/spec_helper.rb
```diff
@@ -5,8 +5,8 @@
   minimum_coverage 97
 end
 
-Dir['./spec/support/**/*'].each do |f|
-  require f.sub(%r{\./spec/}, '')
+Dir['./spec/support/**/*.rb'].each do |f|
+  require f.sub(%r|\./spec/|, '')
```
JonRowe: Revert this unless needed please :)

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -405,23 +404,24 @@ def supports_block_expectations?
...
         def positive_failure_reason
-          return 'was not a block' unless @probe.has_block?
-
           'yielded with unexpected arguments' \
           "\nexpected: #{surface_descriptions_in(@expected).inspect}" \
           "\n     got: [#{@actual_formatted.join(", ")}]"
         end
 
         def negative_failure_reason
-          return 'was not a block' unless @probe.has_block?
```
JonRowe: Why is this removed?

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -405,23 +404,24 @@ def supports_block_expectations?
...
+        end
+
       private
 
         def expected_arg_description
           @expected.map { |e| description_of e }.join(', ')
         end
 
         def positive_failure_reason
-          return 'was not a block' unless @probe.has_block?
```
JonRowe: Why is this removed?

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -10,7 +10,6 @@ module BuiltIn
       class YieldProbe
         def self.probe(block, &callback)
           probe = new(block, &callback)
-          return probe unless probe.has_block?
```
JonRowe: I'm not sure I understand any of this files changes... can you elaborate?

JonRowe (dismissed):
Sorry for the delay @pirj got a bit exhausted by this change then forgot about!

pirj:
@JonRowe Is it better finishing this before we release 4.0? Or do you think it will be ok to introduce this warning in 4.0.x and remove in 4.1.0?

JonRowe:
> Or do you think it will be ok to introduce this warning in 4.0.x and remove in 4.1.0?
No that would be a breaking change, any removal would be in a major version.

pirj:
Sounds good 👍 
I intend to wrap this PR up for 3.11 or 3.12, and handle the removal for 4.0.

spec/spec_helper.rb
pirj: I can't:
```
LoadError:
  cannot load such file -- support/shared_examples
```

pirj: Reverted the `%r{}` -> `%r||` change. It triggers RuboCop locally, but we don't check `spec/` on CI.

pirj:
The one and only diff-lcs example failure on Travis seems unrelated.

lib/rspec/expectations/expectation_target.rb
JonRowe: > I don't recall how it was done in 2.99, don't we fail on deprecation warnings in the pre-major version?
No. Or they wouldn't be warnings.

JonRowe: Rather than `"You must"` rephrase to `"<x> is deprecated, and you should do <y> instead"`

spec/support/shared_examples/block_matcher.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    unless options[:disallows_negation]
+      expect {
+        expect(3).not_to matcher
+      }.to fail_with(/was not( given)? a block/)
+    end
+  end
+
+  it 'prints a deprecation warning when given a value' do
+    expect_warn_deprecation(/You must pass a block/)
```
pirj, line 82: Spotted https://github.com/rspec/rspec-expectations/issues/1233

lib/rspec/expectations/expectation_target.rb
pirj: The message is fixed.

pirj:
Hold on, I used 1.9 hash syntax, let me fix it real quick.

JonRowe:
I'll try to review this at the weekend, its too large to do during this week with my work schedule atm.

pirj:
Ping just in case.

pirj:
Ping.

JonRowe:
@pirj this is still on my to do list don't worry, it's just lower than rspec-rails 6.1 support (please don't send ping comments)

lib/rspec/expectations/expectation_target.rb
```diff
@@ -42,7 +42,7 @@ def self.for(value, block)
         elsif block
           raise ArgumentError, "You cannot pass both an argument and a block to `expect`."
         else
-          new(value)
+          ValueExpectationTarget.new(value)
```
JonRowe, line 45: Do we now ever instantiate an `ExpectationTarget`? Could/should this class be merged into it?

JonRowe: Oh wait it's because `BlockExpectationTarget` inherits from it too right?

JonRowe (commented):
I've mostly reviewed this, just need to do some sanity checking on the shared examples when I get a bit more time.

What I am a little concerned about is the proximity to release. We won't get a lot of feedback.

pirj:
We had some feedback in https://github.com/rspec/rspec-expectations/issues/1134
I've ["fixed" it for `rspec-puppet`](https://github.com/rodjek/rspec-puppet/pull/767/files#diff-745b723dc6f8291ba9a7cc938d38bde2R13) including a [monkey-patched `raise_error` matcher](https://github.com/rodjek/rspec-puppet/pull/767/files#diff-225bb60162eb52e6e6555a13d9f882ad5dedbe6fe987a51dabb4c302c882b6deR13).

My idea was to:
 - deprecate in 3.x (if something goes wrong - we can fix in 3-9-maintenance)
 - remove completely (basically re-apply #1125) in 4.0

pirj:
I've just realized that one-liner `should`/`should_not` from Core  `enforce_value_expectation` check as they instantiate `PositiveExpectationHandler`/`NegativeExpectationHandler` directly. So,

```ruby
subject(:action) { -> { raise } }
it { should raise_error StandardError }
```
will **not** print a warning, while when `should` is replaced with `is_expected.to`, it will.

Green light to change those two definitions as below?
```diff
      def should(matcher=nil, message=nil)
-        RSpec::Expectations::PositiveExpectationHandler.handle_matcher(subject, matcher, message)
+        expect(subject).to(matcher, message)
      end

      def should_not(matcher=nil, message=nil)
-        RSpec::Expectations::NegativeExpectationHandler.handle_matcher(subject, matcher, message)
+        expect(subject).not_to(matcher, message)
      end
```
Is there a good reason it wasn't like that in the first place?

pirj:
@JonRowe I'd love to work on the removal of this syntax in `4-0-dev` sometimes maybe Sunday or Tuesday, and the idea was to try to merge `main` in `4-0-dev`. So a gentle reminder.

pirj:
Rebased and resolved a conflict.

pirj:
I intend to incorporate changes to specs made here and add an error for implicit block expectation syntax to `4-0-dev`.
@JonRowe Do you have major concerns regarding this change?

lib/rspec/expectations/expectation_target.rb
```diff
@@ -90,6 +90,44 @@ def prevent_operator_matchers(verb)
...
+      def enforce_value_expectation(matcher)
+        return if supports_value_expectations?(matcher)
+
+        RSpec.deprecate(
+          "expect(value).to #{RSpec::Support::ObjectFormatter.format(matcher)}",
+          :message =>
+            "The implicit block expectation syntax is deprecated, you should pass " \
+            "a block rather than an argument to `expect` to use the provided " \
+            "block expectation matcher (#{RSpec::Support::ObjectFormatter.format(matcher)}), " \
+            "or the matcher must implement `supports_value_expectations?`."
```
JonRowe, line 120:
```suggestion
            "The implicit block expectation syntax is deprecated, you should pass " \
            "a block rather than an argument to `expect` to use the provided " \
            "block expectation matcher or the matcher must implement " \
            "`supports_value_expectations?`. e.g  `expect { value }.to " \
            "#{RSpec::Support::ObjectFormatter.format(matcher)}` not " \
            "`expect(value).to #{RSpec::Support::ObjectFormatter.format(matcher)}`
```

JonRowe:
Ok! I have finally reviewed this. I have the following concerns:

* That this change is going to be unwelcome based on previous feedback.

I think I'm happy to proceed regardless however, as we can always walk it back if we get enough feedback, and most of these changes are good anyway.

* That there is no example of the deprecated code in the deprecation.

e.g. 

> The implicit block expectation syntax is deprecated, you should pass a block rather than an argument to `expect` to use the provided block expectation matcher (raise Exception), or the matcher must implement `supports_value_expectations?`.

I would prefer something like the suggestion I've applied.

pirj:
Sounds good. Applied the suggestion here, and made a similar change in #1285

JonRowe:
Thanks for your patience and hard work on this :)

PR #1143, opened 2019-11-22
Emit a special warning when `nil` is passed to `raise_error`

### LINKED ISSUE #1142: Unintentional passing of `nil` to `raise_error`
### Subject of the issue

We issue a warning when `raise_error` is called with no exception class and no block:
> Instead consider providing a specific error class or message

However, it's hard to imagine why would someone explicitly pass a `nil` as an exception class.
I've stumbled across a situation when this is done unitentionally:

```ruby
let(:status_code_exception_mapping) { {404: SomethingNotFound} }
let(:expected_exception_class) do
  status_code_exception_mapping[status_code]
end
let(:status_code) { 502 }

it "raises a specific error" do
  expect { process! }
    .to raise_error(expected_exception_class)
end
```

In this example, a `raise_error(nil)` was called, even though it wasn't at all obvious since `status_code_exception_mapping` was buried in the code.
What made it even worse is that there was an exception of a different type raised, and the only way to notice that was the warning about possible exception mismatch.

This was hiding a real exception that was way different from anything expected.

### Your environment
* Ruby version: 2.5.1
* rspec-expectations version: 3.6.0 (however, glancing over the latest code it seems that the behaviour has not changed)

### Steps to reproduce

```
# spec/a_spec.rb
RSpec.describe do
  it "raises a specific error" do
    expect { process! }
      .to raise_error(nil)
  end
end
```

### Expected behavior

Error message resulting in spec failure, something like:
> `nil` is not a valid value for an expected exception.

I would expect it to be raised disregarding `on_potential_false_positives` setting.

### Actual behavior

```
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call.
Actual error raised was #<TypeError: no implicit conversion of Hash into String>.
Instead consider providing a specific error class or message.
This message can be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`.
Called from ...
```

### Proposed solution

Use
```ruby
      UndefinedValue = Module.new
```
to make a distinction between `raise_error()` and `raise_error(nil)`.

### Open questions

Is this a breaking change? Should this target 4.0?

pirj:
Usage of `nil` with `raise_error` is with a high chance unintentional, including negation.

Fixes #1142

### DIFF
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -767,7 +767,7 @@ module RSpec
     #   expect { do_something_risky }.to raise_error("that was too risky")
     #
     #   expect { do_something_risky }.not_to raise_error
-    def raise_error(error=nil, message=nil, &block)
+    def raise_error(error=BuiltIn::RaiseError::UndefinedValue, message=nil, &block)
       BuiltIn::RaiseError.new(error, message, &block)
     end
     alias_method :raise_exception, :raise_error
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -316,9 +316,9 @@ module RSpec
     def be_falsey
       BuiltIn::BeFalsey.new
     end
-    alias_matcher :be_falsy,       :be_falsey
+    alias_matcher :be_falsy, :be_falsey
     alias_matcher :a_falsey_value, :be_falsey
-    alias_matcher :a_falsy_value,  :be_falsey
+    alias_matcher :a_falsy_value, :be_falsey
 
     # Passes if actual is nil
     def be_nil
@@ -379,7 +379,7 @@ module RSpec
       BuiltIn::BeAKindOf.new(expected)
     end
     alias_method :be_kind_of, :be_a_kind_of
-    alias_matcher :a_kind_of,  :be_a_kind_of
+    alias_matcher :a_kind_of, :be_a_kind_of
 
     # Passes if actual.between?(min, max). Works with any Comparable object,
     # including String, Symbol, Time, or Numeric (Fixnum, Bignum, Integer,
@@ -406,7 +406,7 @@ module RSpec
       BuiltIn::BeWithin.new(delta)
     end
     alias_matcher :a_value_within, :be_within
-    alias_matcher :within,         :be_within
+    alias_matcher :within, :be_within
 
     # Applied to a proc, specifies that its execution will cause some value to
     # change.
@@ -492,8 +492,8 @@ module RSpec
     def change(receiver=nil, message=nil, &block)
       BuiltIn::Change.new(receiver, message, &block)
     end
-    alias_matcher :a_block_changing,  :change
-    alias_matcher :changing,          :change
+    alias_matcher :a_block_changing, :change
+    alias_matcher :changing, :change
 
     # Passes if actual contains all of the expected regardless of order.
     # This works for collections. Pass in multiple args and it will only
@@ -511,7 +511,7 @@ module RSpec
       BuiltIn::ContainExactly.new(items)
     end
     alias_matcher :a_collection_containing_exactly, :contain_exactly
-    alias_matcher :containing_exactly,              :contain_exactly
+    alias_matcher :containing_exactly, :contain_exactly
 
     # Passes if actual covers expected. This works for
     # Ranges. You can also pass in multiple args
@@ -529,7 +529,7 @@ module RSpec
       BuiltIn::Cover.new(*values)
     end
     alias_matcher :a_range_covering, :cover
-    alias_matcher :covering,         :cover
+    alias_matcher :covering, :cover
 
     # Matches if the actual value ends with the expected value(s). In the case
     # of a string, matches against the last `expected.length` characters of the
@@ -544,8 +544,8 @@ module RSpec
       BuiltIn::EndWith.new(*expected)
     end
     alias_matcher :a_collection_ending_with, :end_with
-    alias_matcher :a_string_ending_with,     :end_with
-    alias_matcher :ending_with,              :end_with
+    alias_matcher :a_string_ending_with, :end_with
+    alias_matcher :ending_with, :end_with
 
     # Passes if <tt>actual == expected</tt>.
     #
@@ -559,7 +559,7 @@ module RSpec
       BuiltIn::Eq.new(expected)
     end
     alias_matcher :an_object_eq_to, :eq
-    alias_matcher :eq_to,           :eq
+    alias_matcher :eq_to, :eq
 
     # Passes if `actual.eql?(expected)`
     #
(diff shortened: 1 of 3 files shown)


lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -9,13 +9,20 @@ module BuiltIn
...
+        # argument. We can't use `nil` for that because we need to warn when `nil` is
+        # passed in a different way. It's an Object, not a Module, since Module's `===`
+        # does not evaluate to true when compared to itself.
+        UndefinedValue = Object.new
+
+        def initialize(expected_error_or_message, expected_message, &block)
           @block = block
           @actual_error = nil
-          @warn_about_bare_error = expected_error_or_message.nil?
+          @warn_about_bare_error = UndefinedValue.equal?(expected_error_or_message)
```
JonRowe: A `===` is more our style here...

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -9,13 +9,20 @@ module BuiltIn
       class RaiseError
         include Composable
 
-        def initialize(expected_error_or_message=nil, expected_message=nil, &block)
+        # Used as a sentinel value to be able to tell when the user did not pass an
+        # argument. We can't use `nil` for that because we need to warn when `nil` is
+        # passed in a different way. It's an Object, not a Module, since Module's `===`
+        # does not evaluate to true when compared to itself.
```
JonRowe: 👍Great explanation. Perhaps freeze the object too?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -166,11 +177,37 @@ def warn_about_bare_error
...
+                         "any other error would cause the expectation to pass, including " \
+                         "those raised by Ruby (e.g. `NoMethodError`, `NameError` and " \
+                         "`ArgumentError`), meaning the code you are intending to test may " \
+                         "not even get reached. " \
+                         "Instead consider using `expect { }.not_to raise_error` or " \
+                         "`expect { }.to raise_error(DifferentSpecificErrorClass)`. " \
+                         "This message can be suppressed by setting: " \
+                         "`RSpec::Expectations.configuration.on_potential_false" \
+                         "_positives = :nothing`")
+        end
```
JonRowe: Would this not be the same as the original negative error warning?

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -166,11 +177,37 @@ def warn_about_bare_error
...
+                         "unintentional, it risks false positives, since `raise_error` " \
+                         "will match when Ruby raises a `NoMethodError`, `NameError` or " \
+                         "`ArgumentError`, potentially allowing the expectation to pass " \
+                         "without even executing the method you are intending to call. " \
+                         "#{warning}"\
+                         "Instead consider providing a specific error class or message. " \
+                         "This message can be suppressed by setting: " \
+                         "`RSpec::Expectations.configuration.on_potential_false" \
+                         "_positives = :nothing`")
+        end
```
JonRowe: Risks or will cause? :joy:

lib/rspec/matchers.rb
```diff
@@ -761,20 +761,22 @@ def output(expected=nil)
...
     #   expect { do_something_risky }.to raise_error(PoorRiskDecisionError, /oo ri/)
+    #   expect { do_something_risky }.to raise_error("that was too risky")
     #
     #   expect { do_something_risky }.not_to raise_error
-    def raise_error(error=nil, message=nil, &block)
+    def raise_error(error=BuiltIn::RaiseError::UndefinedValue, message=nil, &block)
       BuiltIn::RaiseError.new(error, message, &block)
     end
-    alias_method :raise_exception,  :raise_error
+    alias_method :raise_exception,   :raise_error
```
JonRowe: Honestly we should just change all these to `, :` rather than trying to align them.

lib/rspec/matchers/built_in/raise_error.rb
JonRowe: I mean could you not do this with `warn_about_negative_false_positive!("raise_error(nil)")`

pirj: I've added this distinctive message intentionally. First off, I don't think anyone would deliberately write `.not_to raise_error(nil)`, so there should be some kind of mistake, and a regular message telling that didn't pass anything to `raise_error` and they'd better use `raise_error(DifferentSpecificErrorClass)` might be confusing and would suggest an idea of an RSpec bug, since they obviously passed something. Another point was to make it more visible to remind us that we planned to switch it to raise an error eventually.
Do you think it's reasonable enough to keep a separate message?

JonRowe: > I've added this distinctive message intentionally.

Except it's not distinct? It's identical except for swapping `#{expression}` for `raise_error(nil)` so why not set expression to `"raise_error(nil)"` and generate the same string using the existing method?

pirj: Sorry for the delay in getting your idea. Fixed, it turns out to be more concise the way you suggested 👍

JonRowe:
Do you want to a change log for this @pirj? Or shall I?

pirj:
Added in [704f0a26](https://github.com/rspec/rspec-expectations/commit/704f0a26).

pirj:
I've received a notification about a note, but the author removed the note. Presumably from @mathieujobin or @mjobin-mdsol.

> I wonder if via a config option, we could make this raise an InvalidError instead of a simple warning...

Adding such an option makes total sense to me.

Would you like to send a PR to `main` to make this configurable?

If you want to go a step further and raise an exception (I'd go with `ArgumentError`), you can also:
 - send another PR to `main`, but targeting the release [3.99](https://github.com/rspec/rspec-expectations/milestone/10) that would warn users that the option will be removed, and the exception would become the default
 - send a PR to [`4-0-dev`](https://github.com/rspec/rspec-expectations/tree/4-0-dev) with removal of the option and making `ArgumentError` the default behaviour

@JonRowe Do you support the idea?

mjobin-mdsol:
Thanks for the ping @pirj 

I can't commit to open a PR with a new config options, I'm sorry. But I created a branch on my end for testing existing projects

https://github.com/mjobin-mdsol/rspec-expectations/commit/140845384f39c5b36a5b63e7ea2f95e5dd85bcef

JonRowe:
@pirj I'm happy for this idea to be a thing if someone wants to make it so.

PR #1151, opened 2020-01-01
Update travis build scripts 2020 01 01 for master

benoittgt:
See: https://github.com/rspec/rspec-dev/pull/233

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
 end
 
 gem 'coderay' # for syntax highlighting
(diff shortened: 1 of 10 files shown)


Gemfile
```diff
@@ -13,7 +13,7 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
 end
 
 gem 'coderay' # for syntax highlighting
-gem 'yard', '~> 0.9.12', :require => false
+gem 'yard', '~> 0.9.22', :require => false
```
JonRowe: Can we limit this to 2.7.0 only until we resolve the docs differences? Unless you want to go through the diff now?

PR #1154, opened 2020-01-11
Fix ruby 2.7 keyword args deprecation

### LINKED ISSUE #1150: Warnings for custom matchers with keyword args on Ruby 2.7
### Subject of the issue
Custom matchers with keyword arguments cause the following warning:

> ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/rspec-expectations-3.9.0/lib/rspec/matchers/dsl.rb:467: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

### Your environment
* Ruby version: 2.7.0
* rspec-expectations version: 3.9.0

### Steps to reproduce
```ruby
require 'rspec'

RSpec::Matchers.define(:be_foo) do |bar: nil|
  match { expect(true).to be true }
end

RSpec.describe 'Foo' do
  it 'works' do
    expect(:foo).to be_foo(bar: :qux)
  end
end

RSpec::ExampleGroups::Foo.run
```

### Expected behavior
No warning is emitted.

### Actual behavior
A warning is emitted.

deivid-rodriguez:
In particular,

```
<path/to/rspec-expectations>/lib/rspec/matchers/dsl.rb:526: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

Fixes #1150.

### DIFF
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -522,7 +522,7 @@ module RSpec
         # Rails' test helper methods, but it's also a useful
         # feature in its own right.
         if RUBY_VERSION.to_f >= 2.7
-          require_relative "dsl27"
+          require "rspec/matchers/dsl27"
           include DSL27
         else
           def method_missing(method, *args, &block)
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -521,11 +521,21 @@ module RSpec
         # rspec-rails so that it can define matchers that wrap
         # Rails' test helper methods, but it's also a useful
         # feature in its own right.
-        def method_missing(method, *args, &block)
-          if @matcher_execution_context.respond_to?(method)
-            @matcher_execution_context.__send__ method, *args, &block
-          else
-            super(method, *args, &block)
+        if RUBY_VERSION.to_f >= 2.7
+          def method_missing(method, *args, **kwargs, &block)
+            if @matcher_execution_context.respond_to?(method)
+              @matcher_execution_context.__send__ method, *args, **kwargs, &block
+            else
+              super(method, *args, *kwargs, &block)
+            end
+          end
+        else
+          def method_missing(method, *args, &block)
+            if @matcher_execution_context.respond_to?(method)
+              @matcher_execution_context.__send__ method, *args, &block
+            else
+              super(method, *args, &block)
+            end
           end
         end
       end
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -522,13 +522,8 @@ module RSpec
         # Rails' test helper methods, but it's also a useful
         # feature in its own right.
         if RUBY_VERSION.to_f >= 2.7
-          def method_missing(method, *args, **kwargs, &block)
-            if @matcher_execution_context.respond_to?(method)
-              @matcher_execution_context.__send__ method, *args, **kwargs, &block
-            else
-              super(method, *args, *kwargs, &block)
-            end
-          end
+          require_relative "dsl27"
+          include DSL27
         else
           def method_missing(method, *args, &block)
             if @matcher_execution_context.respond_to?(method)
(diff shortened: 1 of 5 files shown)


lib/rspec/matchers/dsl.rb
```diff
@@ -521,11 +521,21 @@ def actual_arg_for(block)
...
         # feature in its own right.
-        def method_missing(method, *args, &block)
-          if @matcher_execution_context.respond_to?(method)
-            @matcher_execution_context.__send__ method, *args, &block
-          else
-            super(method, *args, &block)
+        if RUBY_VERSION.to_f >= 2.7
+          def method_missing(method, *args, **kwargs, &block)
+            if @matcher_execution_context.respond_to?(method)
+              @matcher_execution_context.__send__ method, *args, **kwargs, &block
```
pirj: 1.8 won't be able to parse it.
You might be interested in checking this out https://github.com/rspec/rspec-support/pull/392#discussion_r361765961

deivid-rodriguez: A wrapped the whole thing with a version check so that I don't change anything in older rubies, is that not ok?

deivid-rodriguez: Oh, I get it now. Ruby 1.8 can't parse the `**` syntax. I'll read the linked issue :+1:.

lib/rspec/matchers/dsl27.rb
```diff
@@ -0,0 +1,16 @@
+module RSpec
+  module Matchers
+    # Helper module to include code not parsable on very old rubies
+    module DSL27
```
JonRowe: Honestly I'd namespace this under DSL and call it
```suggestion
    module DSL
      module KeywordArgumentSupport
```

lib/rspec/matchers/dsl.rb
```diff
@@ -521,11 +521,16 @@ def actual_arg_for(block)
         # rspec-rails so that it can define matchers that wrap
         # Rails' test helper methods, but it's also a useful
         # feature in its own right.
-        def method_missing(method, *args, &block)
-          if @matcher_execution_context.respond_to?(method)
-            @matcher_execution_context.__send__ method, *args, &block
-          else
-            super(method, *args, &block)
+        if RUBY_VERSION.to_f >= 2.7
```
JonRowe: Any keyword argument supporting ruby supports this, so we can use the `RubyFeatures.kw_args_supported?` feature detection instead.
```suggestion
        if RubyFeatures.kw_args_supported?
```

spec/rspec/expectations_spec.rb
```diff
@@ -12,6 +12,18 @@
       /stringio/, # Used by `output` matcher. Can't be easily avoided.
       /rbconfig/  # required by rspec-support
     ]
+  }
+
+  if RUBY_VERSION.to_f >= 2.7
```
JonRowe:
```suggestion
  if RubyFeatures.kw_args_supported?
```

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -7,6 +7,11 @@ def ok
     "ok"
   end
 
+  if RUBY_VERSION.to_f >= 2.7
+    require "rspec/matchers/dsl27_syntax"
+    include RSpec::Matchers::DSL27Syntax
```
JonRowe: If this is just loading the test code I think it needs to be named differently so its clear.

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -29,6 +34,13 @@ def ok
     expect { matcher_b.i_dont_exist }.to raise_error(NameError)
   end
 
+  if RUBY_VERSION.to_f >= 2.7
+    it "doesn't warn when method available in the scope of the example uses keyword args" do
+      RSpec::Matchers.define(:matcher_kw) {}
+      expect(matcher_kw.kw(a: 1)).to eq(1)
```
JonRowe: I'd expect this to blow up, because its still eval'd.

deivid-rodriguez:
I don't like this either, I'll just close and let you handle it!

PR #1155, opened 2020-01-11
Improve composability of `include` matcher

pirj:
To match
```ruby
hash = {OpenStruct.new(value: :special) => true}
```
a workaround had to be used
```ruby
expect(hash.to_a)
  .to contain_exactly([have_attributes(value: :special), true])
```

This change makes it simpler:
```ruby
expect(hash)
  .to include(have_attributes(value: :special) => true)
```

Found here: https://github.com/rubocop-hq/rubocop-rspec/pull/851#discussion_r362733064

### DIFF
--- a/features/built_in_matchers/include.feature
+++ b/features/built_in_matchers/include.feature
@@ -23,6 +23,8 @@ Feature: `include` matcher
     expect(:a => 1, :b => 2).to include(:a, :b)
     expect(:a => 1, :b => 2).to include(:a => 1)
     expect(:a => 1, :b => 2).to include(:b => 2, :a => 1)
+    expect(:a => 1, :b => 2).to include(match(/b/) => 2)
+    expect(:a => 1, :b => 2).to include(match(/b/) => be_even)
     expect(:a => 1, :b => 2).not_to include(:c)
     expect(:a => 1, :b => 2).not_to include(:a => 2)
     expect(:a => 1, :b => 2).not_to include(:c => 3)
@@ -141,4 +143,3 @@ Feature: `include` matcher
       | expected {:a => 7, :b => 5} not to include :a                 |
       | expected {:a => 7, :b => 5} to include {:d => 3}              |
       | expected {:a => 7, :b => 5} not to include {:a => 7}          |
-
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -107,7 +107,10 @@ module RSpec
         end
 
         def actual_hash_includes?(expected_key, expected_value)
-          actual_value = actual.fetch(expected_key) { return false }
+          actual_value =
+            actual.fetch(expected_key) do
+              actual.find(Proc.new { return false }) { |actual_key, _| values_match?(expected_key, actual_key) }[1]
+            end
           values_match?(expected_value, actual_value)
         end
 
--- a/spec/rspec/matchers/built_in/include_spec.rb
+++ b/spec/rspec/matchers/built_in/include_spec.rb
@@ -749,18 +749,59 @@ RSpec.describe "#include matcher" do
 
     describe "expect(hash).to include(key_matcher)" do
       it "passes when the matcher matches a key", :if => (RUBY_VERSION.to_f > 1.8) do
-        expect(:drink => "water", :food => "bread").to include(a_string_matching(/foo/))
+        expect(:drink => "water", :food => "bread").to include(match(/foo/))
       end
 
       it 'provides a description' do
-        description = include(a_string_matching(/foo/)).description
-        expect(description).to eq("include (a string matching /foo/)")
+        description = include(match(/foo/)).description
+        expect(description).to eq("include (match /foo/)")
       end
 
       it 'fails with a clear message when the matcher does not match', :if => (RUBY_VERSION.to_f > 1.8) do
         expect {
-          expect(:drink => "water", :food => "bread").to include(a_string_matching(/bar/))
-        }.to fail_matching('expected {:drink => "water", :food => "bread"} to include (a string matching /bar/)')
+          expect(:drink => "water", :food => "bread").to include(match(/bar/))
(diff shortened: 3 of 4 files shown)


spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -764,6 +764,39 @@ def matches?(_)
       end
     end
 
+    describe "expect(hash).to include(key_matcher => value)" do
+      it "passes when the matcher matches a pair", :if => (RUBY_VERSION.to_f > 1.8) do
```
pirj: I have little to no idea why `:if` is needed. Just copy and pasted it.

pirj:
AppVeyor Ruby 2.0 build stucks at `bundle --version`. Seems unrelated. Restarted the build, and the problem persists.

pirj:
Please hold off merging this. I think I've found a weird side effect on `have_broadcasted_to` in case of composable matchers and `.with_indifferent_hash`.

spec/rspec/matchers/built_in/include_spec.rb
JonRowe: I mean... does it fail without this?

lib/rspec/matchers/built_in/include.rb
```diff
@@ -107,8 +107,10 @@ def comparing_hash_to_a_subset?(expected_item)
         end
 
         def actual_hash_includes?(expected_key, expected_value)
-          actual_value = actual.fetch(expected_key) { return false }
-          values_match?(expected_value, actual_value)
+          actual.any? do |actual_key, actual_value|
+            values_match?(expected_key, actual_key) &&
+              values_match?(expected_value, actual_value)
+          end
```
JonRowe: So... this changes the matcher to iterate over the hash, which could have an impact on performance in large hashes... Something like this:

```Ruby
actual_value =
  actual.fetch(expected_key) do
    actual.find(proc { return false }) { |actual_key, _| values_match?(expected_key, actual_key) }[0]
  end
values_match?(expected_value, actual_value)
```

Would allow shortcuts for set keys, but still search for fuzzy keys... (The extra proc keeps the `return false` safeguard and prevents `if`s)... Its not as pretty... but shouldn't cause regressions in performance.... I'm on the fence as to wether thats better...

pirj: Nice, I love that much more than my solution.

Do you want me to benchmark on huge large hashes? I believe that with normal keys which are significantly more common it will sure be faster.

It also doesn't break `have_broadcasted_to` :D

Small correction `[0]` => `[1]`.

pirj:
It seems that tests were limited to Ruby > 1.8 since in 1.8.7:
```ruby
/foo/ === :food # => false
```
while in later versions it is `true`.

lib/rspec/matchers/built_in/include.rb
JonRowe: Er... Why `1`? Oh because you want the value not the key?

pirj: Yep 👍

spec/rspec/matchers/built_in/include_spec.rb
pirj: It seems that tests were limited to Ruby > 1.8 since in 1.8.7:
```ruby
/foo/ === :food # => false
```
while in later versions it is `true`.
Specs using symbolic keys and regexp natural matchers all fail.

JonRowe: Well what is the downside of trying `to_s` on the keys, even if just for 1.8.7?

pirj: Won't `expect({a: 1}).to include('a' => 1)` start passing for other versions for no reason?

JonRowe: Hmm, I guess the problem is actually in `a_string_matching`, I mean half of me is wondering if it is even valid for that to pass for symbols...

pirj: `a_string_matching` is a semantic alias to `match` which is:
>    # Given a `Regexp` or `String`, passes if `actual.match(pattern)`

So what probably has to be fixed is the spec that uses `a_string_matching` to match against symbols.

JonRowe: Actually ok, I'm happy with that, change the alias out for match, and then its ok because its clear it is Ruby's match semantics at work, not us. (I'm just worried about giving the wrong impression).

pirj: Done. Also fixed some neighbour examples with the same mismatch.

features/built_in_matchers/include.feature
```diff
@@ -23,6 +23,8 @@ Feature: `include` matcher
     expect(:a => 1, :b => 2).to include(:a, :b)
     expect(:a => 1, :b => 2).to include(:a => 1)
     expect(:a => 1, :b => 2).to include(:b => 2, :a => 1)
+    expect(:a => 1, :b => 2).to include(a_string_matching(/b/) => 2)
+    expect(:a => 1, :b => 2).to include(a_string_matching(/b/) => be_even)
```
JonRowe: These need to be changed too :)

pirj: Nice catch :+1:
Done.

lib/rspec/matchers/built_in/include.rb
```diff
@@ -107,7 +107,10 @@ def comparing_hash_to_a_subset?(expected_item)
         end
 
         def actual_hash_includes?(expected_key, expected_value)
-          actual_value = actual.fetch(expected_key) { return false }
+          actual_value =
+            actual.fetch(expected_key) do
+              actual.find(Proc.new { return false }) { |actual_key, _| values_match?(expected_key, actual_key) }[1]
```
benoittgt: That's good!

PR #1156, opened 2020-01-17
Be able to set custom failure message with mocks

benoittgt:
- See: https://github.com/rspec/rspec-mocks/issues/1250

### To do:

- [x] Finalize https://github.com/rspec/rspec-mocks/pull/1312

### DIFF
--- a/lib/rspec/expectations/handler.rb
+++ b/lib/rspec/expectations/handler.rb
@@ -44,10 +44,16 @@ module RSpec
 
     # @private
     class PositiveExpectationHandler
-      def self.handle_matcher(actual, initial_matcher, message=nil, &block)
-        ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
+      def self.handle_matcher(actual, initial_matcher, custom_message=nil, &block)
+        ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::PositiveOperatorMatcher.new(actual) unless initial_matcher
-          matcher.matches?(actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message)
+
+          is_a_match = matcher.matches?(actual, &block)
+          if custom_message && is_a_match.respond_to?(:error_generator)
+            is_a_match.error_generator.opts[:message] = custom_message
+          end
+
+          is_a_match || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message)
         end
       end
 
@@ -66,10 +72,16 @@ module RSpec
 
     # @private
     class NegativeExpectationHandler
-      def self.handle_matcher(actual, initial_matcher, message=nil, &block)
-        ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
+      def self.handle_matcher(actual, initial_matcher, custom_message=nil, &block)
+        ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::NegativeOperatorMatcher.new(actual) unless initial_matcher
-          does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated)
+
+          is_not_a_match = does_not_match?(matcher, actual, &block)
+          if custom_message && is_not_a_match.respond_to?(:error_generator)
+            is_not_a_match.error_generator.opts[:message] = custom_message
+          end
+
+          is_not_a_match || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message_when_negated)
         end
       end
 
--- a/lib/rspec/expectations/handler.rb
+++ b/lib/rspec/expectations/handler.rb
@@ -48,12 +48,12 @@ module RSpec
         ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::PositiveOperatorMatcher.new(actual) unless initial_matcher
 
-          is_a_match = matcher.matches?(actual, &block)
-          if custom_message && is_a_match.respond_to?(:error_generator)
-            is_a_match.error_generator.opts[:message] = custom_message
+          match_result = matcher.matches?(actual, &block)
+          if custom_message && match_result.respond_to?(:error_generator)
+            match_result.error_generator.opts[:message] = custom_message
           end
 
-          is_a_match || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message)
+          match_result || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message)
         end
       end
 
@@ -76,12 +76,12 @@ module RSpec
         ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::NegativeOperatorMatcher.new(actual) unless initial_matcher
 
-          is_not_a_match = does_not_match?(matcher, actual, &block)
-          if custom_message && is_not_a_match.respond_to?(:error_generator)
-            is_not_a_match.error_generator.opts[:message] = custom_message
+          negated_match_result = does_not_match?(matcher, actual, &block)
+          if custom_message && negated_match_result.respond_to?(:error_generator)
+            negated_match_result.error_generator.opts[:message] = custom_message
           end
 
-          is_not_a_match || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message_when_negated)
+          negated_match_result || ExpectationHelper.handle_failure(matcher, custom_message, :failure_message_when_negated)
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/rspec/expectations/handler.rb
```diff
@@ -44,10 +44,16 @@ def self.handle_failure(matcher, message, failure_message_method)
...
     # @private
     class PositiveExpectationHandler
-      def self.handle_matcher(actual, initial_matcher, message=nil, &block)
-        ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
+      def self.handle_matcher(actual, initial_matcher, custom_message=nil, &block)
+        ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::PositiveOperatorMatcher.new(actual) unless initial_matcher
-          matcher.matches?(actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message)
+
+          is_a_match = matcher.matches?(actual, &block)
```
pirj: Maybe `is_a_match` => `matches`?

JonRowe: I'd go for `match_result`

lib/rspec/expectations/handler.rb
```diff
@@ -66,10 +72,16 @@ def self.opposite_should_method
...
     # @private
     class NegativeExpectationHandler
-      def self.handle_matcher(actual, initial_matcher, message=nil, &block)
-        ExpectationHelper.with_matcher(self, initial_matcher, message) do |matcher|
+      def self.handle_matcher(actual, initial_matcher, custom_message=nil, &block)
+        ExpectationHelper.with_matcher(self, initial_matcher, custom_message) do |matcher|
           return ::RSpec::Matchers::BuiltIn::NegativeOperatorMatcher.new(actual) unless initial_matcher
-          does_not_match?(matcher, actual, &block) || ExpectationHelper.handle_failure(matcher, message, :failure_message_when_negated)
+
+          is_not_a_match = does_not_match?(matcher, actual, &block)
```
JonRowe: If you change above then I'd call this `negated_match_result`

spec/rspec/expectations/handler_spec.rb
```diff
@@ -130,6 +130,18 @@ module Expectations
 
           RSpec::Expectations::PositiveExpectationHandler.handle_matcher(actual, matcher, failure_message)
         end
+
+        it "add custom failure message to mock matcher if there is one" do
+          error_generator = double(:opts => { :message => {} })
+          verifying_message_expectation = double(:error_generator => error_generator)
```
JonRowe: Whats the real thing for this, can we either use a real one or use an `instance_double`?

JonRowe: Ah it's mock specific isn't it, should be able to use an instance double of `RSpec::Mocks::ErrorGenerator` etc?

spec/rspec/expectations/handler_spec.rb
```diff
@@ -209,6 +221,22 @@ module Expectations
...
+          matcher = double(
+            "matcher",
+            :failure_message_when_negated => "custom",
+            :does_not_match? => verifying_message_expectation
+          )
```
JonRowe: As above I'd like to see verifying doubles at least here

lib/rspec/expectations/handler.rb
benoittgt: Yep good idea! Much better like this.

lib/rspec/expectations/handler.rb
benoittgt: Yep good idea! Much better like this.

spec/rspec/expectations/handler_spec.rb
benoittgt: I was not sure about this when we wrote the test. Do we want to see some `rspec-mocks` code into our specs? We have only one occurrence at the moment in `rspec-expectation`.

https://github.com/rspec/rspec-expectations/blob/ed736c3fc996ca0c978e869806bf41359db395a3/spec/rspec/expectations/configuration_spec.rb#L32

JonRowe: This is integration with `rspec-mocks` so I think it's important to double the right thing. This way if the implementation changes on the other side it will fail. You can add a note as well explaining if that makes you feel more comfortable?

spec/rspec/expectations/handler_spec.rb
```diff
@@ -130,6 +130,21 @@ module Expectations
...
+          error_generator = instance_double("RSpec::Mocks::ErrorGenerator", :opts => { :message => {} })
+          verifying_message_expectation = instance_double(
+            "RSpec::Mocks::VerifyingMessageExpectation",
+            :error_generator => error_generator
+          )
+          matcher = double("matcher", :failure_message => "message", :matches? => verifying_message_expectation)
+
+          actual = Object.new
+
+          handle_matcher_answer = RSpec::Expectations::PositiveExpectationHandler.handle_matcher(actual, matcher, "custom")
```
JonRowe: I'd keep this name in line with above:
```suggestion
          handle_matcher_result = RSpec::Expectations::PositiveExpectationHandler.handle_matcher(actual, matcher, "custom")
```

spec/rspec/expectations/handler_spec.rb
```diff
@@ -130,6 +130,21 @@ module Expectations
...
+
+          expect(handle_matcher_answer.error_generator.opts).to eq({ :message => "custom" })
```
JonRowe:
```suggestion
          expect(handle_matcher_result.error_generator.opts).to eq({ :message => "custom" })
```

spec/rspec/expectations/handler_spec.rb
```diff
@@ -209,6 +224,25 @@ module Expectations
...
+          )
+          matcher = double(
+            "matcher",
+            :failure_message_when_negated => "custom",
+            :does_not_match? => verifying_message_expectation
+          )
+
+          actual = Object.new
+
+          handle_matcher_answer = RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher, "custom")
```
JonRowe:
```suggestion
          handle_matcher_result = RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher, "custom")
```

spec/rspec/expectations/handler_spec.rb
```diff
@@ -209,6 +224,25 @@ module Expectations
...
+
+          expect(handle_matcher_answer.error_generator.opts).to eq({ :message => "custom" })
```
JonRowe:
```suggestion
          expect(handle_matcher_result.error_generator.opts).to eq({ :message => "custom" })
```

spec/rspec/expectations/handler_spec.rb
benoittgt: I prefer with `instance_double`. I did the change in https://github.com/rspec/rspec-expectations/pull/1156/commits/7a776d6aa8f71dff8186dda33da2227dc4a15a39. @JonRowe if you can just confirm it's ok to have `RSpec::Mocks` references in `rspec-expectation` then we are good. 😊

benoittgt:
I added a _WIP_ commit to use the proper branch with the last change in specs. https://github.com/rspec/rspec-expectations/pull/1156/commits/0ac96f862cb5ec98265dd08c5c3cd1b288a8e376

I don't know if it is the way RSpec team does it. 🤷‍♂

JonRowe:
Yeah we normally cross pin the branches, check they go green, remove the commit from one, merge ignoring failure, remove the commit from the other, check still green, merge, rebuild other master to check still green

JonRowe:
This is a pain point that having a monorepo would fix.

benoittgt:
CI is ok:
- https://travis-ci.org/rspec/rspec-expectations/builds/641676081?utm_source=github_status&utm_medium=notification
- https://ci.appveyor.com/project/rspec/rspec-expectations/builds/30349505

I added the changelog here because I too often fail with proper grammar. :)

spec/rspec/expectations/handler_spec.rb
```diff
@@ -209,6 +224,25 @@ module Expectations
...
+            "matcher",
+            :failure_message_when_negated => "custom",
+            :does_not_match? => verifying_message_expectation
+          )
+
+          actual = Object.new
+
+          handle_matcher_result = RSpec::Expectations::NegativeExpectationHandler.handle_matcher(actual, matcher, "custom")
+
+          expect(handle_matcher_result.error_generator.opts).to eq({ :message => "custom" })
```
pirj: Does it have to be explicitly wrapped in curly braces?

JonRowe:
@benoittgt I'm happy for you to merge this and rspec/rspec-mocks#1312 but they need to be done, "together" is it were so I'm leaving it to you to manage if thats ok?

spec/rspec/expectations/handler_spec.rb
benoittgt: I did this for old Rubies if I remember.

benoittgt:
Yes @JonRowe. I will merge this one first. Then follow the flow :)

PR #1157, opened 2020-01-19
[DO NOT MERGE] Attempt to fail rspec-rails sub-build

pirj:
There was a problem with the initial implementation of https://github.com/rspec/rspec-expectations/pull/1155 that make `have_broadcasted_to` spec to fail locally (for a reason), but on CI `rspec-rails` was green.

This is an attempt to detect if `rspec-rails` sub-build fails if we deliberately fail [this example](https://github.com/rspec/rspec-rails/blob/b1ffe94265b33b02616658ec3c1921fc0355f662/spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb#L164).

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -108,10 +108,9 @@ module RSpec
 
         def actual_hash_includes?(expected_key, expected_value)
           fail 'Oh just not David!' if expected_value == 'David'
-          actual_value =
-            actual.fetch(expected_key) do
-              actual.find(Proc.new { return false }) { |actual_key, _| values_match?(expected_key, actual_key) }[1]
-            end
+          actual_value = actual.fetch(expected_key) do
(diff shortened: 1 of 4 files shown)


pirj:
Houston, we have a problem.

pirj:
Houston, we have a problem.

pirj:
Houston, we have a problem.

benoittgt:
Maybe @palkan have an idea 😊

palkan:
Action Cable specs are skipped 'cause the Rails version is <6: https://travis-ci.org/rspec/rspec-expectations/jobs/639225826

```
...
Using rails 5.0.7.2
```

pirj:
Nice catch!
Should we [default](https://github.com/rspec/rspec-rails/blob/master/Gemfile-rails-dependencies#L31) to the latest instead `rails` in `rspec-rails`?

JonRowe:
You can if you want, it's historically not been because of issues we've had running it locally and it makes no difference to ci

PR #1164, opened 2020-02-18
Handle broken method handles in `respond_to` and `have_attributes` matchers.

### LINKED ISSUE #1162: have_attributes/respond_to matchers don't work with ActiveRecord
### Subject of the issue

I am trying to use the `have_attributes` matcher on a Rails ActiveRecord instance. It fails with the undefined method error. The issue is with the `respond_to` matcher that `have_attributes` is using under the hood. Specifically, I've tracked it down to this line: https://github.com/rspec/rspec-support/blob/9cec8e858228ee1d1395e4e6186ac39c2a83d705/lib/rspec/support.rb#L51
The problem is that ActiveRecord attributes are not regular methods, so `.method(:attribute_name)` fails. However, both `.send(:attribute_name)` and `.respond_to?(:attribute_name)` work just fine, which is why I consider this a bug, since the docs state that:
```ruby
expect(obj).to respond_to(:foo) # pass if obj.respond_to?(:foo)
```

This seems similar to what was happening in #1107.

### Your environment
* Ruby version: 2.5.5p157
* Rails version: 5.0.7.2
* rspec-expectations version: 3.8.4

### Steps to reproduce
```ruby
expect(active_record_instance).to respond_to :attribute_name
```

### Expected behavior
Expectation passes if the ActiveRecord model has an attribute called `attribute_name` (i.e. `active_record_instance.responds_to?(:attribute_name)` returns `true`)

### Actual behavior
`binding.pry` before [the culrpit line](https://github.com/rspec/rspec-support/blob/9cec8e858228ee1d1395e4e6186ac39c2a83d705/lib/rspec/support.rb#L51):
```
[14] pry(RSpec::Support)> KERNEL_METHOD_METHOD.bind(object).call(method_name)
NameError: undefined method `total' for class `#<Class:0x00000001133d0fe0>'
from (pry):20:in `method'
[15] pry(RSpec::Support)> object.__send__(:total)
=> 0
[16] pry(RSpec::Support)> KERNEL_METHOD_METHOD.bind(object).call(:total)
NameError: undefined method `total' for class `#<Class:0x00000001133d0fe0>'
from (pry):22:in `method'
[17] pry(RSpec::Support)> object.respond_to? :total
=> true
```

JonRowe:
Some objects using dynamic attributes (Rails) don't always define methods, but we require them to when checking the number of arguments a method `responds_to`, convert the error into something explicit that explains this. 

As a follow up this then means that the `have_attributes` matcher will never work with such objects as it always checks that an attribute method has `0` arguments. So we add a private api to allow the `have_attributes` matcher to check arguments when they are defined normally, but suppress the warning generated and consider that a pass.

Fixes #1162

### DIFF
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -1,5 +1,7 @@
 RSpec::Support.require_rspec_support "method_signature_verifier"
 
+# TODO: Refactor this file to be under our class length
+# rubocop:disable ClassLength
 module RSpec
   module Matchers
     module BuiltIn
@@ -194,3 +196,4 @@ module RSpec
     end
   end
 end
+# rubocop:enable ClassLength
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -11,6 +11,7 @@ module RSpec
           @names = names
           @expected_arity = nil
           @expected_keywords = []
+          @ignoring_method_signature_failure = false
           @unlimited_arguments = nil
           @arbitrary_keywords = nil
         end
@@ -100,6 +101,12 @@ module RSpec
           "respond to #{pp_names}#{with_arity}"
         end
 
+        # @api private
+        # Used by other matchers to suppress a check
+        def ignoring_method_signature_failure!
+          @ignoring_method_signature_failure = true
+        end
+
       private
 
         def find_failing_method_names(actual, filter_method)
@@ -135,6 +142,7 @@ module RSpec
             Support::StrictSignatureVerifier.new(method_signature_for(actual, name)).
               with_expectation(expectation).valid?
           rescue NameError
+            return true if @ignoring_method_signature_failure
             raise ArgumentError, "The #{matcher_name} matcher requires that " \
                                  "the actual object define the method(s) in " \
                                  "order to check arity, but the method " \
--- a/lib/rspec/matchers/built_in/respond_to.rb
+++ b/lib/rspec/matchers/built_in/respond_to.rb
@@ -109,7 +109,7 @@ module RSpec
           end
         end
 
-        def matches_arity?(actual, name)
+        def setup_method_signature_expectation
           expectation = Support::MethodSignatureExpectation.new
 
           if @expected_arity.is_a?(Range)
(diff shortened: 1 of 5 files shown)


lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -100,6 +103,12 @@ def description
           "respond to #{pp_names}#{with_arity}"
         end
 
+        # @api private
+        # Used by other matchers to supress a check
```
pirj: Sup**p**ress.

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -130,6 +132,12 @@ def matches_arity?(actual, name)
           expectation.expect_unlimited_arguments = @unlimited_arguments
           expectation.expect_arbitrary_keywords  = @arbitrary_keywords
```
pirj: Unrelated. Just wondering how strict is it to only consider numeric `@expected_arity` as a `min`.
Can it be:
```ruby
expectation.min_count = @expected_arity
expectation.max_count = @expected_arity
```

lib/rspec/matchers/built_in/respond_to.rb
```diff
@@ -123,10 +132,25 @@ def matches_arity?(actual, name)
...
           return true if expectation.empty?
 
-          Support::StrictSignatureVerifier.new(method_signature_for(actual, name)).
-            with_expectation(expectation).valid?
+          begin
+            Support::StrictSignatureVerifier.new(method_signature_for(actual, name)).
+              with_expectation(expectation).valid?
+          rescue NameError
+            return true if @ignoring_method_signature_failure
+            raise ::ArgumentError, "The #{matcher_name} matcher requires that " \
```
pirj: Does it have to be specified as a global const?

lib/rspec/matchers/built_in/respond_to.rb
JonRowe: Its set to min later on in rspec-support if its never set, its a bit more complex than doing this.

lib/rspec/matchers/built_in/respond_to.rb
JonRowe: You mean the `::` probably not

JonRowe:
Tidied up those bits

PR #1167, opened 2020-03-15
yield_control hardening

marcandre:
This fixes two misuses of `yield_control`.

The following resulted in strange failures
```
expect{ .. }.to yield_control.at_least(:trice)
```

Worse, the following doesn't test what is thought to be tested:

```
expect{ .. }.to yield_control.at_least(2).at_most(4).times
```

This PR fixes both.

### DIFF
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -98,7 +98,7 @@ module RSpec
       # Not intended to be instantiated directly.
       class YieldControl < BaseMatcher
         def initialize
-          at_least(:once)
+          @expectation_type = @expected_yields_count = nil
         end
 
         # @api public
@@ -153,7 +153,7 @@ module RSpec
         def matches?(block)
           @probe = YieldProbe.probe(block)
           return false unless @probe.has_block?
-
+          return @probe.num_yields > 0 unless @expectation_type
           @probe.num_yields.__send__(@expectation_type, @expected_yields_count)
         end
 
@@ -182,6 +182,8 @@ module RSpec
       private
 
         def set_expected_yields_count(relativity, n)
+          raise "Multiple count constraints are not supported" if @expectation_type
+
           @expectation_type = relativity
           @expected_yields_count = count_constraint_to_number(n)
         end
@@ -200,23 +202,24 @@ module RSpec
 
         def failure_reason
           return ' but was not a block' unless @probe.has_block?
-          " #{human_readable_expectation_type}#{human_readable_count(@expected_yields_count)}" \
-          " but yielded #{human_readable_count(@probe.num_yields)}"
+          "#{human_readable_expectation_type}#{human_readable_count(@expected_yields_count)}" \
+          " but yielded#{human_readable_count(@probe.num_yields)}"
         end
 
         def human_readable_expectation_type
           case @expectation_type
-          when :<= then 'at most '
-          when :>= then 'at least '
+          when :<= then ' at most'
+          when :>= then ' at least'
           else ''
           end
         end
 
         def human_readable_count(count)
           case count
-          when 1 then 'once'
-          when 2 then 'twice'
-          else "#{count} times"
+          when nil then ''
+          when 1 then ' once'
+          when 2 then ' twice'
+          else " #{count} times"
           end
         end
       end
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -183,17 +183,23 @@ module RSpec
 
         def set_expected_yields_count(relativity, n)
           @expectation_type = relativity
-          @expected_yields_count = case n
-                                   when Numeric then n
-                                   when :once then 1
-                                   when :twice then 2
-                                   when :thrice then 3
-                                   end
+          @expected_yields_count = count_constraint_to_number(n)
+        end
+
+        def count_constraint_to_number(n)
+          case n
+          when Numeric then n
+          when :once then 1
+          when :twice then 2
+          when :thrice then 3
+          else
+            raise ArgumentError, "Expected a number, :once, :twice or :thrice," \
+              " but got #{n}"
+          end
         end
 
         def failure_reason
           return ' but was not a block' unless @probe.has_block?
-          return '' unless @expected_yields_count
           " #{human_readable_expectation_type}#{human_readable_count(@expected_yields_count)}" \
           " but yielded #{human_readable_count(@probe.num_yields)}"
         end
--- a/spec/rspec/matchers/built_in/yield_spec.rb
+++ b/spec/rspec/matchers/built_in/yield_spec.rb
@@ -58,7 +58,7 @@ RSpec.describe "yield_control matcher" do
     it 'fails if the block does not yield' do
       expect {
         expect { |b| _dont_yield(&b) }.to yield_control
-      }.to fail_with(/expected given block to yield control/)
+      }.to fail_with(/expected given block to yield control but/)
     end
 
     it 'does not return a meaningful value from the block' do
@@ -71,6 +71,16 @@ RSpec.describe "yield_control matcher" do
       expect { yield_control.exactly('2') }.to raise_error(ArgumentError)
       expect { yield_control.at_least(:trice_with_typo) }.to raise_error(ArgumentError)
       expect { yield_control.at_most(nil) }.to raise_error(ArgumentError)
+    end
+
+    it 'is yet to support multiple calls to compatible count constraints' do
+      pending
+      expect { |b| 1.upto(4, &b) }.to yield_control.at_least(3).at_most(4).times
+      expect { |b| 1.upto(2, &b) }.not_to yield_control.at_least(3).at_most(4).times
+    end
+
+    it 'raises an error on multiple incompatible calls to count constraints' do
+      expect { yield_control.once.twice }.to raise_error(/multiple/i)
     end
 
     context "with exact count" do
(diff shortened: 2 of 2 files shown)


spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -67,6 +67,16 @@ def each_arg(*args, &block)
...
     end
 
+    it 'raises an error if given an invalid count argument' do
+      expect { yield_control.exactly('2') }.to raise_error(ArgumentError)
+      expect { yield_control.at_least(:trice_with_typo) }.to raise_error(ArgumentError)
+      expect { yield_control.at_most(nil) }.to raise_error(ArgumentError)
+    end
+
+    it 'forbids multiple calls to count constraints' do
+      expect { yield_control.at_least(1).at_most(4).times }.to raise_error(/multiple/i)
```
JonRowe: I would expect this to work, are you saying it doesn't or do you want to change this to not work?

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -98,7 +98,9 @@ def assert_valid_expect_block!
       # Not intended to be instantiated directly.
       class YieldControl < BaseMatcher
         def initialize
+          @already_set = false
           at_least(:once)
+          @already_set = false
```
JonRowe: I'd prefer refactoring so that we can se the count internally and just check if the expectation is set.

spec/rspec/matchers/built_in/yield_spec.rb
marcandre: Currently, the call to `at_most(4)` silently overwrites the constraint set by `at_least(1)`. This (amended) PR simply raises a `NotImplementedError`. If it is deemed necessary to support ranges, it could be implemented.

lib/rspec/matchers/built_in/yield.rb
marcandre: I agree, that was a mistake. I actually changed the implementation for #1168 but thought this PR had been merged already. Fixed.

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -154,7 +154,7 @@ def matches?(block)
           @probe = YieldProbe.probe(block)
           return false unless @probe.has_block?
 
-          @probe.num_yields.__send__(@expectation_type, @expected_yields_count)
+          @probe.num_yields.__send__(@expectation_type || :>=, @expected_yields_count || 1)
```
JonRowe: Ok I'd prefer a guard that checks if these are both nil, sends the default, rather than this or, even if in theory its impossible one could be nil and the other not.

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -182,35 +182,44 @@ def supports_block_expectations?
       private
 
         def set_expected_yields_count(relativity, n)
+          raise NotImplementedError, "Multiple count constraints are not supported" if @expectation_type
```
JonRowe:
```suggestion
          raise ArgumentError, "Multiple constraints are not currently supported" if @expectation_type
```

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -182,35 +182,44 @@ def supports_block_expectations?
...
+          @expected_yields_count = count_constraint_to_number(n)
+        end
+
+        def count_constraint_to_number(n)
+          case n
+          when Numeric then n
+          when :once then 1
+          when :twice then 2
+          when :thrice then 3
+          else raise ArgumentError,
```
JonRowe: Raise on a different line please.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -67,6 +67,16 @@ def each_arg(*args, &block)
       expect(val).to be_nil
     end
 
+    it 'raises an error if given an invalid count argument' do
+      expect { yield_control.exactly('2') }.to raise_error(ArgumentError)
+      expect { yield_control.at_least(:trice_with_typo) }.to raise_error(ArgumentError)
+      expect { yield_control.at_most(nil) }.to raise_error(ArgumentError)
+    end
+
+    it 'forbids multiple calls to count constraints' do
```
JonRowe:
```suggestion
    # TODO: This needs implementing.
    it 'is yet to support multiple calls to count constraints' do
```

lib/rspec/matchers/built_in/yield.rb
marcandre: It's definitely *not* and `ArgumentError`, i.e. there are no possible arguments that would not generate this error and there might not be arguments at all. Some cases will also never be implemented and always be an error (e.g. `include(...).twice.once`). Maybe just a `RuntimeError`?

JonRowe: Its an `ArgumentError` because it is attempting to overwrite existing arguments.

lib/rspec/matchers/built_in/yield.rb
marcandre: Ok. Note: not possible for one to be `nil` and not the other (after this PR)

lib/rspec/matchers/built_in/yield.rb
JonRowe: I don't feel that strongly about it though, you can change to a `StandardError` for now, and I hope to find some time to fix it properly, in which case `.once.twice` etc will raise an `ArgumentError` (because thats an invalid combination) whilst `at_least(:once).at_most(:thrice)` would work.

lib/rspec/matchers/built_in/yield.rb
JonRowe: Its possible, its just not supported.

lib/rspec/matchers/built_in/yield.rb
marcandre: Let's avoid an argument and agree to disagree on what is an argument 😅

I believe it is always a mistake to raise a (naked) `StandardError`. I'm not aware of Ruby ever raising one for example. That class is meant for rescuing only.

I've changed it to a `RuntimeError`.

lib/rspec/matchers/built_in/yield.rb
marcandre: > Its possible, its just not supported.

I'm surprised by how confident you seem to be about this. How you would get one set to `nil` and the other at something else (after this PR)?

JonRowe: Extensions of this class and/or modifying the internals.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -67,6 +67,17 @@ def each_arg(*args, &block)
       expect(val).to be_nil
     end
 
+    it 'raises an error if given an invalid count argument' do
+      expect { yield_control.exactly('2') }.to raise_error(ArgumentError)
+      expect { yield_control.at_least(:trice_with_typo) }.to raise_error(ArgumentError)
+      expect { yield_control.at_most(nil) }.to raise_error(ArgumentError)
+    end
+
+    # TODO: This needs implementing.
```
pirj: Maybe make it a `pending` spec?
```ruby
it 'is yet to support multiple calls to count constraints' do
  pending
  expect { |b| 1.upto(4, &b) }.to yield_control.at_least(:once).at_most(3) }.to raise_error ...
  expect { |b| 1.upto(3, &b) }.to yield_control.at_least(:once).at_most(3) }.not_to raise_error
end
```

marcandre: Ok. I added a test for incompatible constraints (e.g. `include(...).once.twice`) and a pending test for `at_least.at_most`.

pirj: 👍

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -67,6 +67,23 @@ def each_arg(*args, &block)
...
+      expect { yield_control.at_most(nil) }.to raise_error(ArgumentError)
+    end
+
+    it 'is yet to support multiple calls to compatible count constraints' do
+      pending
+      expect { |b| 1.upto(4, &b) }.to yield_control.at_least(3).at_most(4).times
+      expect { |b| 1.upto(2, &b) }.not_to yield_control.at_least(3).at_most(4).times
+    end
+
+    # TODO: This needs implementing.
```
pirj: Leftover TODO. Don't think `.once.twice` is possible to implement in a reasonable way :D

marcandre: Oops, fixed.

spec/rspec/matchers/built_in/yield_spec.rb
```diff
@@ -67,6 +67,22 @@ def each_arg(*args, &block)
       expect(val).to be_nil
     end
 
+    it 'raises an error if given an invalid count argument' do
```
pirj: Just for clarity, do you mind to add examples for a failure message like
```ruby
expect {
  expect { |b| 0.times.each(&b) }.to yield_control.at_least(:once)
}.to fail_with(/expected given block to yield control at least once but did not yield/)

expect {
  expect { |b| 2.times.each(&b) }.to yield_control.at_most(:once)
}.to fail_with(/expected given block to yield control at most once but yielded twice/)

expect {
  expect { |b| 1.times.each(&b) }.to yield_control.at_least(:twice)
}.to fail_with(/expected given block to yield control at least twice but yielded once/)

expect {
  expect { |b| 0.times.each(&b) }.to yield_control
}.to fail_with(/expected given block to yield control but did not yield/)
```

JonRowe:
Thanks!

PR #1169, opened 2020-03-18
Add support for multiple yield control combinations

JonRowe:
Follow up to #1167

### DIFF
--- a/lib/rspec/matchers/built_in/yield.rb
+++ b/lib/rspec/matchers/built_in/yield.rb
@@ -149,12 +149,22 @@ module RSpec
           self
         end
 
+        if RUBY_VERSION.to_f > 1.8
+          def cover?(count, number)
+            count.cover?(number)
+          end
+        else
+          def cover?(count, number)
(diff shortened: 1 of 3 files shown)


pirj:
Thanks!

JonRowe:
Sorry was messing with the history, its ready for review now

lib/rspec/matchers/built_in/yield.rb
```diff
@@ -182,10 +194,37 @@ def supports_block_expectations?
...
 
         def set_expected_yields_count(relativity, n)
-          raise "Multiple count constraints are not supported" if @expectation_type
+          raise_unsupported_yield_expectation if unsupported_yield_expectation?(relativity)
+
+          constraint_count = count_constraint_to_number(n)
 
-          @expectation_type = relativity
-          @expected_yields_count = count_constraint_to_number(n)
+          if @expectation_type == :<= && relativity == :>=
```
pirj: What do you think about adding a check so that `at_least(3).at_most(2)` can't be called?

PR #1176, opened 2020-04-05
Fix Ruby 2.7 keyword args deprecation in DSL

### LINKED ISSUE #1150: Warnings for custom matchers with keyword args on Ruby 2.7

JonRowe:
Fixes #1150, replaces #1154

### DIFF
--- a/spec/rspec/matchers/dsl_spec.rb
+++ b/spec/rspec/matchers/dsl_spec.rb
@@ -30,6 +30,7 @@ RSpec.describe "a matcher defined using the matcher DSL" do
   end
 
   if RSpec::Support::RubyFeatures.required_kw_args_supported?
+    binding.eval(<<-CODE, __FILE__, __LINE__)
     it 'supports the use of required keyword arguments in definition block' do
       RSpec::Matchers.define(:match_required_kw) do |bar:|
         match { expect(actual).to eq bar }
@@ -45,9 +46,11 @@ RSpec.describe "a matcher defined using the matcher DSL" do
       RSpec::Matchers.define(:matcher_required_kw_on_method) {}
       expect(matcher_required_kw_on_method.kw(a: 1)).to eq(1)
     end
+    CODE
   end
 
   if RSpec::Support::RubyFeatures.kw_args_supported?
+    binding.eval(<<-CODE, __FILE__, __LINE__)
     it 'supports the use of optional keyword arguments in definition block' do
       RSpec::Matchers.define(:match_optional_kw) do |bar: nil|
         match { expect(actual).to eq bar }
@@ -63,6 +66,7 @@ RSpec.describe "a matcher defined using the matcher DSL" do
       RSpec::Matchers.define(:matcher_optional_kw_on_method) {}
       expect(matcher_optional_kw_on_method.optional_kw(a: 1)).to eq(1)
     end
+    CODE
   end
 
   it "clears user instance variables between invocations" do
--- a/spec/rspec/matchers/dsl_spec.rb
+++ b/spec/rspec/matchers/dsl_spec.rb
@@ -29,6 +29,42 @@ RSpec.describe "a matcher defined using the matcher DSL" do
     expect { matcher_b.i_dont_exist }.to raise_error(NameError)
   end
 
+  if RSpec::Support::RubyFeatures.required_kw_args_supported?
+    it 'supports the use of required keyword arguments in definition block' do
+      RSpec::Matchers.define(:match_required_kw) do |bar:|
+        match { expect(actual).to eq bar }
+      end
+      expect(1).to match_required_kw(bar: 1)
+    end
+
+    def kw(a:)
+      a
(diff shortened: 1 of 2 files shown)


pirj:
This change doesn't seem to fix the original issue:
```ruby
    RSpec::Matchers.define(:be_foo) do |bar: nil|
      match { expect(true).to be true }
    end

    it 'works' do
      expect(:foo).to be_foo(bar: :qux)
    end
```

JonRowe:
Oh hmph, then the original implementation can't have fixed it either.

JonRowe:
Added additional specs and fix.

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -29,6 +29,46 @@ def ok
...
+      expect(1).to match_optional_kw(bar: 1)
+    end
+
+    def optional_kw(a: nil)
+      a
+    end
+
+    it "supports the use of optional keyword arguments on methods" do
+      RSpec::Matchers.define(:matcher_optional_kw_on_method) {}
+      expect(matcher_optional_kw_on_method.optional_kw(a: 1)).to eq(1)
```
benoittgt, line 67: I do not understand the link between `def optional_kw(a: nil)` and `matcher_optional_kw_on_method`. How `RSpec::Matchers.define(:matcher_optional_kw_on_method) {}` can use it?

JonRowe: We pass `self` through to the matchers when they are created and they use `method_missing` to invoke the methods on example.

benoittgt: ookk. I missed the `self`. Thanks

JonRowe:
Happy with this @pirj?

lostie:
Hi there 👋 , when are you planning to release a new version of the gem that includes these changes?

JonRowe:
As a volunteer project we don't have a timescale sorry. Soon, theres some other keyword warning removals in progress.

pirj:
@lostie It would be very helpful if you could run RSpec's `master` against your project and open a ticket if you happen to experience more of those warnings.

marcandre:
@JonRowe a release would be appreciated, even if it doesn't resolve all issues, it resolves some of them. IMO, the time to create a release is so small compared to just about anything involving maintenance of a gem... when you have a great tests suite like `rspec` does, frequent bug fix releases make a lot of sense to me.

marcandre:
FWIW, this resolves the only warnings I had in 2.7 🎉 

I had to modify my `Gemfile` with:
```
gem 'rspec', git: 'https://github.com/rspec/rspec.git'
gem 'rspec-expectations', git: 'https://github.com/rspec/rspec-expectations.git'
gem 'rspec-core', git: 'https://github.com/rspec/rspec-core.git'
gem 'rspec-mocks', git: 'https://github.com/rspec/rspec-mocks.git'
gem 'rspec-support', git: 'https://github.com/rspec/rspec-support.git'
```

I'm betting 0.25$ that figuring this out took me more than it takes to make a release.

JonRowe:
@marcandre yes I plan to do one soon, the pattern is generally to try to batch releases together but not wait too long, after all too frequent releases are exhausting for both user and maintainer. 

However I haven't had any "not tired" time since this was re highlighted and I don't do releases when I'm tired because I tend to make mistakes that are then very hard to reverse in a project as widely used as RSpec.

FWIW theres a bunch that goes into releasing one of the RSpec gems, checking any dependencies on the other gems, going through master and checking its been properly merged into `3-9-maintenance`,  then finally bumping the release, pushing it up, waiting for the build, then cutting the release is the final step.

JonRowe:
This has been released as 3.9.2, I apologise for the delay but for your, 25¢ it took a good 45 minutes to check and sort the status of the two branches, get everything built and released, so I really hope it took you less time to figure out how to use the master branches :joy:

marcandre:
Thanks @JonRowe  I owe you 25¢ 😆
I'll make that a beer if we get the chance to cross paths!

lostie:
Thank you @JonRowe !

PR #1182, opened 2020-05-10
swap ordering for values with units

kwstannard:
The way `@delta * @expected.abs` is ordered currently, the expected
object *must* coerce into a Numeric to work with `#percent_of`.

Example of where this falls down:

```
expect(Weight.new("5 lb"))
  .to be_within(20).percent_of(Weight.new("4 lb"))
```

This is a statement that makes logical sense, but impossible to do with
`@delta * @expected.abs`.

By swapping the order, the expected merely has to respond to `#abs` and
`#*` and also be comparable.

I was debating making a test, but I couldn't think of a class in the stdlib that would trigger this problem and didn't want to write a whole large class just to swap the order of two variables.

### DIFF
--- a/lib/rspec/matchers/built_in/be_within.rb
+++ b/lib/rspec/matchers/built_in/be_within.rb
@@ -23,7 +23,7 @@ module RSpec
         # a percent comparison.
         def percent_of(expected)
           @expected  = expected
-          @tolerance = @delta * @expected.abs / 100.0
+          @tolerance = @expected.abs * @delta / 100.0
           @unit      = '%'
           self
         end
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks for this, an interesting find! See review for a suggested class to add a test (which we would want!) :)

kwstannard:
Thanks for the minimal implementation. Will try that out tomorrow.

kwstannard:
I had to add a few more methods, but this test exposed the issue and passes with the change.

kwstannard:
Looks like there is a preexisting bug in the output.

JonRowe:
No it's just 1.8.7.

kwstannard:
I am referring to how one place uses `actual_formatted` but another uses `@expected` when it should probably use `expected_formatted`

JonRowe:
This has gone green, but the status is lagging

kwstannard:
🎉

PR #1184, opened 2020-05-12
Remove kw args warning from has_<n> matcher

### LINKED ISSUE #1183: Ruby 2.7 deprecations in `has` matchers
### Subject of the issue
Expectations like:
```ruby
expect(foo).to have_bar(kw: arg)
```
cause deprecation warnings in Ruby 2.7. Presumably, this is because `RSpec::Matchers::BuiltIn::Has` accepts and passes on `*args` but not `**args`. 

### Your environment
* Ruby version: 2.7.1

### Steps to reproduce
As mentioned above, this happens with an expectation like:
```ruby
expect(foo).to have_bar(kw: arg)
```

### Expected behavior
There should be no deprecation warning.

### Actual behavior
There is a deprecation warning.

JonRowe:
Fixes #1183 Requires rspec/rspec-support#413

### DIFF
--- a/lib/rspec/matchers/built_in/has.rb
+++ b/lib/rspec/matchers/built_in/has.rb
@@ -65,7 +65,7 @@ module RSpec
         end
 
         def predicate_matches?
-          @actual.__send__(predicate, *@args, &@block)
+          RSpec::Support::WithKeywordsWhenNeeded.send_to(@actual, predicate, *@args, &@block)
         end
 
         def predicate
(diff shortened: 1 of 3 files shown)


JonRowe:
Build failures is because of nested spec runs for rspec-core not having the right dependency, if the support PR is merged and pin dropped it should be fixed.

lib/rspec/matchers/built_in/has.rb
```diff
@@ -65,7 +65,7 @@ def predicate_exists?
         end
 
         def predicate_matches?
-          @actual.__send__(predicate, *@args, &@block)
+          RSpec::Support::WithKeywordsWhenNeeded.send_to(@actual, predicate, *@args, &@block)
```
benoittgt, line 68: That's great to have this. LGTM

PR #1187, opened 2020-05-26
Ensure that has_<n> matcher works silently with keyword arguments

JonRowe:
Alternative to #1184 that doesn't use rspec-support meta magic.

### DIFF
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -181,10 +181,21 @@ module RSpec
       class BePredicate < BaseMatcher
         include BeHelpers
 
-        def initialize(*args, &block)
-          @expected = parse_expected(args.shift)
-          @args = args
-          @block = block
+        if RSpec::Support::RubyFeatures.kw_args_supported?
+          binding.eval(<<-CODE, __FILE__, __LINE__)
+          def initialize(*args, **kwargs, &block)
+            @expected = parse_expected(args.shift)
+            @args = args
+            @kwargs = kwargs
+            @block = block
+          end
+          CODE
+        else
+          def initialize(*args, &block)
+            @expected = parse_expected(args.shift)
+            @args = args
+            @block = block
+          end
         end
 
         def matches?(actual, &block)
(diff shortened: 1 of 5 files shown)


JonRowe:
@pirj @benoittgt how do we feel about this? Realistically the only repetition we could avoid with this approach is the final send with / without keywords, it means a lot more string evals but should be as performant as the original which I know my support branch is not. 

Whats interesting is that `*args, **kwargs` is not the equivalent of `*args`, as an empty hash is sent if you are not expecting keyword args... hence the if check surrounding empty kwargs.

lib/rspec/matchers/built_in/be.rb
```diff
@@ -236,9 +247,22 @@ def private_predicate?
...
-        def predicate_matches?
-          method_name = actual.respond_to?(predicate) ? predicate : present_tense_predicate
-          @predicate_matches = actual.__send__(method_name, *@args, &@block)
+        if RSpec::Support::RubyFeatures.kw_args_supported?
+          binding.eval(<<-CODE, __FILE__, __LINE__)
+          def predicate_matches?
+            method_name = actual.respond_to?(predicate) ? predicate : present_tense_predicate
+            if @kwargs.empty?
+              @predicate_matches = actual.__send__(method_name, *@args, &@block)
+            else
```
pirj, line 256: What fails if we remove this?
```ruby
k={}
def a(x)
end
def b(y, z: 1)
end
a(1, **k)
b(2, **k)
```
neither of those calls results in a warning 😕

pirj: I see the failures, but I'm having hard times understanding it, especially that the ArgumentError isn't raised where it should.

JonRowe:
@pirj you need the master branch of `rspec-core` to avoid that warning

lib/rspec/matchers/built_in/be.rb
JonRowe: It doesn't cause a warning, it sends an extra parameter to some places not expecting it.

JonRowe: e.g. splatting keyword arguments, creates keyword arguments, and nil is not `{}`

pirj: > you need the master branch of rspec-core to avoid that warning

👍 that recent change slipped my mind.

Got it. So I've played around with it, and it seems that `be_predicate()` is sending `__send__(:predicate, *[], **{})` and `**{}` goes for a regular parameter as `{}` 🤦 

```ruby
def m(p)
  puts p
end

__send__(:m, *[], **{})
```
```
warning: Passing the keyword argument as the last hash parameter is deprecated
{}
```

I don't have enough words to express my grief.

PR #1193, opened 2020-06-11
Improve Hash formatting in `EnglishPhrasing.list`

### LINKED ISSUE #1192: Default failure message for custom matcher incorrectly formats `expected` hashes
### Subject of the issue

When using a hash as the `expected` value in a custom matcher, the default failure message formats the `expected` value as a list of 2-element arrays.

(An `actual` value which is a hash is formatted as expected in the default error message.)

### Your environment
* Ruby version: 2.6.5

### Steps to reproduce
```ruby
begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.9.0" # Activate the gem and version you are reporting the issue against.
end

require 'rspec/autorun'

RSpec::Matchers.define :custom_match do |expected|
  match do |actual|
    actual == expected
  end
end

# These are all intended to fail in order to demonstrate the failure messages
# for different `expected` values.
RSpec.describe 'default failure message' do
  it 'correctly formats a string' do
    expect(nil).to custom_match('foo')
  end

  it 'correctly formats an Object' do
    expect(nil).to custom_match(Object.new)
  end

  it 'correctly formats an array' do
    expect(nil).to custom_match([1, 'foo', Object.new])
  end

  it 'incorrectly formats a hash as an array' do
    expect(nil).to custom_match({a: 1, foo: 'bar', bang: Object.new})
  end
end
```

### Expected behavior
When a hash is the `expected` object, the failure message should format the hash as a hash.

### Actual behavior
When a hash is the `expected` object, the failure message formats the hash as a list of 2-element key-value arrays:

```
  1) default failure message incorrectly formats a hash as an array
     Failure/Error: expect(nil).to custom_match({a: 1, foo: 'bar', bang: Object.new})
       expected nil to custom match [:a, 1], [:foo, "bar"], and [:bang, #<Object:0x00007fecf73d6dc8>]
     # rspec_report.rb:39:in `block (2 levels) in <main>'
```

---

I believe this is a result of the use of `EnglishPhrasing.list` in `BaseMatcher#description`:

https://github.com/rspec/rspec-expectations/blob/eb1787f0e1a5ec2c2650048ee60a45d4c9738d78/lib/rspec/matchers/built_in/base_matcher.rb#L60-L64

`EnglishPhrasing.list` in turn appears to format all enumerables, including hashes, as a list:

https://github.com/rspec/rspec-expectations/blob/eb1787f0e1a5ec2c2650048ee60a45d4c9738d78/lib/rspec/matchers/english_phrasing.rb#L26-L39

I'm happy to open a PR to fix, provided some guidance on the preferred solution.

reshleman:
We're currently treating all Enumerables as an Array in this method,
resulting in a Hash being formatted as a list of two-element key-value
arrays.

This results in unexpected formatting in the default failure message for
a custom matcher when the `expected` value is a Hash.

This commit updates `EnglishPhrasing.list` to no longer format a Hash as
if it were an Array.

Fixes #1192.

Before:
```
Failure/Error: expect(nil).to custom_match({a: 1, foo: 'bar', bang: Object.new})
  expected nil to custom match [:a, 1], [:foo, "bar"], and [:bang, #<Object:0x00007fb89542bad0>]
 ```

After:
```
Failure/Error: expect(nil).to custom_match({a: 1, foo: 'bar', bang: Object.new})
  expected nil to custom match {:a=>1, :bang=>#<Object:0x00007f9f363b0b28>, :foo=>"bar"}
 ```

### DIFF
--- a/lib/rspec/matchers/english_phrasing.rb
+++ b/lib/rspec/matchers/english_phrasing.rb
@@ -24,7 +24,7 @@ module RSpec
       #     list([]) #=> ""
       #
       def self.list(obj)
-        return " #{RSpec::Support::ObjectFormatter.format(obj)}" if !obj || Struct === obj
+        return " #{RSpec::Support::ObjectFormatter.format(obj)}" if !obj || Struct === obj || Hash === obj
         items = Array(obj).map { |w| RSpec::Support::ObjectFormatter.format(w) }
         case items.length
         when 0
--- a/spec/rspec/matchers/english_phrasing_spec.rb
+++ b/spec/rspec/matchers/english_phrasing_spec.rb
@@ -30,7 +30,16 @@ module RSpec
           end
         end
 
-        context "given an Enumerable" do
+        context "given a Hash" do
+          it "returns value from inspect, and a leading space" do
+            banana = { :flavor => 'Banana' }
+            expect(
+              described_class.list(banana)
+            ).to eq(" #{banana.inspect}")
+          end
+        end
+
+        context "given an Enumerable other than a Hash" do
           before do
             allow(RSpec::Support::ObjectFormatter).to(
               receive(:format).and_return("Banana")


pirj:
Thank you!
You're on the [wall of fame](https://github.com/rspec/rspec-expectations/commit/6009c7c1366b635b46670657a8a657d0f8b0ca5b#diff-9b8fab691c00b9e5380b19ce882f3271R12) now.

PR #1195, opened 2020-06-17
Factorize PredicateMatcher. Improve failure message for `Has`.

marcandre:
This PR improves `Has`'s output and makes it consistent with `BePredicate`:

1) Fixes the failure message where the result was not true or false. It used to say ", got true/false" no matter what the results are. So it could say "expected ... to return false, got true" when in actuality it's expecting the result to be falsey (upcoming PR) and it got a result of `42`!

2) Failure message for now outputs the receiver.

3) The description:
> ...is expected to have taste for "wine", "cheese"

becomes:

> is expected to have taste for "wine" and "cheese"

Misc: I also changed a feature example because it was showing a custom matcher `has_all_string_keys` which, while instructive, could be misinterpreted as an actually useful matcher when the builtin matchers are much better here (e.g. `all(be_a String)`)

This is a prelude to an upcoming PR.

### DIFF
--- a/spec/rspec/matchers/built_in/has_spec.rb
+++ b/spec/rspec/matchers/built_in/has_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe "expect(...).to have_sym(*args)" do
   it "fails if #has_sym?(*args) returns false" do
     expect {
       expect({ :b => "B" }).to have_key(:a)
-    }.to fail_with("expected #has_key?(:a) to return true, got false")
+    }.to fail_with('expected `{:b=>"B"}.has_key?(:a)` to return true, got false')
   end
 
   obj_with_block_method = Object.new
@@ -65,7 +65,7 @@ RSpec.describe "expect(...).to have_sym(*args)" do
     def o.has_some_stuff?; false; end
     expect {
       expect(o).to have_some_stuff
-    }.to fail_with("expected #has_some_stuff? to return true, got false")
+    }.to fail_with("expected `#{o.inspect}.has_some_stuff?` to return true, got false")
   end
 
   it 'includes multiple args in the failure message if multiple args were given to the matcher' do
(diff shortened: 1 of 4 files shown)


marcandre:
@JonRowe rebased, ready for review & merging

benoittgt (approved):
LGTM. Thanks for the side change on feature test. It is much clearer like that.

JonRowe:
I have concerns about inspecting the target in the message here, if the goal is to be consistent I'd almost rather go the other way and make `be` not print the inspected targets in the message, e.g. if you have a really big hash (or er, just any ActiveRecord object) then `expect(superlong_string).predicate? to return 42 got false` is going to be obscured by that string...

marcandre:
> if the goal...

The goal is to remove 30 lines of mostly redundant code with a misleading failure message to boot.

> I have concerns about inspecting the target in the message here

I don't understand your concern. `BaseMatcher#actual_formatted` is used in many other matchers, including `eq`, `be_x`, ... I'm not sure what logic it has to avoid overly long outputs, but that is orthogonal to this PR.

Please let me what I can do, as I have other yaks waiting...

marcandre:
@JonRowe What changes would you like for this to be merged?

JonRowe:
I played around with this some and couldn't produce a better output without drastically changing our current behaviour in specs, so merged as is.

marcandre:
Super thanks 🎉

spec/rspec/matchers/built_in/has_spec.rb
```diff
@@ -65,15 +65,15 @@ def obj_with_block_method.has_some_stuff?; yield; end
     def o.has_some_stuff?; false; end
     expect {
       expect(o).to have_some_stuff
-    }.to fail_with("expected #has_some_stuff? to return true, got false")
+    }.to fail_with("expected `#{o.inspect}.has_some_stuff?` to return true, got false")
```
pirj, line 68: I'm a bit out of context. Where does the `inspect` part come from? I would expect the message to be "expected `o.has_some_stuff?` to return true, got false". Is that something you were referring to in "produce a better output"?

JonRowe: Its because `actual` e.g `o` is being printed by the matcher

pirj: Got it. I didn't parse it at first.

PR #1197, opened 2020-06-18
Factorize common DynamicPredicate base.

marcandre:
The `has_...` and `be_...` dynamic predicate matchers have a lot in common.

This PR refactors both in terms of a common base class.

Output is kept the same temporarily

Precursor to #1195


lib/rspec/matchers/built_in/has.rb
```diff
@@ -2,9 +2,11 @@ module RSpec
   module Matchers
     module BuiltIn
       # @api private
```
pirj, line 4: Since this is now home for both dynamic`be_` and `has_`, maybe name the file somehow specifically?

marcandre: I'm thinking it would be good to rename `has.rb` to `dynamic_predicate.rb`. IMO the 3 classes benefit from being in the same file. To help git follow, I'd rename the file after everything else is merged (or I could do it as the very first commit).

pirj: Makes sense.
Even though probably Github won't probably be able to follow anyway, but it's worth giving it a try.

marcandre:
Rebased, just to be sure.

@JonRowe could this be merged please? Two subsequent PRs are waiting on this.

lib/rspec/matchers/built_in/has.rb
```diff
@@ -68,38 +70,87 @@ def private_predicate?
...
-            @actual.__send__(predicate, *@args, &@block)
+          def predicate_result
+            @predicate_result = actual.__send__(predicate_method_name, *@args, &@block)
           end
         end
 
-        def predicate
+        def predicate_method_name
+          predicate
+        end
```
JonRowe, line 91: I feel like this should be in the has class as it's only used there? (I've checked the other PR so I think thats the case?)

lib/rspec/matchers/built_in/has.rb
```diff
@@ -34,25 +36,25 @@ def does_not_match?(actual, &block)
...
-          [method_description, args_description].compact.join(' ')
+          "#{method_description}#{args_to_sentence}"
         end
 
       private
 
         def predicate_accessible?
-          !private_predicate? && predicate_exists?
+          @actual.respond_to? predicate
         end
```
JonRowe, line 58: This change is fine for be, but changes has behaviour?

JonRowe (changes requested):
Sorry for the delay, after having had the time to finally give this a proper review, I have two questions.

lib/rspec/matchers/built_in/has.rb
JonRowe: Ah actually, I see, its not used elsewhere in has, ignore this comment.

lib/rspec/matchers/built_in/has.rb
JonRowe: Ahh actually I see, its not used other than in validity message by `Has, but is used by `Be`.

JonRowe (approved):
Answered my own question and the remaining one is not a blocker.

PR #1203, opened 2020-08-01
Support objects that do not change with a specifier

pirj:
This is a draft of the fix.
I struggle really hard with naming, made it work, but it needs ironing out and lacks specs.

Fixes https://github.com/rspec/rspec-rails/issues/1173

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -180,7 +180,11 @@ module RSpec
 
         # @private
         def matches?(event_proc)
-          perform_change(event_proc) && @change_details.changed? && @matches_before && matches_after?
+          if perform_change(event_proc) && @change_details.changed?
+            @matches_before && matches_after?
+          else
+            @nomatches_before && matches_after?
+          end
         end
 
         # @private
@@ -212,6 +216,7 @@ module RSpec
             # returns). We need to cache these values before the `before` value
             # they are based on potentially gets mutated.
             @matches_before = values_match?(@expected_before, actual_before)
+            @nomatches_before = !values_match?(@expected_after, actual_before)
             @actual_before_description = description_of(actual_before)
           end
         end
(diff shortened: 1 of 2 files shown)


lib/rspec/matchers/built_in/change.rb
```diff
@@ -180,7 +180,11 @@ def initialize(change_details, from, to)
 
         # @private
         def matches?(event_proc)
-          perform_change(event_proc) && @change_details.changed? && @matches_before && matches_after?
+          if perform_change(event_proc) && @change_details.changed?
```
pirj, line 183: Not sure if those two should be &&'d or we'd better use a nested conditional:
```ruby
if perform_change(event_proc)
  if @change_details.changed?
    @matches_before && matches_after?
  else
    @nomatches_before && matches_after?
  end
end
```

lib/rspec/matchers/built_in/change.rb
```diff
@@ -212,6 +216,7 @@ def perform_change(event_proc)
             # returns). We need to cache these values before the `before` value
             # they are based on potentially gets mutated.
             @matches_before = values_match?(@expected_before, actual_before)
+            @nomatches_before = !values_match?(@expected_after, actual_before)
```
pirj, line 219: Naming for this variable is broken.
It says "the value or matcher passed in `to(...)` does **not** match the value before the change".

spec/rspec/matchers/built_in/change_spec.rb
```diff
@@ -356,6 +356,22 @@ def self.to_s
...
     end.to fail_with("expected `SomethingExpected#some_value` not to have changed, but did change from 5 to 6")
   end
+
+  it 'tolerates objects that do not change themselves' do
+    a = []
+    b = 1
+    expect { a << b }
+      .to change { b }
+      .to satisfy { |value| a.include?(value) }
+  end
```
myronmarston, line 366: IMO this example specifies a bug.  The expectation here is that the block changed `b` in a way that satisfies the passed matcher.  It only makes sense for this expectation to pass if `b` did in fact change. If `b` did not change, then `expect { ... }.to change { b }` should not pass no matter what further customizations are chained on the matcher.  In this example, `b` did not change (and in fact, cannot change: numeric values can never be mutated, and `b` was not reassigned to a different value).

pirj: Welcome back!

Agree. This gets especially confusing if the matcher would pass with a specifier, but would fail without one.

We actually made a change to address this very issue in a different way in #1132, but it seems it's not released as part of the latest release.

PR #1205, opened 2020-08-19
Remove rubysl

JonRowe:
This is an attempt to fix the rspec-mocks build, its the only difference in the Gemfile I can see and its mentioned in the output.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -78,10 +78,6 @@ platforms :jruby do
   end
 end
 
-platforms :rbx do
-  gem 'rubysl'
-end
-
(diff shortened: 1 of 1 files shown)


JonRowe:
This worked so huzzah

pirj:
🎉 

Is rbx alive at all?

JonRowe:
It is, but I'm thinking with RSpec 4.0 we should specify that we support MRI compatible rubies and test against only MRI, if another runtime wishes to run RSpec they should aim for MRI compatibility

PR #1208, opened 2020-08-26
Prevent be operator false positives

### LINKED ISSUE #1207: `expect(...).not_to be == ...` passes when it should fail, when certain exceptions are raised
### Subject of the issue

The operator form of `be` swallows `ArgumentError` and `NoMethodError` exceptions, so that it can fail with a nicer error message. However, this means that `expect(...).not_to be` behaves as if those exceptions are intended, resulting in passing tests when those exceptions should have caused a failure.

I have traced this back to [`BeComparedTo#matches?`](https://github.com/rspec/rspec-expectations/blob/e91c65babdfce12a5565949f5d180f90f3d928ab/lib/rspec/matchers/built_in/be.rb#L148) introduced in PR #972 to fix issue #971.

My first thought was to workaround this with `RSpec::Matchers.define_negated_matcher :not_be, :be` but that results in:

    ArgumentError:
      The expect syntax does not support operator matchers, so you must pass a matcher to `#to`.

### Your environment
* Ruby version: 2.7.1

### Steps to reproduce

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.9.0" # Activate the gem and version you are reporting the issue against.
end

require 'rspec/autorun'

class Cranky
  def ==(whatever)
    self.problem == whatever
  end
end

RSpec.describe Cranky do
  it 'defines equality based on `problem`' do
    expect(Cranky.new).not_to be == 5
  end
end
```

### Expected behavior
```txt
F

Failures:

  1) Cranky defines equality based on `problem`
     Failure/Error: self.problem == whatever

     NoMethodError:
       undefined method `problem' for #<Cranky:0x00007ff81e13fbc8>
     # tmp.rb:21:in `=='
     # tmp.rb:27:in `block (2 levels) in <main>'

Finished in 0.00174 seconds (files took 0.0622 seconds to load)
1 example, 1 failure

Failed examples:

rspec tmp.rb:26 # Cranky defines equality based on `problem`
```

### Actual behavior
```txt
.

Finished in 0.0031 seconds (files took 0.06511 seconds to load)
1 example, 0 failures
```

JonRowe:
We rescue `ArgumentError` and `NoMethodError` when checking comparison operators for objects, this is due to our desired behaviour to fail this matcher and not blow up when encountering things like `'a' < 1` or internal comparisons of attributes. However when negated these errors cause the matcher to pass, which is obviously not the intent of the rescue, so we now do that separately.

It's debatable if we should stop this behaviour in a future version if the objects being compared are more complicated methods but for now this helps.

Fixes #1207

### DIFF
--- a/lib/rspec/matchers/built_in/be.rb
+++ b/lib/rspec/matchers/built_in/be.rb
@@ -143,8 +143,13 @@ module RSpec
         end
 
         def matches?(actual)
-          @actual = actual
-          @actual.__send__ @operator, @expected
+          perform_match(actual)
+        rescue ArgumentError, NoMethodError
+          false
+        end
+
+        def does_not_match?(actual)
+          !perform_match(actual)
         rescue ArgumentError, NoMethodError
           false
         end
@@ -172,6 +177,13 @@ module RSpec
         # @return [String]
         def description
           "be #{@operator} #{expected_to_sentence}#{args_to_sentence}"
+        end
+
+      private
+
+        def perform_match(actual)
+          @actual = actual
+          @actual.__send__ @operator, @expected
         end
       end
 
--- a/spec/rspec/matchers/built_in/be_spec.rb
+++ b/spec/rspec/matchers/built_in/be_spec.rb
@@ -716,6 +716,30 @@ RSpec.describe "expect(...).not_to with comparison operators" do
       expect("foo").not_to be > "bar"
     }.to fail_with('`expect("foo").not_to be > "bar"` not only FAILED, it is a bit confusing.')
   end
+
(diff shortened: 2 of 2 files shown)


pirj (approved):
You won the race!

JonRowe:
A potential followup improvement if anyone wanted to tackle it, is to capture that an exception happened, so that it can be reported in the failure message, something like:
```
but `NameError` was encountered whilst calling `==`
```

PR #1209, opened 2020-08-26
Replace symbol array constructor

JonRowe:
I merged #1208 on a green status but it seemed the build had not passed, but not updated Github, this fixes the build.


pirj (approved):
Seems `%i[` is 2.0+ for triangle brackets-alike. 👍

PR #1212, opened 2020-08-30
Add benchmark for count vs select size

JonRowe:
From #1168


pirj:
We could use [`rubocop-performance`](https://github.com/rubocop-hq/rubocop-performance), but probably not until we drop Ruby version conditional branches.

PR #1218, opened 2020-09-28
Updates from rspec-dev (2020-09-28)

benoittgt:
These are some updates, generated from rspec-dev's rake tasks.

https://github.com/rspec/rspec-dev/pull/251

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2020-09-28T16:30:00+02:00 from the rspec-dev repo.
+# This file was generated on 2020-09-28T21:53:53+02:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # In order to install old Rubies, we need to use old Ubuntu distibution.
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -41,6 +41,8 @@ matrix:
       env: JRUBY_OPTS='--dev --1.8'
     - rvm: 2.7.1
       env: DIFF_LCS_VERSION="~> 1.3.0"
+    - rvm: 2.7.1
+      env: DIFF_LCS_VERSION="1.4.3"
```
JonRowe, line 45: Ok so this is problematic, the reason why we test for this on rspec-support is to check all the versions, but obviously `1.4.4` fixed this, and we can ignore the different changes of 1.4.3 on the other repos, but our support code doesn't cater for this...

benoittgt: Not sure I get your point. Do think we should add a test here from `DIFF_LCS_VERSION="~> 1.3.0"` and `DIFF_LCS_VERSION="1.4.3"` ?

JonRowe: It is problematic because this is failing this build.

This build works on `1.4.4` and `1.3`, there is a bug in `1.4.3` that this repo does not need to care about.

So really only `rspec-support` needs this change, I had hoped it just would be easy to sync up the changes but alas...

benoittgt: Not sure this is what you want. https://github.com/rspec/rspec-expectations/pull/1218/commits/be2b5b5244762fbb0e8fe8b084f16d6bbb3adeba

JonRowe:
Closing as stale, I'm going to take a crack at converting the last of our builds this weekend @benoittgt

Issue #1221, opened 2020-10-23
Regression in rspec-expectations 3.9.3 if argument to predicate method implements to_hash


joshcooper:
### Subject of the issue

Our spec tests starting failing with the latest rspec-expectations 3.9.3 release due to keyword argument handling changes.

### Your environment
* Ruby version:
```
$ bundle exec ruby --version
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin18]
```

* rspec-expectations version:

```
$ bundle exec gem list

*** LOCAL GEMS ***

betest (0.1.0)
bundler (default: 2.1.4)
diff-lcs (1.4.4)
rake (12.3.3)
rspec (3.9.0)
rspec-core (3.9.3)
rspec-expectations (3.9.3)
rspec-mocks (3.9.1)
rspec-support (3.9.4)
```

### Steps to reproduce

```
$ git clone https://github.com/joshcooper/betest
$ cd betest
$ bundle install
$ bundle exec rspec spec/betest_spec.rb
Betest
  passes
  fails (FAILED - 1)

Failures:

  1) Betest fails
     Failure/Error: obj.thing

     NoMethodError:
       undefined method `thing' for {:name=>"something"}:Hash
     # ./spec/betest_spec.rb:17:in `accept?'
     # ./spec/betest_spec.rb:27:in `block (2 levels) in <top (required)>'

Finished in 0.00208 seconds (files took 0.0791 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/betest_spec.rb:26 # Betest fails
```

### Expected behavior

When an object is passed to the `be_xxx` predicate matcher, it should be passed through unchanged to the predicate method.

### Actual behavior

Something in rspec's keyword argument handling results in the object's `to_hash` method getting called, the value of which is passed to the predicate method instead of the original object. Note in the test how https://github.com/joshcooper/betest/blob/680b6489441f1f0dc1f5e2d55332d449e391343b/spec/betest_spec.rb#L22 passes but https://github.com/joshcooper/betest/blob/680b6489441f1f0dc1f5e2d55332d449e391343b/spec/betest_spec.rb#L26 fails because `B` implements `to_hash`.

Downgrading to rspec-expectations 3.9.2 works around the problem

```
bundle update
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
Using rake 12.3.3
Using betest 0.1.0 from source at `.`
Using bundler 2.1.4
[...]

PR #1222, opened 2020-10-24
Switch to ruby2_keywords for passing arguments to dynamic matchers.

JonRowe:


JonRowe:
@benoittgt @pirj fixes #1221

benoittgt (approved):
Nice! LGTM and thanks @eregon for the suggestion

lib/rspec/matchers/built_in/has.rb
```diff
@@ -7,17 +7,10 @@ module BuiltIn
...
-          end
-          CODE
-        else
-          def initialize(method_name, *args, &block)
-            @method_name, @args, @block = method_name, args, block
-          end
+        def initialize(method_name, *args, &block)
+          @method_name, @args, @block = method_name, args, block
         end
+        ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
```
JonRowe, line 13: The key bit I was missing before was accidentally not including private methods when checking for ruby2_keywords I think, there is some thinking to do around how we handle this in Ruby 3 I suspect but for now this passes both the existing tests and the new one this PR adds.

eregon: Yes, that is a tricky bit (not forgetting the `, true`), it also bit me once.
Ruby 3 still has `ruby2_keywords`, so it will work fine.
On Ruby 3, `*args, **kwargs`-delegation like above works correctly, but it's probably not worth maintaining 2 versions of the method while `ruby2_keywords` exists (the current plan seems to remove it after Ruby 2 EOL).

PR #1225, opened 2020-11-14
Return true from aggregate_failures

### LINKED ISSUE #1223: Please return true/false from aggregate_failures in a custom matcher
### Subject of the issue
Make `aggregate_failures` easier to use in custom matchers defined as groups of expectations. Currently it returns `nil`, so you must end the `match` block with `true` else the matcher will always fail.

The steps below are simplified, and it could be argued that it should be separate matchers, or move the `aggregate_failures` block into the example.

However, I'm currently using this with the `all` matcher on an array of differing length (shared examples), so I need to `aggregate_failures` around each item of the array to make it easier to spot which items in the array are failing. In this way, it's not possible to group expectations into a custom matcher and run all the assertions before failing.

### Your environment
* Ruby version: `2.6.6`
* rspec-expectations version: `3.9.2`

### Steps to reproduce
```rb
context 'with custom matcher' do
  matcher :be_custom do
    match(notify_expectation_failures: true) do |actual|
      aggregate_failures do
        expect(actual).to start_with 'c'
        expect(actual).to end_with 'm'
        expect(actual).to include 'usto'
      end
    end
  end

  it { expect('custom').to be_custom }
end
```

### Expected behavior
Test passes ☺️ 

### Actual behavior
Test fails 😢 
```
with custom matcher
  is expected to be custom (FAILED - 1)

Failures:

  1) with custom matcher is expected to be custom
     Failure/Error: fit { expect('custom').to be_custom }
       expected "custom" to be custom
     # ./spec/custom_spec.rb:12:in `block (2 levels) in <top (required)>'

Finished in 0.02801 seconds (files took 0.16621 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/custom_spec.rb:12 # with custom matcher is expected to be custom
```

### Custom fix in the meantime
Return `true` after the `aggregate_failures` block. However, this means the matcher cannot be used as negated 😞 

```rb
context 'with custom matcher' do
  matcher :be_custom do
    match(notify_expectation_failures: true) do |actual|
      aggregate_failures do
        expect(actual).to start_with 'c'
        expect(actual).to end_with 'm'
        expect(actual).to include 'usto'
      end
      true # how can we detect success or failure in the block above?
    end
  end

  it { expect('custom').to    be_custom } # passes
  it { expect('hello').not_to be_custom } # fails
end
```

**Thanks very much for an awesome testing library ☺️ 🎉 **

JonRowe:
Allows the result of `aggregate_failures` to be asserted on if desired, fixes #1223

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+### Development
+[Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.10.0...main)
+
+Enhancements:
+
+* Return `true` from `aggregate_failures` when no exception occurs. (Jon Rowe, #1225)
+
 ### 3.10.0 / 2020-10-30
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.9.3...v3.10.0)
 
--- a/lib/rspec/expectations/failure_aggregator.rb
+++ b/lib/rspec/expectations/failure_aggregator.rb
@@ -80,7 +80,7 @@ module RSpec
         all_errors = failures + other_errors
 
         case all_errors.size
-        when 0 then return nil
+        when 0 then return true
         when 1 then RSpec::Support.notify_failure all_errors.first
         else RSpec::Support.notify_failure MultipleExpectationsNotMetError.new(self)
         end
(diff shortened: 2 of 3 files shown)


pirj (approved):
Awesome! ❤️

JonRowe:
Merging because failure is unrelated (windows 2.2 childprocess install problem)

PR #1245, opened 2020-12-18
Scrape out monkey patching

pirj:
Sibling PRs:
 - https://github.com/rspec/rspec-mocks/pull/1365
 - https://github.com/rspec/rspec-core/pull/2803
 - https://github.com/rspec/rspec-support/pull/462

Prerequisite PRs:
 - https://github.com/rspec/rspec-rails/pull/2420
 - https://github.com/rspec/rspec-support/pull/457
 - https://github.com/rspec/rspec-rails/pull/2425

- [x] Remove temporary commit. CI is expected to fail on Core/Mocks/Support/Rails steps, as `disable_monkey_patching!` calls methods that were removed with no replacement. I'll use the `maintenance-branch` temporary commit hack to fraternize those sibling PRs.

http://rspec.info/blog/2013/07/the-plan-for-rspec-3/#zero_monkey_patching_mode

> we do want to encourage people to switch to the new syntax, so we plan to make RSpec 3 print a warning on first usage of any the old syntax methods (should, should_not, should_receive, etc) unless the should syntax has been explicitly enabled. This should nudge folks towards the new syntax while keeping RSpec friendly to new users and will pave the way for the old syntax to be disabled by default in RSpec 4.

> zero-monkey-patching mode for RSpec...  We plan for these config options to become the defaults in RSpec 4.0, so that RSpec 4.0 will have zero monkey patching out of the box.

As for "disabled by default" vs "completely removed" and "default, out of the box" vs "impossible" I can only say that RSpec 4 was probably planned to be released earlier, as:

> we'll probably be dropping support for 1.8.7 in RSpec 4

but we've also dropped 1.9, 2.0, 2.1 and 2.2

https://github.com/rspec/rspec-core/issues/2301#issuecomment-234612382

> In RSpec 4, we plan to extract all monkey patching from RSpec and move it into a separate gem, so that monkey patching is opt-in instead of opt-out and users have to explicitly install and load a gem to get it.

`rspec-should` (or `rspec-monkey` as it's also about exposing example group DSL in the top-level/Module?) will be released later.

Those using the monkey-patched `should` syntax are not encouraged to update to RSpec 4 until this gem is extracted.

Those using the globally-exposed DSL are encouraged to use `RSpec.describe`/`RSpec.shared_examples_for` instead.


README.md
```diff
@@ -219,16 +219,15 @@ expect(
 ## `should` syntax
 
 In addition to the `expect` syntax, rspec-expectations continues to support the
-`should` syntax:
+non-monkey patching `should` syntax:
```
JonRowe, line 222: Nah, leave this as is, it makes it clear why its supported.

JonRowe, line 222: I think we should remove this syntax as well, it's confusing to keep it when we are dropping the rest of it (I know it's not monkey patching) the rspec-should can add this and optionally monkey patch.

pirj: Sounds good, it's even better than moving non-MP `should` over to `rspec-expectations` I proposed [here](https://github.com/rspec/rspec-core/pull/2800#discussion_r545910120). 👍

maintenance-branch
```diff
@@ -1 +1 @@
-4-0-dev
+remove-monkey-patching
```
JonRowe, line 1: Again, is this actually dependant on others?

lib/rspec/expectations/expectation_target.rb
```diff
@@ -73,7 +73,7 @@ def to(matcher=nil, message=nil, &block)
         # @param [String or Proc] message optional message to display when the expectation fails
         # @return [Boolean] false if the negative expectation succeeds (else raises)
         # @see RSpec::Matchers
-        def not_to(matcher=nil, message=nil, &block)
+        def not_to(matcher, message=nil, &block)
```
JonRowe, line 76: Is there a reason for this? Or is this just desired? I'm pondering the ramifications

pirj: Good question!
The reason was to allow this:
```ruby
is_expected.to > 3
```
and print a meaningful error `The expect syntax does not support operator matchers, so you must pass a matcher to #to.`
With the removal of `=nil`, it now is `wrong number of arguments (given 0, expected 1..2)` now instead.

I relied on our test suite and this change is a mistake.

Let me quickly revert this and add a spec for this message.

pirj: Fixed, thanks for noticing!

maintenance-branch
pirj: Only on Core.

features/built_in_matchers/README.md
```diff
@@ -1,13 +1,12 @@
...
-respectively on an object. Most matchers can also be accessed using the `(...).should` and
-`(...).should_not` syntax; see [using should syntax](https://github.com/rspec/rspec-expectations/blob/main/Should.md) for why we recommend using `expect`.
+respectively on an object.
 
 e.g.
 
     expect(result).to   eq(3)
     expect(list).not_to be_empty
-    pi.should be > 3
+    should be > 3
```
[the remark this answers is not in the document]
pirj: Ooops, leftover 👍
Found even more.

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -19,7 +19,7 @@
     }.to fail_with("the failure message")
```
pirj, line 19: Those examples don't belong to `extensions/kernel` here since we don't patch `Kernel`/`Object`.
Also, we describe `Object` and removed methods.

```ruby
RSpec.describe Object, "#should" do
RSpec.describe Object, "#should_not" do
```

I'd love to move this somewhere. What would you recommend, @JonRowe ?

Changelog.md
```diff
@@ -4,6 +4,7 @@
 Breaking Changes:
 
 * Ruby < 2.3 is no longer supported. (Phil Pirozhkov, #1231)
+* Extract `should*` (both monkey-patching and non-monkey-patching). (Phil Pirozhkov, #1245)
```
pirj, line 7: Ended up keeping just this, since the DSL was exposed globally in Core.

pirj:
```
      NoMethodError:
       undefined method `syntax=' for #<RSpec::Expectations::Configuration:0x000055ba304c1f18>
     # ./spec/rspec/core/configuration_spec.rb:2729:in `block in in_fully_monkey_patched_rspec_environment'
```

`4-0-dev` is ATM more like `main` in terms of monkey-patching. Are you aiming to make 4.0 somehow compatible with it?

I suggest following our practice:

5. Remove the commit from step 2. We will merge ignoring the failure.
6. Remove the commit from the other, check it passes with the other commit now on main.
7. Merge the other.
8. We will trigger builds for the ~`main`~ `4-0-dev` branch of affected repositories to check if everything is in order.

pirj:
Green 💚

pirj:
Filed https://github.com/seattlerb/minitest/issues/861

JonRowe:
> Filed seattlerb/minitest#861

I've already re-worked our build around it, its just the PR isn't merged yet

pirj:
Me too https://github.com/seattlerb/minitest/pull/862 😄

spec/rspec/expectations/extensions/kernel_spec.rb
JonRowe: The remaining `"#should"` / `"#should_not"` context specs seem to be about the protocol, a bunch of those are in `spec/rspec/expectations/syntax_spec.rb` maybe that would be a good place for those? WDYT?

The basic object ones were a should regression check when written, and could arguably be removed

pirj: It seems that it's extensively covered in `spec/rspec/matchers/dsl_spec.rb`. I'll just remove this.

pirj:
Anything left to be done here?

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-    expect(@target).to @matcher
-  end
-
-  it "asks for a failure_message when matches? returns false" do
-    expect(@matcher).to receive(:matches?).with(@target).and_return(false)
-    expect(@matcher).to receive(:failure_message).and_return("the failure message")
-    expect {
-      expect(@target).to @matcher
-    }.to fail_with("the failure message")
-  end
```
JonRowe, line 20: Did you find duplicates of these? I thought you were moving them (if so a comment saying "these are duplicates of file:line" etc is helpful when reviewing).

JonRowe (changes requested):

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-    @target = "target"
-    @matcher = double("matcher")
-    allow(@matcher).to receive(:matches?).and_return(true)
-    allow(@matcher).to receive(:failure_message)
-  end
-
-  it "accepts and interacts with a matcher" do
-    expect(@matcher).to receive(:matches?).with(@target).and_return(true)
-    expect(@target).to @matcher
-  end
```
pirj, line 12: This is covered by more specific specs:
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/expectations/handler_spec.rb#L77
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/matchers/dsl_spec.rb#L825

spec/rspec/expectations/extensions/kernel_spec.rb
pirj, line 20: Covered in:
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/expectations/handler_spec.rb#L99
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/matchers/dsl_spec.rb#L825

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-
-  it "asks for a failure_message when matches? returns false" do
-    expect(@matcher).to receive(:matches?).with(@target).and_return(false)
-    expect(@matcher).to receive(:failure_message).and_return("the failure message")
-    expect {
-      expect(@target).to @matcher
-    }.to fail_with("the failure message")
-  end
-
-  context "on interpretters that have BasicObject" do
```
pirj, line 22: We do not support such interpreters.

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-    end
-
-    it 'does not break the deprecation check on BasicObject-subclassed proxy objects' do
-      begin
-        should_enabled = RSpec::Expectations::Syntax.should_enabled?
-        RSpec::Expectations::Syntax.enable_should unless should_enabled
-        proxy_class.new(BasicObject.new).should be_proxied
-      ensure
-        RSpec::Expectations::Syntax.disable_should if should_enabled
-      end
```
pirj, line 54: We are dropping support for monkey-patching `should`.

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-    @target = "target"
-    @matcher = double("matcher")
-  end
-
-  it "accepts and interacts with a matcher" do
-    expect(@matcher).to receive(:matches?).with(@target).and_return(false)
-    allow(@matcher).to receive(:failure_message_when_negated)
-
-    expect(@target).not_to @matcher
-  end
```
pirj, line 70: Covered in:
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/matchers/dsl_spec.rb#L455
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/expectations/handler_spec.rb#L163

spec/rspec/expectations/extensions/kernel_spec.rb
```diff
@@ -1,79 +0,0 @@
...
-
-  it "asks for a failure_message_when_negated when matches? returns true" do
-    expect(@matcher).to receive(:matches?).with(@target).and_return(true)
-    expect(@matcher).to receive(:failure_message_when_negated).and_return("the failure message for should not")
-    expect {
-      expect(@target).not_to @matcher
-    }.to fail_with("the failure message for should not")
-  end
```
pirj, line 78: Covered in:
 - https://github.com/rspec/rspec-expectations/blob/bf9db76a293852a23ec4c46b55bf4ea58a66288c/spec/rspec/expectations/handler_spec.rb#L182

PR #1258, opened 2020-12-26
Make specs inter-dependent, and backwards compatible'ish

pirj:
https://github.com/rspec/rspec-core/pull/2819#issuecomment-751375468


JonRowe:
I assume your goal here is to get the rspec-core PR merged before the rspec-expectations one?

pirj:
Yes, that was my plan to merge rspec-core first 😄 
But now the order doesn't matter much - I hope they all to be green on their own. Let me quickly rebase and restart builds.

PR #1269, opened 2020-12-29
Refactor spec setup to use :apply_to_host_groups

pirj:
See: https://github.com/rspec/rspec-core/issues/2832

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -49,6 +49,7 @@ RSpec.configure do |config|
     mocks.verify_partial_doubles = true
   end
 
+  config.shared_context_metadata_behavior = :apply_to_host_groups if config.respond_to?(:shared_context_metadata_behavior=) # RSpec 4 dropped this setting
   config.disable_monkey_patching! if config.respond_to?(:disable_monkey_patching!) # RSpec 4 dropped this method
 
   # We don't want rspec-core to look in our `lib` for failure snippets.
@@ -59,7 +60,7 @@ RSpec.configure do |config|
   config.project_source_dirs -= ["lib"]
 end
 
-RSpec.shared_context "with #should enabled", :uses_should do
+RSpec.shared_context "with #should enabled" do
   orig_syntax = nil
 
   before(:all) do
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -31,35 +31,7 @@ def hash_inspect(hash)
   end
 end
 
-RSpec.configure do |config|
```
pirj, line 34: Had to move this below shared contexts.

spec/spec_helper.rb
```diff
@@ -128,4 +99,37 @@ def with_minitest_loaded
...
+    expectations.syntax = :expect
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
+    expectations.strict_predicate_matchers = true
+  end
+
+  config.mock_with :rspec do |mocks|
+    mocks.verify_partial_doubles = true
+  end
+
+  config.include_context "with #should enabled", :uses_should
```
pirj, line 120: Because they are included here now.

spec/spec_helper.rb
```diff
@@ -128,4 +99,37 @@ def with_minitest_loaded
...
+  config.include_context "with #should exclusively enabled", :uses_only_should
+  config.include_context "with warn_about_potential_false_positives set to false", :warn_about_potential_false_positives
+
+  config.shared_context_metadata_behavior = :apply_to_host_groups if config.respond_to?(:shared_context_metadata_behavior=) # RSpec 4 dropped this setting
```
pirj, line 124: And switched this to the new default.

spec/spec_helper.rb
JonRowe: Rather than moving it, split it.

spec/spec_helper.rb
JonRowe: Only these need to be below, the rest can stay up top.

spec/spec_helper.rb
JonRowe: The rest could go back up the top

JonRowe (dismissed):

PR #1277, opened 2021-01-06
Turn strict_predicate_matchers on by default

pirj:
Depends on:
 - https://github.com/rspec/rspec-mocks/pull/1397
 - https://github.com/rspec/rspec-support/pull/486

### DIFF
--- a/features/built_in_matchers/predicates.feature
+++ b/features/built_in_matchers/predicates.feature
@@ -39,7 +39,7 @@ Feature: Predicate matchers
 
   In either case, RSpec provides nice, clear error messages, such as:
 
-    `expected zero? to be truthy, got false`
+    `expected zero? to return true, got false`
 
   Calling private methods will also fail:
 
@@ -60,7 +60,7 @@ Feature: Predicate matchers
       """
     When I run `rspec be_zero_spec.rb`
     Then the output should contain "2 examples, 1 failure"
-     And the output should contain "expected `7.zero?` to be truthy, got false"
+     And the output should contain "expected `7.zero?` to return true, got false"
 
   Scenario: is_expected.not_to be_empty (based on Array#empty?)
     Given a file named "not_to_be_empty_spec.rb" with:
@@ -75,7 +75,7 @@ Feature: Predicate matchers
       """
     When I run `rspec not_to_be_empty_spec.rb`
     Then the output should contain "2 examples, 1 failure"
-     And the output should contain "expected `[].empty?` to be falsey, got true"
+     And the output should contain "expected `[].empty?` to return false, got true"
 
    Scenario: is_expected.to have_key (based on Hash#has_key?)
     Given a file named "have_key_spec.rb" with:
@@ -88,7 +88,7 @@ Feature: Predicate matchers
       """
     When I run `rspec have_key_spec.rb`
     Then the output should contain "2 examples, 1 failure"
-     And the output should contain "expected `{:foo=>7}.has_key?(:bar)` to be truthy, got false"
+     And the output should contain "expected `{:foo=>7}.has_key?(:bar)` to return true, got false"
 
    Scenario: is_expected.to have_decimals (based on custom #have_decimals? method)
      Given a file named "have_decimals_spec.rb" with:
@@ -114,7 +114,7 @@ Feature: Predicate matchers
        """
      When I run `rspec have_decimals_spec.rb`
      Then the output should contain "2 examples, 1 failure"
-      And the output should contain "expected `42.0.has_decimals?` to be truthy, got false"
+      And the output should contain "expected `42.0.has_decimals?` to return true, got false"
 
    Scenario: matcher arguments are passed on to the predicate method
      Given a file named "predicate_matcher_argument_spec.rb" with:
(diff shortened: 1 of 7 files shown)


features/built_in_matchers/predicates.feature
```diff
@@ -159,14 +159,14 @@ Feature: Predicate matchers
...
+            let(:strict) { true }
+            # deliberate failure
             it { is_expected.to have_strange_result }
           end
 
-          context 'with strict matchers' do
-            let(:strict) { true }
-            # deliberate failure
+          context 'with non-strict matchers' do
+            let(:strict) { false }
```
JonRowe, line 169: Shouldn't this change the output below?

lib/rspec/expectations/configuration.rb
```diff
@@ -126,9 +126,9 @@ def on_potential_false_positives=(behavior)
         @on_potential_false_positives = behavior
       end
 
-      # Configures RSpec to check predicate matchers to `be(true)` / `be(false)` (strict),
-      # or `be_truthy` / `be_falsey` (not strict).
-      # Historically, the default was `false`, but `true` is recommended.
+      # Configures RSpec to check the values returned by predicate matchers
+      # to be exactly true/false (strict), or truthy/be_falsey (not strict).
```
JonRowe, line 130:
```suggestion
      # to be exactly true/false (strict), or truthy/falsey (not strict).
```

features/built_in_matchers/predicates.feature
pirj: No, I've just swapped those two examples so that the new `(default)` is on top.

JonRowe: Yes but that inverts the output here, as it does above...

JonRowe: L#176 should now be `to be truthy`

spec/spec_helper.rb
```diff
@@ -70,6 +69,19 @@ def hash_inspect(hash)
 end
 RSpec.configuration.include_context "with warn_about_potential_false_positives set to false", :warn_about_potential_false_positives
 
+RSpec.shared_context "with modified configuration" do
+  around do |example|
+    configuration = example.metadata[:with_configuration]
+    key = configuration.keys.first
+    value = configuration.values.first
```
JonRowe, line 76: I mean if you're going to extract this, you might as well detect / presume the hash and map on them...

features/built_in_matchers/predicates.feature
pirj: No, the deliberate failure is for the strict option. As `has_strange_result?` returns 42, it passes with non-strict, and with it still fails with `return true`.

spec/spec_helper.rb
pirj: Sure, I've added to my todo to do that, and also to rework "with warn_about_potential_false_positives" and other shared examples that flip configuration bits.
Might even be extracted to Support if it's useful in other repos.

But I decided to leave this refactoring for later, to after 4.0.

features/built_in_matchers/predicates.feature
JonRowe: Ah I see you've just swapped the two examples 🤦

spec/spec_helper.rb
JonRowe: Please do it here first. Theres no point this being a hash and not used.

features/built_in_matchers/predicates.feature
pirj: ☕ 😄

spec/spec_helper.rb
pirj: > Theres no point this being a hash and not used

I tend to agree and started implementing it this way, but it looked quite clumsy (due to the lack of `Hash#slice` in Ruby 2.3).
Another approach could be to use two different metadata keys, `with_configuration_key` and `with_configuration_value`, quite clumsy too.

The value passed to `with_configuration` could be an array and using `first`/`last`. That imposes the question of how it should behave if we would be passing an array with more than two entries.

So, until this shared example gained some critical mass of usages of it, I decided not to grow its complexity.

JonRowe: > So, until this shared example gained some critical mass of usages of it, I decided not to grow its complexity.

It currently has unnecessary complexity for its use case in the form of the dynamic methods anyway, either simplify it so its specific to the config we're changing, or complete it please :)

JonRowe (changes requested):

spec/spec_helper.rb
pirj: Done!

pirj: I can also rework existing usages of alternative approaches but would prefer to do this later, after https://github.com/rspec/rspec-expectations/pull/1253 is merged and after we remove `include_chain_clauses_in_custom_matcher_descriptions` et al to avoid unnecessary work.

spec/spec_helper.rb
```diff
@@ -72,12 +72,17 @@ def hash_inspect(hash)
 RSpec.shared_context "with modified configuration" do
   around do |example|
     configuration = example.metadata[:with_configuration]
-    key = configuration.keys.first
-    value = configuration.values.first
-    original = RSpec::Expectations.configuration.public_send(key)
-    RSpec::Expectations.configuration.public_send("#{key}=", value)
+    # Rework to use `slice` when we drop support for Ruby 2.3
```
JonRowe, line 75: What would `slice` give us here, it just drops keys?

pirj: Good catch. `RSpec::Expectations.configuration` isn't a Hash, so calling `slice` on it is impossible.
Removed this note.

JonRowe (approved):

pirj:
cc @marcandre

marcandre:
Great 🎉

PR #1286, opened 2021-01-28
Remove support for present tense predicates

### LINKED ISSUE #935: Remove support from dynamic predicate matchers for present-tense predicate
https://github.com/rspec/rspec-expectations/blob/a271631e9b8e1749be876dbcc65dbd7a6a40d2db/lib/rspec/matchers/built_in/be.rb#L248-L250

It's confusing and can lead to bugs like rspec/rspec-rails#1681.  It should be removed from RSpec 4 and deprecated in RSpec 3.99.

pirj:
Fixes #935


pirj:
Green except for a cryptic jruby-9.1 failure: `private method `require' called for Kernel:Module`.

JonRowe (approved):
In general this is good, needs the PR number and I've suggested a tweak to the specs but I don't mind what you do.

PR #1285, opened 2021-01-28
Restrict block matchers use with value expectations

pirj:
https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html 
https://rspec.rubystyle.guide/#implicit-block-expectations

Spec changes are due to:

```ruby
   matcher.matches?(invalid_value)
```

doesn't work with block-only matchers, as no block is passed in, and it fails with a message:

       1) RSpec::Matchers::BuiltIn::Change behaves like an RSpec block-only
    matcher uses the `ObjectFormatter` for `failure_message`
         Failure/Error: expect(message).to include("detailed inspect")
           expected "expected `@k` to have changed, but was not given a block"
    to include "detailed inspect"

The redundant (due to existing check in ExpectationTarget) `Proc === @event_proc` checks were removed, as an error is raised earlier than those checks.

Also see:
 - https://github.com/rspec/rspec-expectations/pull/1139
 - https://github.com/rspec/rspec-expectations/pull/1125

PR #1288, opened 2021-02-10
Remove a couple of remaining deprecations

pirj:


lib/rspec/matchers/built_in/change.rb
```diff
@@ -365,11 +363,6 @@ def initialize(matcher_name, receiver=nil, message=nil, &block)
           @receiver = receiver
           @message = message
           @value_proc = block
-          # TODO: temporary measure to mute warning of access to an initialized
-          # instance variable when a deprecated implicit block expectation
-          # syntax is used. This may be removed once `fail` is used, and the
-          # matcher never issues this warning.
-          @actual_after = UNDEFINED
```
JonRowe, line 372: Is this right?

pirj: Yes, previously we weren't initializing it https://github.com/rspec/rspec-expectations/blame/bd10f0cf3970932781efcd09b5e427877d16a6c2/lib/rspec/matchers/built_in/change.rb#L353

JonRowe: No I mean the removal in this PR

pirj: Oh. Yes. I've overlooked to remove this in #1285 after resolving conflicts with #1139.

I think this PR is the last change in Expectations to get it ready for the 4.0 release (apart from adding deprecation warnings to `main` and rebasing `4-0-dev` on top of it).

Sanity check, however, shows that nothing fails on https://github.com/rspec/rspec-expectations/tree/fe0c4b7f (#1139) even with the removed `@actual_after = UNDEFINED`. Can't recall the exact reason I've added it in the first place.

pirj: I'll extract this removal into a separate PR.

pirj: Done: #1289

JonRowe (approved):

PR #1292, opened 2021-02-19
Fix YARD type syntax

kachick:
Both does not raise error in `bundle exec yard`, but looks wrong syntax in https://yardoc.org/types.html.

スクリーンショット 2021-02-20 5 35 29
スクリーンショット 2021-02-20 5 36 07

Or https://github.com/lsegal/yard-types-parser is wrong? Looks an ancient project 😓 

スクリーンショット 2021-02-20 5 47 03

### DIFF
--- a/lib/rspec/expectations/expectation_target.rb
+++ b/lib/rspec/expectations/expectation_target.rb
@@ -57,7 +57,7 @@ module RSpec
         #   expect { perform }.to raise_error
         # @param [Matcher]
         #   matcher
-        # @param [String or Proc] message optional message to display when the expectation fails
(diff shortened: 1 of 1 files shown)


kachick:
The site is using this 🤔 

https://github.com/lsegal/yardoc.org/blob/62841134db091994fab292978c885dcf87af6863/public/js/types.js#L73-L161

JonRowe (approved):

JonRowe:
Thanks that does seem to be a mistake on our part, this also seems to be the only place we do it.

kachick:
Thank you!

PR #1294, opened 2021-03-05
Fix support for dynamic matchers

pirj:
See https://github.com/rspec/rspec-collection_matchers/pull/48

If a matcher supports dynamic arbitrary methods called on it, e.g.  `have_at_least(100).bucks`, it would also consider a `supports_value_expectations?` call as a dynamic part.

Instead of calling `supports_*?` methods on a matcher, explicitly check if it responds to it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,8 +6,13 @@ Enhancements:
 * Return `true` from `aggregate_failures` when no exception occurs. (Jon Rowe, #1225)
 
 Deprecations:
+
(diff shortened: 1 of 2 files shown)


pirj:
@JonRowe Can I consider your approval for https://github.com/rspec/rspec-expectations/pull/1295 as an approval for this PR as well? I believe fixing `main` for those who use `rspec-collection_matchers` on Rails 6.1, `rspec-rails` `4.1.0.pre` was the primary objective.

PR #1295, opened 2021-03-05
Fix support for dynamic matchers

pirj:
See https://github.com/rspec/rspec-collection_matchers/pull/48

If a matcher supports dynamic arbitrary methods called on it, e.g.  `have_at_least(100).bucks`, it would also consider a `supports_value_expectations?` call as a dynamic part.

Instead of calling `supports_*?` methods on a matcher, explicitly check if it responds to it.

Symmetrical change for `main` that suffers when `rspec-rails` `4.1.0.pre` is used and `rspec-expectations` `3.11.0.pre` is at https://github.com/rspec/rspec-expectations/pull/1294

### DIFF
--- a/lib/rspec/expectations/expectation_target.rb
+++ b/lib/rspec/expectations/expectation_target.rb
@@ -121,9 +121,7 @@ module RSpec
       end
 
       def supports_value_expectations?(matcher)
-        matcher.supports_value_expectations?
-      rescue NoMethodError
-        true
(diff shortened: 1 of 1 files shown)


pirj:
Pretty weird, several builds failed with
```
Fetching gem metadata from https://rubygems.org/....
Retrying dependency api due to error (2/4): Bundler::HTTPError Net::HTTPBadGateway: I/O error
```
while https://status.rubygems.org/ reports no incidents. Re-running.

JonRowe (approved):
A shame but sure 😂

PR #1296, opened 2021-04-07
EQ matcher to return diff when testing numbers

mcyoung:
When using the `.eq` matcher to test numbers this allows there to be feedback on the difference between `expected` and `actual` results.  This allows for updating tests that use this matcher to be done in a quicker way than manually checking the results.

For example, if you've got a test written to verify that the total of a method is intentionally `1234`, but then update the method and your tests fails with:
```
   expected: 1234
        got: 1983
       diff: -749
```
the developer can much quicker know if this code change was doing what they wanted (in knowing that the difference of `-749` was an expected result) and therefore update their test suite accordingly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 gemspec
 
 branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
-%w[rspec rspec-core rspec-mocks rspec-support].each do |lib|
+%w[rspec rspec-mocks rspec-support].each do |lib|
   library_path = File.expand_path("../../#{lib}", __FILE__)
   if File.exist?(library_path) && !ENV['USE_GIT_REPOS']
     gem lib, :path => library_path
@@ -11,6 +11,7 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
     gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => branch
   end
 end
+gem 'rspec-core', :git => "https://github.com/mcyoung/rspec-core.git", :branch => "eq-diff-matcher"
 
 if RUBY_VERSION < '1.9.3'
   gem 'rake', '< 11.0.0' # rake 11 requires Ruby 1.9.3 or later
--- a/features/aggregating_failures.feature
+++ b/features/aggregating_failures.feature
@@ -35,6 +35,7 @@ Feature: Aggregating Failures
 
         1) expected: 200
                 got: 404
+               diff: -204
 
            (compared using ==)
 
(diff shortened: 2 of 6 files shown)


pirj:
May I kindly ask to tell about the workflow in more detail, what is the test where this will be used, what is the code under test?

Subjectively,
```
          |      expected: 2
          |           got: 3
          |          diff: -1
```
is wrong. I expect to have one moon in the sky. There are two. This is one more than I have expected (+1).

Also, the diff with `nil` is weird.

mcyoung:
> May I kindly ask to tell about the workflow in more detail, what is the test where this will be used, what is the code under test?
> 
> Subjectively,
> 
> ```
>           |      expected: 2
>           |           got: 3
>           |          diff: -1
> ```
> 
> is wrong. I expect to have one moon in the sky. There are two. This is one more than I have expected (+1).
> 
> Also, the diff with `nil` is weird.

An example of the testing scenario would be having a method that sums up line items, calculated sales tax, discount applications and provides a specific total:
```ruby
def grand_total( items, tax, discount )
  total = items.sum(&:amount)
  total -= discount.amount
  total = total * tax
  total.round(2)
end
# ---------------------

let(:items) { [ {amount: 323}, {amount: 154}, {amount: 481} ] }
let(:tax) { 0.075 }
let(:discount) { 15 }

it "provide the correct total" do
  expect(Calculator.grand_total(items, tax, discount)).to eq(1_046.14)
end
```

This would provide the output of:
```sh
      expected: 1046.14
           got: 1029.85
          diff: -16.29
```

What this gives the programmer is the ability to quickly grok that diff number (in this case `-16.29`) with out having to manually calculate that difference.  This can be beneficial in knowing _"hey that `-16.29` is my expected difference and because the tax value is indeed changing to `0.92` and I'm seeing that in these 6 other tests.  Therefore I'm good to update my test suite"_.

A great point, that perhaps it should be `actual - expected` so that if you have more than expected, you'll have positive, and conversely, less than expected would be negative.  I like that.

Also, your point on `nil` is valid.  I debated a bit on what to do with that scenario, so maybe it shouldn't show a diff as if it's a `0`, but rather just not show the `diff`

pirj:
Will creating a custom matcher, say `eq_with_diff` based on `RSpec::Matchers::BuiltIn::Eq` work for you? `Eq` is marked as private, but its interface is unlikely to be changed in the near future.

mcyoung:
I suppose that would be a way of going about it.  I personally am just overriding this method locally in my projects for the time being.  I still feel that having this included would be a good improvement.

JonRowe:
It would be an overloaded term for us to call this the "diff", it's not, the diff is:

```diff
- <expected>
+ <actual>
```

Its the delta, we also don't do this in any other matcher, instead leveraging a diff tool (diff_lcs currently)

pirj:
Closing as stale

PR #1303, opened 2021-06-06
Follow `frozen_string_literal` and `equal` and `be` matchers behavior in docs

kachick:
I understand this PR just updates documentations part of #997

### DIFF
--- a/features/built_in_matchers/equality.feature
+++ b/features/built_in_matchers/equality.feature
@@ -92,6 +92,8 @@ Feature: Equality matchers
   Scenario: compare using equal (equal?)
     Given a file named "compare_using_equal.rb" with:
       """ruby
+      # frozen_string_literal: false
+
       RSpec.describe "a string" do
         it "is equal to itself" do
           string = "this string"
@@ -114,6 +116,8 @@ Feature: Equality matchers
   Scenario: compare using be (equal?)
     Given a file named "compare_using_be.rb" with:
       """ruby
+      # frozen_string_literal: false
+
       RSpec.describe "a string" do
         it "is equal to itself" do
           string = "this string"
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -582,7 +582,7 @@ module RSpec
     #
     # @example
     #   expect(5).to       equal(5)   # Integers are equal
-    #   expect("5").not_to equal("5") # Strings that look the same are not the same object
+    #   expect(5).not_to equal(5.0) # Integer and Float are not the same object
     def equal(expected)
       BuiltIn::Equal.new(expected)
     end


lib/rspec/matchers.rb
```diff
@@ -582,7 +582,7 @@ def eql(expected)
     #
     # @example
     #   expect(5).to       equal(5)   # Integers are equal
-    #   expect("5").not_to equal("5") # Strings that look the same are not the same object
```
pirj, line 585: I believe there is no point to compare two string literals, so this comment describes the case when some non-literal actual value is compared with `be_equal` with an expected literal might result in a failure.

BTW, the use of `equal` instead of `be_equal` seems to be the real problem here.

features/built_in_matchers/equality.feature
```diff
@@ -92,6 +92,8 @@ Feature: Equality matchers
   Scenario: compare using equal (equal?)
     Given a file named "compare_using_equal.rb" with:
       """ruby
+      # frozen_string_literal: false
```
pirj, line 95: Why?

lib/rspec/matchers.rb
kachick: Sure.

```suggestion
    #   expect(+"5").not_to equal(+"5") # Strings that look the same are not the same object
```

So this is a way (or use `dup`). Is it better?

features/built_in_matchers/equality.feature
kachick: As far as I know, `# frozen_string_literal: false` makes the old ruby behaviors for string literals.
Recent projects basically prefers `# frozen_string_literal: true`.  Then ruby handles same string literal as a same object.

So clarifying to set `false` does not make confusion when reading docs as a tutorial. I think.

lib/rspec/matchers.rb
pirj: What are you after with this change in the first place?

kachick: https://github.com/rspec/rspec-expectations/pull/1303#discussion_r646165448 is the my motivation

features/built_in_matchers/equality.feature
pirj: We use `.feature` files for two purposes:

1. Documentation, that is published to https://relishapp.com/rspec/
I'd prefer not to show an example that the usage of `frozen_string_literal` makes sense for spec files, as I haven't seen any evidence of that.

2. End-to-end tests of our code
Is there a good reason to explicitly specify the already default non-frozen string literal in our own tests?

lib/rspec/matchers.rb
pirj: To me, the current doc says "don't use `equals` unless you absolutely know what you're doing, e.g. if you compare identical strings that are not the same object, the comparison would fail".
Do you think adding `+` would help to make this more clear for developers of all experience levels?

kachick: Ah thanks, that is the documentation direction of this project, I understood.

features/built_in_matchers/equality.feature
kachick: Thanks for letting me know the `cucumber` role in this project.

Personally I don’t have any problem. So I will left from this issue!

pirj:
Thanks for understanding.

PR #1322, opened 2021-09-09
Fix `expect(Array).to include(Hash).times` case

ojab:
`#expected` has a [special case](https://github.com/rspec/rspec-expectations/commit/8693ecb75) for single `Hash` and returning this
single `Hash` instead of `[Hash]`, so `expected.first` returns first
`key => value` pair and matcher suddenly fails

### DIFF
--- a/features/built_in_matchers/include.feature
+++ b/features/built_in_matchers/include.feature
@@ -150,3 +150,31 @@ Feature: `include` matcher
       | expected {:a => 7, :b => 5} not to include :a                 |
       | expected {:a => 7, :b => 5} to include {:d => 3}              |
       | expected {:a => 7, :b => 5} not to include {:a => 7}          |
+
+  Scenario: counts usage
+    Given a file named "include_matcher_with_counts_spec.rb" with:
+      """ruby
+        RSpec.describe [{:c => 7}, {:a => 1}, {:b => 2}, {:c => 1}, {:a => 3}, {:c => 7}] do
+          it { is_expected.to include(:b => 2).exactly(1).times }
+          it { is_expected.to include(:b => 2).once }
+          it { is_expected.to include(have_key(:a)).twice }
+          it { is_expected.to include(have_key(:c)).at_least(2).times }
+          it { is_expected.to include(have_key(:a)).at_least(:once) }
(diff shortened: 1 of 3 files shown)


ojab:
Marking as ready for review because of `First-time contributors need a maintainer to approve running workflows. Learn more.`

pirj (approved):
Nicely done, thank you.
Would you like to add this to the documentation at `features/built_in_matchers/include.feature`, too?

ojab:
Not sure where/what exactly I should put there, added single example to the `hash usage`.

ojab:
oh, and this addition is bogus, it's about matching `actual` hash to some included keys. Apparently my brain doesn't work at this time.

So yeah, should it be just new example in a ruby codeblock or some new test like `Scenario: count usage`?

pirj:
> should it be just new example in a ruby codeblock or some new test like `Scenario: count usage`?

It's a corner case on the "used against a list of hashes" and "when used with expectation counts". I'm on the fence if it needs its own focused scenario or doesn't need one at all since without counts `expect([{ 'a' => 1 }]).to include('a' => 1)` works just fine. Leaving it up to you.

features/built_in_matchers/include.feature
```diff
@@ -108,6 +108,7 @@ Feature: `include` matcher
         it { is_expected.to include(:a) }
         it { is_expected.to include(:b, :a) }
         it { is_expected.to include(:a => 7) }
+        it { is_expected.to include(:b => 5).once }
```
JonRowe, line 111: This doesn't actually pass

JonRowe (commented):
This is a reasonable improvement, but the scenario line added actually raises an error, which is why the build is broken, can you take a look?

ojab:
> First-time contributors need a maintainer to approve running workflows.
:okay:
both specs & cucumber is green locally

JonRowe:
Thanks @ojab sorry it took so long to get back to this

ojab:
Oh, I already forgot about this PR :D  
Anyway, not problem at all, I understand that maintainership is a hard & voluntary work, thanks for  doing that.

PR #1324, opened 2021-10-04
Fix Include Matcher For Ranges

bclayman-sq:
This PR addresses issue rspec/rspec#138 for ruby >= 2.1.9.

Previously, a few parts of the `Include` matcher assumed all `Range`s were iterable.  This caused it to raise errors like `TypeError: Can't iterate from [Float|Time]`

This happens because `Range`s require that their beginning element implement `succ`.  `Float` doesn't, which causes the error.  `Time` is different because it does implement `succ` but

a) `Range#succ` is deprecated as of ruby 1.9.2 and
b) Some Ruby implementations raise an exception when trying to iterate through a range of `Time` objects

This PR does a few things:
1) Fixes the `Include` matcher to handle `Range`s that don't support iteration, while continuing to support `Range`s that do
2) Adds specs for both types of `Range`s in 1).  There weren't any for the `Include` matcher used with `Range`s.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -170,6 +170,9 @@ module RSpec
           # String lacks an `any?` method...
           return false unless actual.respond_to?(:any?)
 
+          # Some objects don't support iteration (e.g. Float, Time, etc.)
+          return false if non_iterable_range?(actual)
+
           actual.any? { |value| values_match?(expected_item, value) }
         end
 
@@ -179,8 +182,21 @@ module RSpec
           end
         else
           def count_enumerable(expected_item)
+            return actual.include?(expected_item) ? 1 : 0 if non_iterable_range?(actual)
             actual.count { |value| values_match?(expected_item, value) }
           end
+        end
+
+        # Determines if actual is a range and contains elements that do not support iteration.
+        # The usual requirement for a Range supporting iteration is that its beginning element
+        # implements `succ`. Time is different; it implements `succ` but that method is
+        # deprecated as of 1.9.2.  Attempting to iterate through a Range of Times raises a
+        # TypeError on many common Ruby implementations.  We treat Ranges of Time objects as
+        # non-iterable to ensure safety.
+        def non_iterable_range?(actual)
+          # leave old behavior untouched, but fix for modern rubies
+          return false unless RUBY_VERSION >= "2.1.9"
+          actual.is_a?(Range) && (!actual.min.respond_to?(:succ) || actual.min.is_a?(Time))
         end
 
         def count_inclusions
--- a/spec/rspec/matchers/built_in/include_spec.rb
+++ b/spec/rspec/matchers/built_in/include_spec.rb
@@ -161,6 +161,20 @@ RSpec.describe "#include matcher" do
     end
   end
 
+  shared_context "time range" do
+    let(:fmt) { RSpec::Support::ObjectFormatter }
+    let(:now) { Time.now.utc }
+    let(:start) { now - 20 }
+    let(:finish) { now - 10 }
+    let(:range) { (start..finish) }
+  end
+
+  shared_context "only runs for modern rubies" do
+    around(:example) do |example|
+      example.run if RUBY_VERSION >= "2.1.9"
+    end
+  end
+
   describe "expect(...).to include(with_one_arg)" do
     it_behaves_like "an RSpec value matcher", :valid_value => [1, 2], :invalid_value => [1] do
       let(:matcher) { include(2) }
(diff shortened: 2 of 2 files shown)


lib/rspec/matchers/built_in/include.rb
```diff
@@ -163,6 +163,9 @@ def actual_collection_includes?(expected_item)
           # String lacks an `any?` method...
           return false unless actual.respond_to?(:any?)
 
+          # Some objects don't support iteration (e.g. Float, Time, etc.)
+          return actual.include?(expected_item) if non_iterable_range?(actual)
```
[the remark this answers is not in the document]
bclayman-sq: Ah yeah, good callout!  Let me fix this 👍

pirj:
Does it make sense to rebase this on `4-0-dev` and get rid of Ruby < 2.3 checks? @JonRowe do we have a timeline for 4.0 release?

bclayman-sq:
> Does it make sense to rebase this on `4-0-dev` and get rid of Ruby < 2.3 checks? @JonRowe do we have a timeline for 4.0 release?

👋 Hi @JonRowe, I wanted to follow up on this and to get your review!

pirj:
It may take a while. We're making our best effort, but often times life and daily job intervenes, and it [may take longer](https://ossert.evilmartians.io/rspec-expectations). Processing has improved this year, though:

bclayman-sq:
@JonRowe @pirj 👋 Anything I can do to get this PR merged?

pirj (approved):
Thank you!

What do you think of re-targeting it for RSpec 4 where we've dropped support for Ruby < 2.2? It would require rebasing on top of `4-0-dev` and chaning the base branch of the PR.

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -111,6 +111,20 @@ def hash.send; :sent; end
...
+    let(:fmt) { RSpec::Support::ObjectFormatter }
+    let(:now) { Time.now.utc }
+    let(:start) { now - 20 }
+    let(:finish) { now - 10 }
+    let(:range) { (start..finish) }
+  end
+
+  shared_context "only runs for modern rubies" do
+    around(:example) do |example|
+      example.run if RUBY_VERSION >= "2.1.9"
```
JonRowe, line 124: Can we change this for an expected error?

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -111,6 +111,20 @@ def hash.send; :sent; end
...
 
+  shared_context "time range" do
+    let(:fmt) { RSpec::Support::ObjectFormatter }
+    let(:now) { Time.now.utc }
+    let(:start) { now - 20 }
+    let(:finish) { now - 10 }
+    let(:range) { (start..finish) }
+  end
+
+  shared_context "only runs for modern rubies" do
```
JonRowe, line 122: If we do the below this should be renamed

JonRowe (changes requested):
I'm happy to merge this once we've addressed what happens on older Rubies, I don't mind it "not working", I'd just like the specs to show what happens on older Rubies, even if thats just this (or equivalent):
```
if RUBY_VERSION >= "2.1.9"
  example.run 
else
  expect { example.run }.to raise_error(WhateverErrorIsRaised)
end
```

spec/rspec/matchers/built_in/include_spec.rb
bclayman-sq: @JonRowe Definitely!

I just pushed a PR that runs examples for all ruby versions so I can see the error across older rubies (the retention window for this PR's [build logs](https://github.com/rspec/rspec-expectations/runs/3797092128?check_suite_focus=true) has elapsed).

Mind kicking off another build, I'll see which errors crop up and figure out where to go from there?

JonRowe: Done, I wish Github remembered approvals, I assume it goes off "merged commiters"

bclayman-sq: @JonRowe Awesome, I just pushed those changes (it looks like old rubies raise a `TypeError` so that's what I now expect to get raised for those versions).  

Mind kicking off one more build?

Also, I'm not 100% sure that my naming for the shared examples is best.  Open to suggestion there!

bclayman-sq:
> I'm happy to merge this once we've addressed what happens on older Rubies, I don't mind it "not working", I'd just like the specs to show what happens on older Rubies, even if thats just this (or equivalent):
> 
> ```
> if RUBY_VERSION >= "2.1.9"
>   example.run 
> else
>   expect { example.run }.to raise_error(WhateverErrorIsRaised)
> end
> ```

Hi @JonRowe,

I'd like to suggest going with the approach that only runs these new tests for ruby >= 2.1.9.  This covers all ruby versions in the last 9 years and fixes the bug.  How would you feel about that?

Here's a bit of technical discussion about why I don't think the proposed approach quite works:

I don't think there's a straightforward way to expect examples to raise a `TypeError` and only a `TypeError`.  This code:
```
if RUBY_VERSION >= "2.1.9"
  example.run 
else
  expect { example.run }.to raise_error(WhateverErrorIsRaised)
end
```

actually causes 2 failures for specs run on pre-2.1.9 ruby:
```
Got 1 failure and 1 other error:

20.1) Failure/Error: expect(range).not_to include(2.0).twice

TypeError:
  can't iterate from Float
# ./spec/rspec/matchers/built_in/include_spec.rb:797:in `block (7 levels) in <top (required)>'
# ./spec/rspec/matchers/built_in/include_spec.rb:127:in `block (4 levels) in <top (required)>'
# ./spec/rspec/matchers/built_in/include_spec.rb:127:in `block (3 levels) in <top (required)>'

20.2) Failure/Error: expect { example.run }.to raise_error(TypeError)
  expected TypeError but nothing was raised
# ./spec/rspec/matchers/built_in/include_spec.rb:127:in `block (3 levels) in <top (required)>'
```
This is because the original expectation is evaluated first and isn't met (it raises a `TypeError`).  Then the second expectation _also_ isn't met because it expects a `TypeError` to be raised and this has already happened.

Do you know a way around this?  If not, I'd love to merge as is since it still represents a solid improvement.  Let me know what you think!

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -161,6 +161,24 @@ def hash.send; :sent; end
...
+  shared_context "runs for modern rubies, raises for old rubies" do
+    around(:example) do |example|
+      if RUBY_VERSION >= "2.1.9"
+        example.run
+      else
+        expect { example.run }.to raise_error(TypeError)
+      end
+    end
+  end
+
```
pirj, line 181:
```suggestion
```

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -198,6 +216,165 @@ def hash.send; :sent; end
       end
     end
 
+    context "for a range target" do
+      include_context "runs for modern rubies, raises for old rubies"
```
pirj, line 220:
```suggestion
      before { pending "raises for old rubies" if RUBY_VERSION < "2.1.9" }
```

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -394,6 +571,93 @@ class PseudoHash < SimpleDelegator
       expect(matcher.description).to eq("include \"a\"")
     end
 
+    context "for a range target" do
+      include_context "runs for modern rubies, raises for old rubies"
```
pirj, line 575:
```suggestion
      before { pending "raises for old rubies" if RUBY_VERSION < "2.1.9" }
```

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -530,6 +794,168 @@ class PseudoHash < SimpleDelegator
       end
     end
 
+    context "for a range target" do
+      include_context "runs for modern rubies, raises for old rubies"
```
pirj, line 798:
```suggestion
      before { pending "raises for old rubies" if RUBY_VERSION < "2.1.9" }
```

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -613,10 +1039,111 @@ class PseudoHash < SimpleDelegator
         }.to fail_matching('expected {:key => "value"} not to include :key')
       end
     end
-
   end
 
   describe "expect(...).not_to include(with, multiple, args)" do
+    context "for a range target" do
+      include_context "runs for modern rubies, raises for old rubies"
```
pirj, line 1046:
```suggestion
      before { pending "raises for old rubies" if RUBY_VERSION < "2.1.9" }
```

spec/rspec/matchers/built_in/include_spec.rb
```diff
@@ -953,6 +1480,40 @@ def matches?(_)
       end
     end
 
+    describe "expect(range).to include(matcher)" do
+      include_context "runs for modern rubies, raises for old rubies"
```
pirj, line 1484:
```suggestion
      before { pending "raises for old rubies" if RUBY_VERSION < "2.1.9" }
```

JonRowe:
This has been migrated to the monorepo, given that the blockers were largely legacy rubys that RSpec 4 won't support I hope to merge this in relatively quickly, sorry it got forgotten about

Issue #1161, opened 2020-02-13
match_array is taking too long to finish


RicardoTrindade:
### Subject of the issue

Was writing some tests in my project, was using match_array to make some expectations and noticed that it causing some tests to never finish.

### Your environment
* Ruby version: ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]
* rspec-expectations version: 3.9.0

### Steps to reproduce

Try and run this test locally and see if it finishes (it will most likely fail nonetheless)
```
    it do
      a = Array.new(50) { rand(1...9) }
      b = Array.new(50) { rand(1...9) }
      expect(a).to match_array(b)
    end
```

### Expected behavior

The test should finish running (regardless of pass/fail)
### Actual behavior

The test never finishes or takes too long to finish (with smaller arrays)

pirj:
Kind of relates to https://github.com/rspec/rspec-expectations/issues/685 and https://github.com/rspec/rspec-expectations/issues/577
Do you think with those things in mind there a good algorithm that would handle matching in linear time?

diei:
I ran into same problem. I use Ruby 2.7.1 and rspec-expectations 3.10.1. Even arrays with only 30 items takes much too long to finish.

pirj:
Would you like to make an analysis of the time complexity of the algorithm, @diei?
Does the fact that it fails or matches affect the time to complete?
Let's dissect this.
Here is the source to get you going https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/built_in/contain_exactly.rb#L83

diei:
I think the complexity should be factorial.

`best_solution_for_pairing` is called `actuals` size times:
https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/built_in/contain_exactly.rb#L236

And in `best_solution_for_pairing` a new `PairingsMaximizer` is created and `find_best_solution` is called with the actuals and expected arrays reduced by one item.
https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/built_in/contain_exactly.rb#L288

In best case `match_array` finishes very fast, even with arrays having 1000 items.

pirj:
[...]

Issue #1006, opened 2017-08-02
`contain_exactly` matcher spins forever, never finishing whatever it's supposed to be doing


trejkaz:
For some reason, this example never completes:

```
describe 'something' do
  it 'fails this test' do
[...]

PR #1325, opened 2021-10-06
Improve ContainExactly matcher speed when elements obey transitivity

bclayman-sq:
This addresses issues #1006, #1161.

The current implementation for `ContainExactly` runs in O(n!).  The crux of the problem is that some elements don't obey transitivity.  As a result, knowing that sorting actual and expected doesn't result in a match *doesn't* guarantee that expected and actual don't match.

This PR makes improvements that ensure any test with comparable elements in `actual` and `expected` runs in O(n log n) time. 

There are two main updates:

**Update the core logic in `ContainExactly` to:**

1. Attempt to sort `actual` and `expected`
2. If the sort succeeds, this tells us the objects are comparable.  Use the result of the sort

This means that when `actual` and `expected` contain comparable elements, we can immediately return `false` when the sorted arrays don't match.  **This speeds up passing tests where `actual` and `expected` don't match from O(n!) to O(n log n).**

**Speed up determining `extra` and `missing` elements**

Previously, the logic for determining `extra` and `missing` items between `actual` and `expected` relied on code in `PairingsMaximizer` that generated all possible element pairings.  So even if you avoided comparing all possible pairings to determine if the matcher should return true, you would still incur that O(n!) cost when generating a failure message.

The code now calculates `extra` and `missing` in linear time.  **This speeds up failing tests from O(n!) to O(n log n).**

More practically, this means that common use cases for `ContainExactly` will enjoy a massive speedup.  Previously, users have examples where comparing arrays of 30 integers "never finishes." With this PR's update, comparing arrays of 10,000 integers runs in < 0.1s on my machine.

### DIFF
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -6,19 +6,6 @@ module RSpec
       # Provides the implementation for `contain_exactly` and `match_array`.
       # Not intended to be instantiated directly.
       class ContainExactly < BaseMatcher
-        def initialize(expected=nil)
-          super
-          @transitive = false
-        end
-
-        # @api public
-        # Specifies that elements contained in actual and expected
-        # obey transitivity.  This lets match run much faster.
-        def transitive
-          @transitive = true
-          self
-        end
-
         # @api private
         # @return [String]
         def failure_message
@@ -100,6 +87,7 @@ module RSpec
         # the slowness of the full matching algorithm, and in common cases this
         # works, so it's worth a try.
         def match_when_sorted?
+          @transitive = true # Be optimistic. Let `safe_sort` reveal non-transitivity.
           @sorted_expected, @sorted_actual = safe_sort(expected), safe_sort(actual)
           values_match?(@sorted_expected, @sorted_actual)
         end
@@ -117,7 +105,7 @@ module RSpec
         def safe_sort(array)
           array.sort
         rescue Support::AllExceptionsExceptOnesWeMustNotRescue
-          raise "Invalid use of `.transitive` with unsortable array #{array}" if @transitive
+          @transitive = false
           array
         end
 
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -86,8 +86,12 @@ module RSpec
 
         def match(_expected, _actual)
           return false unless convert_actual_to_an_array
+
           matched_when_sorted = match_when_sorted?
-          return matched_when_sorted if matched_when_sorted || @transitive
+
+          return true if matched_when_sorted
+          return matched_when_sorted if @transitive
+
           (extra_items.empty? && missing_items.empty?)
         end
 
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -6,6 +6,19 @@ module RSpec
       # Provides the implementation for `contain_exactly` and `match_array`.
       # Not intended to be instantiated directly.
       class ContainExactly < BaseMatcher
+        def initialize(expected=nil)
+          super
+          @transitive = false
+        end
+
+        # @api public
+        # Specifies that elements contained in actual and expected
+        # obey transitivity.  This lets match run much faster.
+        def transitive
+          @transitive = true
+          self
+        end
+
         # @api private
         # @return [String]
         def failure_message
@@ -36,6 +49,7 @@ module RSpec
         def generate_failure_message
           message = expected_collection_line
           message += actual_collection_line
+          @extra_items, @missing_items = fast_calculate_extra_missing if @transitive
           message += missing_elements_line unless missing_items.empty?
           message += extra_elements_line unless extra_items.empty?
           message
@@ -72,7 +86,9 @@ module RSpec
 
         def match(_expected, _actual)
           return false unless convert_actual_to_an_array
-          match_when_sorted? || (extra_items.empty? && missing_items.empty?)
+          matched_when_sorted = match_when_sorted?
+          return matched_when_sorted if matched_when_sorted || @transitive
+          (extra_items.empty? && missing_items.empty?)
         end
 
         # This cannot always work (e.g. when dealing with unsortable items,
@@ -80,7 +96,8 @@ module RSpec
         # the slowness of the full matching algorithm, and in common cases this
         # works, so it's worth a try.
         def match_when_sorted?
-          values_match?(safe_sort(expected), safe_sort(actual))
+          @sorted_expected, @sorted_actual = safe_sort(expected), safe_sort(actual)
+          values_match?(@sorted_expected, @sorted_actual)
         end
 
         def convert_actual_to_an_array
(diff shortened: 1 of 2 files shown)


bclayman-sq:
This is a small proof of concept to suggest a path forward for improving the performance of `ContainExactly` (and by extension, `MatchArray`) dramatically for common use cases.  I'd love to hear what folks think!

bclayman-sq:
@pirj Had to update a few things to pass CI for all ruby versions.  If you think it's reasonable, I can try to put out a more comprehensive PR.    

What do you think of this approach?

lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -72,7 +85,10 @@ def message_line(prefix, collection, surface_descriptions=false)
 
         def match(_expected, _actual)
           return false unless convert_actual_to_an_array
-          match_when_sorted? || (extra_items.empty? && missing_items.empty?)
+          matched_when_sorted = match_when_sorted?
+          return true if matched_when_sorted
+          return matched_when_sorted if @transitive
```
pirj, line 90: I'm struggling to understand how this works.
From my understanding, it would clamp here, and would return `false` early before calculating `extra_items` and `missing_items` if `@transitive` is `true`. Is this how it works?

lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -6,6 +6,19 @@ module BuiltIn
...
       class ContainExactly < BaseMatcher
+        def initialize(expected=nil)
+          super
+          @transitive = false
+        end
+
+        # @api public
+        # Specifies that elements contained in actual and expected
+        # obey transitivity.  This lets match run much faster.
+        def transitive
```
pirj, line 17: Do you think it would be possible to infer this from what `expected` is? E.g. if there's a matcher there or a Regex, they won't be transitive. And imply transitivity otherwise?

lib/rspec/matchers/built_in/contain_exactly.rb
pirj: However, `extra_items` and `missing_items` will be called later when `failure_message` is called, wouldn't they?

bclayman-sq:
@pirj 

I'm gonna reply in-line just to keep a summary of where we're at in one place:

>It would be ideal, though, not to expose transitive, as, frankly, I had to look it up to remind myself what it was.
And still, users (I do not assume stupidity, just human error) might use it wrong. Apart from it being yet another to keep in mind.

I think there are a few ways to address your concerns above:

**1. Better naming:** I'm not wedded to transitive. We could call it sortable, comparable, etc.
**2. Good documentation:** Explicit and detailed documentation will hopefully help users use it correctly.
**3. Informative error messages when someone does misuse transitive:** For example, if a user uses transitive with elements that aren't comparable, we can raise. That can happen [here](https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers/built_in/contain_exactly.rb#L99):
```
def safe_sort(array)
  array.sort
rescue Support::AllExceptionsExceptOnesWeMustNotRescue
  raise "Invalid use of `.transitive` with unsortable array #{array}" if @transitive
  array
end
```
>> Maybe we could calculate the missing and extra items separately

> This sounds like a reasonable approach. It feels that for transitive we can improve pairing maximizer to be O(n) (on two sorted arrays, since sorting that was performed previously already).

Awesome, I think we're on the same page (and thanks for being so helpful in your responses).

I've just refactored my code to calculate extra and missing items in O(n) time when transitive is used. This addresses your earlier concern about `generate_failure_message` incurring O(n!) work by relying on `PairingsMaximizer` methods.

I think it'd be a really nice improvement to `ContainExactly` from O(n!) to O(n log n) when sorting doesn't result in a match and the elements are transitive.  I'd be open to improving naming, updating documentation, and adding error-handling if that would get this across the finish line.

When you get a chance (you've already helped tons, thank you!), would you mind letting me know if you'd be open to merging this functionality into RSpec?

lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -124,6 +142,47 @@ def extra_items
           end
         end
 
+        # We use this to determine extra and missing items between expected
+        # and actual arrays.  This runs in O(n) time which is a big improvement
+        # over the O(n!) work incurred by PairingsMaximizer to evaluate all possible
+        # matchings between arrays
+        # rubocop:disable MethodLength
+        # rubocop:disable Metrics/AbcSize
+        def fast_calculate_extra_missing
```
pirj, line 151: There must be a mistake somewhere in this method:
```ruby
@expected # => [1, 2, 3]
@actual # => [1, 2]
fast_calculate_extra_missing # => [[], [2]]
```
Should be `[[], [3]]`.

spec/rspec/matchers/built_in/contain_exactly_spec.rb
```diff
@@ -95,6 +95,77 @@ def array.send; :sent; end
 end
 
 RSpec.describe "using contain_exactly with expect" do
+  # users have reported using contains_exactly with 50 elements
+  # never finishing!
+  context "with transitive enabled" do
+    require 'benchmark'
+
+    context "with actual and expected containing sortable elements" do
+      shared_examples "runs very fast" do
```
pirj, line 104: Just found a `timeout_if_not_debugging(0.1)` helper down below in this spec.

lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -162,11 +162,11 @@ def fast_calculate_extra_missing
           end
 
           while i < @sorted_actual.size
-            extra << current_actual
+            extra << @sorted_actual[i]
```
bclayman-sq, line 165: Ah nice catch!

pirj:
I'm all good with the change. The only thing I think needs addressing are specs - there is a very similar example with 10 numbers and Timeout check. It makes sense to combine or at least put those two closer to each ***@***.*** @benoitgt wdyt?                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: bclayman-sqSent: Sunday, 10 October 2021 17:16To: rspec/rspec-expectationsReply To: rspec/rspec-expectationsCc: Phil Pirozhkov; MentionSubject: Re: [rspec/rspec-expectations] Proof of concept for improving ContainExactly matcher speed when elements obey transitivity (#1325)

I dared to push some cosmetic changes, what do you think of such an auto-detection of transitivity?

@pirj,
I think this is great and a clear improvement over .transitive; users will get the benefit without having to know about or remember anything about .transitive!
If you're happy with this, is it good to merge?

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.

[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/rspec/rspec-expectations/pull/1325#issuecomment-939490066",
"url": "https://github.com/rspec/rspec-expectations/pull/1325#issuecomment-939490066",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

bclayman-sq:
> I'm all good with the change. The only thing I think needs addressing are specs - there is a very similar example with 10 numbers and Timeout check. It makes sense to combine or at least put those two closer to each

@pirj Yeah, that makes a lot of sense.  I just refactored the tests so that my speed tests are near the 10 number + Timeout check.  Cleaned a few things up with shared examples too.

How does it look?

bclayman-sq:
Closing in favor of https://github.com/rspec/rspec-expectations/pull/1333

Issue #1287, opened 2021-02-03
Incorrect output with a reused argument matcher


pirj:
### Subject of the issue

```ruby
RSpec.describe 'A', :aggregate_failures do
  specify do
    matcher = contain_exactly(eq(1))

    expect([2]).to matcher
    expect([3]).to matcher
  end
end
```

The second failure message seems to have traces from the first evaluation:
```
            the extra elements were:        [2]
```

It appears that `with_matchers_cloned` (`lib/rspec/matchers/composable.rb`) doesn't cover this case.

Adding `.clone` or `.dup` to `.to matcher` solves the issue.

### Your environment

* Ruby version: 2.7.1
* RSpec 3.10
  - rspec-core 3.10.1
  - rspec-expectations 3.10.1

### Expected behavior

```
  1) A is expected to contain exactly (eq 1)
     Got 2 failures:

     1.1) Failure/Error: expect([2]).to matcher.dup

            expected collection contained:  [(eq 1)]
            actual collection contained:    [2]
            the missing elements were:      [(eq 1)]
            the extra elements were:        [2]
          # ./spec/a_spec.rb:5:in `block (2 levels) in <top (required)>'

     1.2) Failure/Error: expect([3]).to matcher.dup

            expected collection contained:  [(eq 1)]
            actual collection contained:    [3]
            the missing elements were:      [(eq 1)]
            the extra elements were:        [3]
          # ./spec/a_spec.rb:6:in `block (2 levels) in <top (required)>'
```

### Actual behavior

```
  1) A is expected to contain exactly (eq 1)
     Got 2 failures:

     1.1) Failure/Error: expect([2]).to matcher

            expected collection contained:  [(eq 1)]
            actual collection contained:    [2]
            the missing elements were:      [(eq 1)]
            the extra elements were:        [2]
          # ./spec/a_spec.rb:5:in `block (2 levels) in <top (required)>'

     1.2) Failure/Error: expect([3]).to matcher

            expected collection contained:  [(eq 1)]
            actual collection contained:    [3]
            the missing elements were:      [(eq 1)]
            the extra elements were:        [2]
          # ./spec/a_spec.rb:6:in `block (2 levels) in <top (required)>'
```

JonRowe:
Some output must be cached between runs, we are supposed to be allowing this use case.

pirj:
Fixed in #1326

PR #1326, opened 2021-10-12
Allow matchers to be reused

bclayman-sq:
This addresses issue #1287.

Previously, matchers retained state after being used as an argument to `RSpec::Expectations::ExpectationTarget#to`.  In particular, it leaked `@extra_items`, `@missing_items`, `@best_solution`, and `@pairings_maximizer`. This caused incorrect results when reusing a matcher in subsequent expectations.

This PR addresses the issue by reinitializing the matcher's state when the matcher is reused.

Co-authored by: Rai Feren <rferen@squareup.com>

### DIFF
--- a/lib/rspec/matchers/built_in/contain_exactly.rb
+++ b/lib/rspec/matchers/built_in/contain_exactly.rb
@@ -31,6 +31,14 @@ module RSpec
           "contain exactly#{list}"
         end
 
+        def matches?(actual)
+          @pairings_maximizer = nil
+          @best_solution = nil
+          @extra_items = nil
+          @missing_items = nil
+          super(actual)
+        end
+
       private
 
         def generate_failure_message
--- a/spec/rspec/matchers/built_in/contain_exactly_spec.rb
+++ b/spec/rspec/matchers/built_in/contain_exactly_spec.rb
@@ -473,6 +473,32 @@ RSpec.describe "Composing `contain_exactly` with other matchers" do
   end
 end
 
+RSpec.describe "Reusing a matcher that memoizes state" do
+  require "rspec/matchers/fail_matchers"
+
+  it "works properly in spite of the memoization" do
+    matcher = contain_exactly(eq(1))
+
+    expect {
+      expect([2]).to matcher
+    }.to fail_including(<<-MESSAGE)
+expected collection contained:  [(eq 1)]
+actual collection contained:    [2]
+the missing elements were:      [(eq 1)]
+the extra elements were:        [2]
+    MESSAGE
+
+    expect {
+      expect([3]).to matcher
+    }.to fail_including(<<-MESSAGE)
+expected collection contained:  [(eq 1)]
+actual collection contained:    [3]
+the missing elements were:      [(eq 1)]
+the extra elements were:        [3]
+    MESSAGE
+  end
+end
+
 module RSpec
   module Matchers
     module BuiltIn


lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -112,6 +115,14 @@ def matcher_name
           end
         end
 
+        # @private
+        def reinitialize
```
pirj, line 119: It is possible to push this method below `private` waterline.

I'd suggest naming it `reset`.

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -34,6 +34,9 @@ def initialize(expected=UNDEFINED)
         # should be defined on a subclass. Takes care of consistently
         # initializing the `actual` attribute.
         def matches?(actual)
+          # if @actual is already set, we know this matcher was used previously and want
+          # to reset memoized fields
+          reinitialize if !!defined?(@actual)
```
pirj, line 39: Why is `!!` needed?

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -112,6 +115,14 @@ def matcher_name
           end
         end
 
+        # @private
+        def reinitialize
+          @pairings_maximizer = nil
+          @best_solution = nil
+          @extra_items = nil
+          @missing_items = nil
```
pirj, line 123: Those are ones from `contain_exactly`? Does it make sense to override this method there, and keeping it empty here?

spec/rspec/matchers/composable_spec.rb
```diff
@@ -49,6 +49,32 @@ module Matchers
         end
       end
 
+      context "when reusing a matcher that memoizes state" do
```
pirj, line 52: Since you've fixed `contain_exactly`, I believe this example would better go to its spec.

lib/rspec/matchers/built_in/base_matcher.rb
bclayman-sq: Yeah, that makes a lot more sense.  I'll pull this particular implementation into `contain_exactly` and leave it empty in the parent class 👍

pirj (approved):
Perfect, thank you!

pirj:
> why ruby 2.2 on windows fails in CI

It segfaulted last time, passed this time, and crashed with no message the first time. I don't think it is caused by your changes or can be fixed. It is not a blocker to merge this PR.

bclayman-sq:
@pirj @JonRowe 👋 Anything I can do to get this PR merged?

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -34,6 +34,9 @@ def initialize(expected=UNDEFINED)
         # should be defined on a subclass. Takes care of consistently
         # initializing the `actual` attribute.
         def matches?(actual)
+          # if @actual is already set, we know this matcher was used previously and want
+          # to reset memoized fields
+          reset if defined?(@actual)
```
JonRowe, line 39:
```suggestion
```

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -126,6 +129,8 @@ def self.underscore(camel_cased_word)
 
       private
 
+        def reset; end
+
```
JonRowe, line 133:
```suggestion
```

lib/rspec/matchers/built_in/contain_exactly.rb
```diff
@@ -33,6 +33,13 @@ def description
 
       private
 
+        def reset
+          @pairings_maximizer = nil
+          @best_solution = nil
+          @extra_items = nil
+          @missing_items = nil
+        end
```
JonRowe, line 41:
```suggestion
        def matches?(actual)
          @pairings_maximizer = nil
          @best_solution = nil
          @extra_items = nil
          @missing_items = nil
          super(actual)
        end
```

JonRowe (changes requested):
This seems like a bug with only the `RSpec::Matchers::BuiltIn::ContainExactly` matcher, how about just overriding `matches?(actual)` rather than adding a method to every matcher?

lib/rspec/matchers/built_in/contain_exactly.rb
bclayman-sq: @JonRowe Yep, this is even nicer than my approach: it keeps the change local to `ContainExactly`.  Just incorporated.  Mind kicking off a build?

bclayman-sq: One more build and should be good to go!

pirj:
Thank you!

Issue #1025, opened 2017-11-20
`end_with` takes the assumption that the receiver is a hash, or string.


envygeeks:
```ruby
[14] pry(<REMOVED>)> rtn.class # => Pathutil
[15] pry(<REMOVED>)> rtn.end_with?("/hello") # => true
[16] pry(<REMOVED>)> expect(rtn).to end_with("/hello")

RSpec::Expectations::ExpectationNotMetError
# => : expected #<Pathutil:/hello> to end with "/hello", but it cannot 
#   be indexed using #[]
```

myronmarston:
`start_with` and `end_with` don't assume the receiver is a hash; they assume the receiver is an object that implements the `[]` indexing operator so that the matcher can directly access the elements at the start (using `[0]`, `[0, 5]`, etc for `start_with`) or at the end (using `[-1]`, `[-5, 5]`, etc for `end_with`).  In fact, the `start_with` and `end_with` matchers don't work on hashes, as there internal ordering, where it exists (e.g. MRI 1.9+), is not exposed by the `[]` operator.

Anyhow, if you want it to work with an object of your own, you can implement `[]` on it, and the matcher should "just work".  `start_with` and `end_with` have never checked a predicate like `start_with?` or `end_with?`.  I think it would be reasonable to do so.  Do you want to open a PR adding that?

TravisSpangle:
Something like ?
```
        # @api private
        # @return [String]
        def failure_message
          super.tap do |msg|
            if @actual_does_not_have_ordered_elements
              msg << ", but it does not have ordered elements"
            elsif !actual.respond_to?(:start_with?)
              msg << ", but #{actual.inspect} does not respond to start_with?"
            elsif !actual.respond_to?(:end_with?)
              msg << ", but #{actual.inspect} does not respond to end_with?"
            elsif !actual.respond_to?(:[])
              msg << ", but it cannot be indexed using #[]"
            end
          end
        end
```

myronmarston:
@TravisSpangle - the fix for this is going to require a change the `match` logic (so that if the object responds to the appropriate predicate method, it is used), as well as a change to `failure_message` so that when the predicate approach was used, the failure message mentions it.

(And sorry about the delayed response...)

JonRowe:
Closing due to inactivity during the monorepo migration, and I feel the original access modifier reasoning still stands

PR #1327, opened 2021-10-13
Update StartWith, EndWith to use start_with?, end_with? if available

bclayman-sq:
This addresses issue #1025.

With this change, the `StartWith` matcher will rely on an object's `start_with?` method if available.  Similarly, the `EndWith` matcher will rely on an object's `end_with?` method if available.

This is especially useful when a class implements `start_with?` but not the indexing operator, or `end_with?` but not the indexing operator.

### DIFF
--- a/lib/rspec/matchers/built_in/start_or_end_with.rb
+++ b/lib/rspec/matchers/built_in/start_or_end_with.rb
@@ -13,11 +13,12 @@ module RSpec
         # @api private
         # @return [String]
         def failure_message
+          response_msg = ", but it does not respond to #{method} and cannot be indexed using #[]"
           super.tap do |msg|
             if @actual_does_not_have_ordered_elements
               msg << ", but it does not have ordered elements"
             elsif !actual.respond_to?(:[])
-              msg << ", but it cannot be indexed using #[]"
+              msg << response_msg
             end
           end
         end
@@ -33,7 +34,9 @@ module RSpec
 
       private
 
-        def match(_expected, actual)
+        def match(expected, actual)
+          # use an object's start_with? or end_with? as appropriate
+          return actual.send(method, expected) if actual.respond_to?(method)
           return false unless actual.respond_to?(:[])
 
           begin
@@ -73,6 +76,10 @@ module RSpec
         def element_matches?
           values_match?(expected, actual[0])
         end
+
+        def method
+          :start_with?
+        end
       end
 
       # @api private
@@ -87,6 +94,10 @@ module RSpec
 
         def element_matches?
           values_match?(expected, actual[-1])
+        end
+
+        def method
+          :end_with?
         end
       end
     end
--- a/spec/rspec/matchers/built_in/start_and_end_with_spec.rb
+++ b/spec/rspec/matchers/built_in/start_and_end_with_spec.rb
@@ -102,11 +102,25 @@ RSpec.describe "expect(...).to start_with" do
   end
 
   context "with an object that does not respond to :[]" do
-    it "fails with a useful message" do
-      actual = Object.new
-      expect {
-        expect(actual).to start_with 0
-      }.to fail_with("expected #{actual.inspect} to start with 0, but it cannot be indexed using #[]")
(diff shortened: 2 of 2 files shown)


pirj (commented):
Looks great!
Wondering why it wasn't done like that from the very beginning.

Wondering if we need to check if there are multiple args:
```ruby
"abc".start_with?('a', 'c') # => true
```
https://ruby-doc.org/core-2.7.1/String.html#method-i-start_with-3F seems to allow all prefixes, while the matcher's semantic is different.

spec/rspec/matchers/built_in/start_and_end_with_spec.rb
```diff
@@ -102,11 +102,25 @@ def ==(other)
...
+      end
+    end
+
+    context "with an object that does not respond to start_with?" do
+      it "fails with a useful message" do
+        actual = Object.new
+        expect {
+          expect(actual).to start_with 0
+        }.to fail_with("expected #{actual.inspect} to start with 0, but it does not respond to start_with? and cannot be indexed using #[]")
+      end
```
JonRowe, line 123: Rather than removing the old spec and nesting additional contexts, these should be seperate cases with the original left as is

spec/rspec/matchers/built_in/start_and_end_with_spec.rb
```diff
@@ -310,11 +324,24 @@ def ==(other)
...
+    end
+
+    context "with an object that does not respond to end_with?" do
+      it "fails with a useful message" do
+        actual = Object.new
+        expect {
+          expect(actual).to end_with 0
+        }.to fail_with("expected #{actual.inspect} to end with 0, but it does not respond to end_with? and cannot be indexed using #[]")
+      end
```
JonRowe, line 344: Same as above

JonRowe (changes requested):
Nice work, but this can't be merged into main as its a breaking change in behaviour, so would have to be for rspec 4.

lib/rspec/matchers/built_in/start_or_end_with.rb
```diff
@@ -13,11 +13,12 @@ def initialize(*expected)
         # @api private
         # @return [String]
         def failure_message
+          response_msg = ", but it does not respond to #{method} and cannot be indexed using #[]"
           super.tap do |msg|
             if @actual_does_not_have_ordered_elements
               msg << ", but it does not have ordered elements"
             elsif !actual.respond_to?(:[])
-              msg << ", but it cannot be indexed using #[]"
+              msg << response_msg
```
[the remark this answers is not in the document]
bclayman-sq: Oh yeah, I botched something here 😅 .  I meant it to be:

```
response_msg = ", but it does not respond to #{method} and cannot be indexed using #[]"
super.tap do |msg|
  if @actual_does_not_have_ordered_elements
    msg << ", but it does not have ordered elements"
  elsif !actual.respond_to?(method) && !actual.respond_to?(:[])
    msg << response_msg
  end
```
Here's what I'm thinking: Suppose `actual` does have ordered elements.  We should only fail if `actual` responds to neither `start_with?` nor indexing.  If it responds to either, we don't need to append to our failure message.  

Does that sound reasonable?

spec/rspec/matchers/built_in/start_and_end_with_spec.rb
bclayman-sq: Oh I'd love to hear more about the rationale for this!  I often set up my tests to be mutually exclusive and nesting makes that easier for me.  So they often have this shape:

```
when A
  when B
  when not B
when not A
  when B
  when not B
```
I wasn't totally sure about leaving this case as is because now, simply knowing that `actual` responds to `:[]` doesn't fully specify what should happen.  If `actual` doesn't respond to `start_with?`, it should have one behavior. If it does, it should have a different behavior.

Could you tell me more about how you were thinking of structuring this given the above?

JonRowe:
Migrated to the monorepo.

PR #1330, opened 2021-10-27
Present the actual error message from raise_error

### LINKED ISSUE #137: Allow configuring exceptions raise_error should capture to display webmock errors correctly
I'm doing: `expect { call }.to raise_error(/ABORTED/)`

this triggers a http request and then it ends up as:
```
       expected /ABORTED/, got #<WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET h...e.sk/projects/sfn_kubernetes/stages")
```

but what I want to see is:
```
     WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: GET https://samson.zende.sk/locks with body '{"per_page":100,"page":1}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer a', 'Content-Type'=>'application/json', 'User-Agent'=>'Ruby'}

       You can stub this request with the following snippet:
blah blah
```

... this does not surface though since `raise_error` captures `Exception` and not `StandardError` like `minitest` does.
I assume changing the default would be too disruptive, so I'd suggest we make that configureable ?

workaround I'm using atm:
```ruby
  undef :raise_error
  undef :raise_exception
  def expect_error(regex)
    yield
  rescue StandardError => e
    raise unless e.message.match?(regex)
  else
    raise "Nothing raised"
  end
```

### LINKED ISSUE #1190: Allow configuring exceptions raise_error should capture to display webmock errors correctly
I'm doing: `expect { call }.to raise_error(/ABORTED/)`

this triggers a http request and then it ends up as:
```
       expected /ABORTED/, got #<WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET h...e.sk/projects/sfn_kubernetes/stages")
```

but what I want to see is:
```
     WebMock::NetConnectNotAllowedError:
       Real HTTP connections are disabled. Unregistered request: GET https://samson.zende.sk/locks with body '{"per_page":100,"page":1}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer a', 'Content-Type'=>'application/json', 'User-Agent'=>'Ruby'}

       You can stub this request with the following snippet:
blah blah
```

... this does not surface though since `raise_error` captures `Exception` and not `StandardError` like `minitest` does.
I assume changing the default would be too disruptive, so I'd suggest we make that configureable ?

genehsu:
Fixes rspec/rspec#137 

Display the actual error message of the raised error when a raise_error
assertion fails.  The failure message will now include the error message
on a separate line followed by the backtrace.  If there is no error
message, only the backtrace will be shown.

If this approach is not correct, then please close this PR without prejudice 😃

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -238,10 +238,10 @@ module RSpec
           return " but nothing was raised" unless @actual_error
 
           backtrace = format_backtrace(@actual_error.backtrace)
-          [
-            ", got #{description_of(@actual_error)} with backtrace:",
-            *backtrace
-          ].join("\n  # ")
+
+          ", got #{description_of(@actual_error)}" +
+            error_message_text +
+            text_with_indent(" with backtrace:", backtrace)
         end
 
         def expecting_specific_exception?
@@ -256,6 +256,25 @@ module RSpec
         def warning
           warning = "Actual error raised was #{description_of(@actual_error)}. "
           warning if @actual_error
+        end
+
+        private
+
+        def error_message_text
+          return '' if @actual_error.message.nil? || @actual_error.message.empty?
+          return '' if @actual_error.message == @actual_error.class.to_s
+
+          text_with_indent(" with message:", @actual_error.message.split(/\n/)) +
+            "\nand"
+        end
+
+        INDENT = ["\n  # ".freeze].freeze
+
+        def text_with_indent(prefix, text_lines)
+          text_lines = Array(text_lines)
+          message_text = [prefix]
+          message_text += ([INDENT] * text_lines.size).zip(text_lines)
+          message_text.join
         end
       end
       # rubocop:enable RescueException
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -105,11 +105,35 @@ RSpec.describe "expect { ... }.to raise_error" do
   end
 
   it "fails if a different error instance is thrown from the one that is expected" do
-    s = StandardError.new("Error 1")
+    error = StandardError.new("Error 1")
     to_raise = StandardError.new("Error 2")
     expect do
-      expect { raise to_raise }.to raise_error(s)
-    end.to fail_with(Regexp.new("expected #{s.inspect}, got #{to_raise.inspect} with backtrace"))
+      expect { raise to_raise }.to raise_error(error)
+    end.to fail_with(Regexp.new("expected #{error.inspect}, got #{to_raise.inspect} with message:\n  # Error 2\nand with backtrace:"))
+  end
+
+  it "fails without a message if a different error instance is thrown from the one that is expected" do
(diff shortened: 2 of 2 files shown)


lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -238,10 +238,18 @@ def given_error
...
+            message_text << "\nand"
+          end
           [
-            ", got #{description_of(@actual_error)} with backtrace:",
-            *backtrace
-          ].join("\n  # ")
+            ", got #{description_of(@actual_error)}",
+            message_text,
+            [" with backtrace:", *backtrace].zip(["\n  # "] * backtrace.size)
+          ].flatten.join('')
```
JonRowe, line 252: This array could probably just be interpolation now, as the intent of the original map is taken care of by the zip.

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -109,7 +109,31 @@ def to_s
     to_raise = StandardError.new("Error 2")
     expect do
       expect { raise to_raise }.to raise_error(s)
-    end.to fail_with(Regexp.new("expected #{s.inspect}, got #{to_raise.inspect} with backtrace"))
+    end.to fail_with(Regexp.new("expected #{s.inspect}, got #{to_raise.inspect} with message:\n  # Error 2\nand with backtrace:"))
+  end
+
+  it "fails without a message if a different error instance is thrown from the one that is expected" do
+    s = StandardError.new("Error 1")
```
JonRowe, line 116: Rather than `s` can you use a proper name like `error`

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -238,10 +238,18 @@ def given_error
           return " but nothing was raised" unless @actual_error
 
           backtrace = format_backtrace(@actual_error.backtrace)
+          if @actual_error.message.nil? || @actual_error.message.empty? || (@actual_error.message == @actual_error.class.to_s)
+            message_text = ''
+          else
+            message_lines = @actual_error.message.split(/\n/)
+            message_text = [" with message:", *message_lines].zip(["\n  # "] * message_lines.size)
```
JonRowe, line 245: given that the magic string `"\n  # "` is now used multiple times it might be best to assign it and reuse it

JonRowe (changes requested):
Thanks for tackling this, can you check out these couple of suggestions?

genehsu:
@JonRowe 
> Thanks for tackling this, can you check out these couple of suggestions?

Thanks for the review. I believe I've addressed your suggestions with clearer code. Let me know if you have any other suggestions to the code or output.

JonRowe:
Migrated to the monorepo as rspec/rspec#128

PR #1332, opened 2021-10-29
Add and_return to change matcher

dsounded:
This request add `and_return` check for `change` matcher (like when we do `expect().to receive().and_return`)

It's sometimes useful when your block mutates some state/prop/variable but return some other value (you may check `SomethingExpectedWithResult` as example but for the real world scenario it can be service which inserts some rows into the DB and returns `true` if it is successful or `false` if some error happened  respectively.

eg:
`expect{ insert_articles }.to change { Article.count }.by(20).and_return(true)`
`expect{ insert_articles_returning_ids }.to change { Article.count }.by(3).and_return([1, 2, 3])`

I thought about something like this more than 4 years ago but now decided to create a PR :)

Right now people do some workarounds like described here:
https://stackoverflow.com/questions/49256551/can-you-simultaneously-test-a-state-change-and-a-return-value-in-rspec

Which is also okay, but RSpec DSL is really cool and this `.and_return` way of describing things is nice to have.

Because right now I do something like:
```
return_value = nil
expect{ return_value = wojit.do_thing }.to change{ wojit.value }
expect( return_value ).to be true
```

So with this change it becomes an one-liner:
`expect(  wojit.do_thing ).to change{ wojit.value }.and_return(true)`

By the way, thanks for this great gem, guys!

### DIFF
--- a/lib/rspec/matchers/built_in/change.rb
+++ b/lib/rspec/matchers/built_in/change.rb
@@ -6,6 +6,15 @@ module RSpec
       # Not intended to be instantiated directly.
       class Change < BaseMatcher
         # @api public
+        # Specifies the returning block's value.
+        def and_return(expected_value)
+          @expected_return = expected_value
+          @return_matcher = Eq.new(expected_value)
+
+          self
+        end
+
+        # @api public
         # Specifies the delta of the expected change.
         def by(expected_delta)
           ChangeRelatively.new(change_details, expected_delta, :by) do |actual_delta|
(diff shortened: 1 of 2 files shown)


dsounded:
Yes, this also can work but it's something similar to other workarounds (and even though it can be one-liner you need to write more boilerplate, and it is also nested) + different error message:

```
expected: false
got: true
```

This version more like DSL-ish (or human language) though I understand it's pretty subjective statement.

JonRowe:
`and_return` is really the wrong language, but I tried to implement this as a compound matcher, which I would have thought would work:
```
expect { ... }.to change { ... }.and eq :value
```
But it doesn't support block expectations, I'd rather change that then add this so this would become:
```
expect{ insert_articles }.to change { Article.count }.by(20).and eq true
```
Which is grammatically more in keeping with our semantics

pirj:
👍 for `change { Article.count }.by(20).and eq(true)`.

Something to keep in mind, though.
```ruby
expect { some_action }
  .to change { Acticle.count }.by(1).and eq(Article.last)
```
won't work as expected because `Article.last` will be executed before `some_action`, and on a clean slate `Article.last` would return `nil`.
That has a risk of false negatives in the case when `some_action` unexpectedly returns `nil`, and `Acticle.last`, too.
`change { ... }.by(1)` is balancing it, but not in all cases.

I'd rather recommend using [`satisfy`](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/satisfy-matcher) instead of `eq` in such expectations. If I'm not mistaken, its block would evaluate _after_ `some_action`.

PR #1333, opened 2021-10-29
Speed up the ContainExactly matcher

bclayman-sq:
Speed up the ContainExactly matcher by pre-emptively matching up corresponding elements in the expected and actual arrays. 

This addresses issues rspec/rspec#84, rspec/rspec#136.

**Before this PR, comparing two arrays of 50 random integers never finished.  Now, comparing arrays of 10,000 random integers takes 1 second.**

This PR is a collaboration between @genehsu and I based on a couple of our earlier PRs and discussion that resulted:
1) https://github.com/rspec/rspec-expectations/pull/1325
2) https://github.com/rspec/rspec-expectations/pull/1328

Co-authored-by: Gene Hsu <gene@hsufarm.com>

PR #1334, opened 2021-11-18
Cache match results for compound failures

### LINKED ISSUE #87: Slow sequential And
### Subject of the issue

It appears that if And compounds are sequential then tests are running for a long time. Eg.:

```ruby
expect('qwe').
  to include('asd').
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd')).
    and(include('asd'))
```
this is running on my computer for 15 seconds, and each additional condition doubles the previous duration.

It's not a problem with nested and:

```ruby
expect('qwe').
  to include('asd').
    and include('asd').
      and include('asd').
        and include('asd').
          and include('asd').
            and include('asd').
              and include('asd').
                and include('asd').
                  and include('asd').
                    and include('asd').
                      and include('asd').
                        and include('asd').
                          and include('asd').
                            and include('asd')
```
This is instant. But I prefer the previous sequential approach as it doesn't increase nesting level.

### Your environment
* Ruby version: 2.6.3
* rspec-expectations version: 3.10.1

genehsu:
fixes rspec/rspec#87 

When a compound matcher fails, the failure message checks both sides of
the match to determine how to display the failure match.  When multiple
compound matchers are chained, that means the match for each side of the
compound matcher may be called repeatedly for exponential growth.

To address this behavior, we can cache the match result for each side of
the compound operator each time a match is executed.  We need to reset the
cache each time the match is executed because the matcher may be re-used
for a different expectation.

Added RSpec context timing before the patch:
```
RSpec::Matchers::BuiltIn::Compound
  when chaining many matchers together
    with long chains of compound matchers
      with a failing first matcher
        generates a failure description quickly
      with all failing matchers
        generates a failure description quickly with or
        generates a failure description quickly with and
      with a failing last matcher
        generates a failure description quickly

Finished in 2 minutes 47.7 seconds (files took 0.43342 seconds to load)
4 examples, 0 failures
```

Added RSpec context timing after the patch:
```
RSpec::Matchers::BuiltIn::Compound
  when chaining many matchers together
    with long chains of compound matchers
      with a failing last matcher
        generates a failure description quickly
      with all failing matchers
        generates a failure description quickly with and
        generates a failure description quickly with or
      with a failing first matcher
        generates a failure description quickly

Finished in 0.05168 seconds (files took 0.42805 seconds to load)
4 examples, 0 failures
```

### DIFF
--- a/lib/rspec/matchers/built_in/compound.rb
+++ b/lib/rspec/matchers/built_in/compound.rb
@@ -75,6 +75,8 @@ module RSpec
         end
 
         def match(_expected, actual)
+          @matcher_1_matches = nil
+          @matcher_2_matches = nil
           evaluator_klass = if supports_block_expectations? && Proc === actual
                               NestedEvaluator
                             else
@@ -97,11 +99,13 @@ module RSpec
         end
 
         def matcher_1_matches?
-          evaluator.matcher_matches?(matcher_1)
+          return @matcher_1_matches unless @matcher_1_matches.nil?
+          @matcher_1_matches = evaluator.matcher_matches?(matcher_1)
         end
 
         def matcher_2_matches?
-          evaluator.matcher_matches?(matcher_2)
+          return @matcher_2_matches unless @matcher_2_matches.nil?
+          @matcher_2_matches = evaluator.matcher_matches?(matcher_2)
         end
 
         def matcher_supports_block_expectations?(matcher)
--- a/spec/rspec/matchers/built_in/compound_spec.rb
+++ b/spec/rspec/matchers/built_in/compound_spec.rb
@@ -882,31 +882,39 @@ module RSpec::Matchers::BuiltIn
 
         context "with a failing first matcher" do
           it "generates a failure description quickly" do
-            compound = failing_matcher
-            15.times { compound = compound.and(passing_matcher) }
-            expect { expect([actual]).to compound }.to fail_including("expected [#{actual}] to include #{not_expected}")
+            timeout_if_not_debugging(0.2) do
+              compound = failing_matcher
+              15.times { compound = compound.and(passing_matcher) }
+              expect { expect([actual]).to compound }.to fail_including("expected [#{actual}] to include #{not_expected}")
+            end
           end
         end
 
         context "with a failing last matcher" do
           it "generates a failure description quickly" do
-            compound  = failing_matcher
-            15.times { compound = passing_matcher.and(compound) }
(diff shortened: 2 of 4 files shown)


pirj (approved):
Looks great!
Does it make sense add `timeout_if_not_debugging` in the same vein you've used in #1333? The difference in running those specs is incredible - from 200s to 0.1s.
Thank you.

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -97,11 +99,19 @@ def compound_failure_message
         end
 
         def matcher_1_matches?
-          evaluator.matcher_matches?(matcher_1)
+          if @matcher_1_matches.nil?
+            @matcher_1_matches = evaluator.matcher_matches?(matcher_1)
+          else
+            @matcher_1_matches
+          end
```
pirj, line 106: To appease RuboCop's `Metrics/ClassLength` it's possible to write this as a guard statement:
```ruby
return @matcher_1_matches unless @matcher_1_matches.nil?
@matcher_1_matches = evaluator.matcher_matches?(matcher_1)
```

genehsu:
@pirj  Style-wise, I was wondering about where the best place to reset state might be. I put the reset logic directly in the match method. I noticed that for a different issue, you add a blank `reset` method to `match?`. Should we incorporate that idea into this PR? or just concentrate on this PR.

All your suggestions seem good to me, so I'll go ahead and make them. Thanks for the look.

genehsu:
> Looks great! Does it make sense add `timeout_if_not_debugging` in the same vein you've used in #1333?

Where do helper functions go when they're used in multiple spec files? It seems that `spec/support` is all shared examples and matchers. There are a few helper functions in `spec_helper.rb`. That seems like where `timeout_if_not_debugging` should go, unless you have a different suggestion?

pirj:
> few helper functions in spec_helper.rb. That seems like where timeout_if_not_debugging should go

A perfect place for it 👍

genehsu:
This commit updates the timeout limit from 0.1s to 0.2s. Hopefully that's enough headroom for ruby 1.8.7.

pirj:
@Lukom does this fix your case?

JonRowe:
Migrated to the monorepo as rspec/rspec#129

PR #1345, opened 2022-01-11
Update Cucumber so that in runs across a range of Rubies

petergoldstein:
This PR allows more recent Rubies to use more recent versions of Cucumber, while still supporting older Cucumber versions.


JonRowe:
Thanks!

PR #1346, opened 2022-01-26
Fix a typo

luilver:


JonRowe:
Thats not a typo, its english rather than boolean algebra.

PR #1371, opened 2022-04-30
Show attribute details when composed have_attributes fails

nellirx:
As requested in rspec/rspec#131

For example
`expect([person]).to match [have_attributes(:age => 10)]`

Before
`expected [#{object_inspect person}] to match [(have attributes {:age => 10})]`

After
`expected [#{object_inspect person}] to match [(have attributes {:age => 10} but had attributes {:age => 33})]`


JonRowe (changes requested):
This is not the right approach to take, `formatted_values`is returning the wrong value for `failure_message` for the array case, thats what we need to fix, not changing the `description` that breaks the contract of our public matcher api.

JonRowe:
Closing due to lack of activity during the monorepo migration

PR #1373, opened 2022-05-15
Mention notify_expectation_failures in custom matcher docs

johnnymo87:
I defined a custom matcher that defined some `expect(...).to ...` expectations inside the `#match` block. I got one of these expectations to fail, and then was stumped as to why the file and line number of this failure didn't appear in the backtrace. Then I stumbled across the `notify_expectation_failures` option for `#match`, and realized that I needed to set it to `true` to get the failure to appear in the backtrace.

I think this would be good to mention in the relishapp documentation? It's currently mentioned in [the API documentation](https://rspec.info/documentation/3.9/rspec-expectations/RSpec/Matchers/DSL/Macros.html), which is good, but I feel this documentation a little more obscure.

### DIFF
--- a/features/custom_matchers/define_matcher.feature
+++ b/features/custom_matchers/define_matcher.feature
@@ -461,27 +461,29 @@ Feature: Define a custom matcher
 
     Given a file named "bubbling_expectation_errors_spec.rb" with:
       """ruby
-      RSpec::Matchers.define :be_close_to_pi do
+      RSpec::Matchers.define :be_a_palindrome do
         match(:notify_expectation_failures => true) do |actual|
-          expect(actual).to be_a(Float)
-          expect(actual).to be_within(0.0015).of(Math::PI)
+          expect(actual).to be_a(String)
+          expect(actual.reverse).to eq(actual)
         end
       end
 
       RSpec.describe "a custom matcher that bubbles up expectation errors" do
         it "bubbles expectation errors" do
-          expect(3.0).to be_close_to_pi
+          expect("carriage").to be_a_palindrome
         end
       end
       """
-    When I run `rspec ./bubbling_expectation_errors_spec.rb`
-    Then it should fail with:
-      """
-      Failures:
+    When I run `rspec bubbling_expectation_errors_spec.rb`
+    Then the output should contain all of these:
+      | Failures:                                                                           |
 
-        1) a custom matcher that bubbles up expectation errors bubbles expectation errors
-           Failure/Error: expect(actual).to be_within(0.0015).of(Math::PI)
-             expected 3.0 to be within 0.0015 of 3.141592653589793
-           # ./bubbling_expectation_errors_spec.rb:4:in `block (2 levels) in <top (required)>'
-           # ./bubbling_expectation_errors_spec.rb:10:in `block (2 levels) in <top (required)>'
-      """
+      |   1) a custom matcher that bubbles up expectation errors bubbles expectation errors |
+      |      Failure/Error: expect(actual.reverse).to eq(actual)                            |
+
+      |        expected: "carriage"                                                         |
+      |             got: "egairrac"                                                         |
+
+      |        (compared using ==)                                                          |
+      |      # ./bubbling_expectation_errors_spec.rb:4                                      |
+      |      # ./bubbling_expectation_errors_spec.rb:10                                     |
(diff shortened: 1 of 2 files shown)


features/custom_matchers/define_matcher.feature
```diff
@@ -454,3 +454,34 @@ Feature: Define a custom matcher
...
+      """
+    When I run `rspec ./bubbling_expectation_errors_spec.rb`
+    Then it should fail with:
+      """
+      Failures:
+
+        1) a custom matcher that bubbles up expectation errors bubbles expectation errors
+           Failure/Error: expect(actual).to be_within(0.0015).of(Math::PI)
+             expected 3.0 to be within 0.0015 of 3.141592653589793
+           # ./bubbling_expectation_errors_spec.rb:4:in `block (2 levels) in <top (required)>'
```
johnnymo87, line 485: If I don't set `notify_expectation_failures` to true, then this line doesn't appear in the backtrace.

features/custom_matchers/define_matcher.feature
```diff
@@ -454,3 +454,34 @@ Feature: Define a custom matcher
       """
     When I run `rspec ./alias_spec.rb --format documentation`
     Then the exit status should be 0
+
+  Scenario: With expectation errors that bubble up
+
+    By default the match block will swallow expectation errors (e.g. caused by using an expectation such as `expect(1).to eq 2`), if you with to allow these to bubble up, pass in the option `:notify_expectation_failures => true`.
```
pirj, line 460: Typo in "with". Also in [the source](https://github.com/rspec/rspec-expectations/blob/87588b1614a7be1c6560c5a024ccf8764fd125b9/lib/rspec/matchers/dsl.rb#L125) and [here](https://github.com/rspec/rspec-expectations/blob/87588b1614a7be1c6560c5a024ccf8764fd125b9/lib/rspec/matchers/dsl.rb#L154), too.

johnnymo87: Thank you, fixed in https://github.com/rspec/rspec-expectations/pull/1373/commits/ed6cf6804f577383993a6883da1a47e7192705a2.

pirj:
The feature failure with older Rubies appears to be due to a slightly different backtrace format.
See `features/aggregating_failures.feature` for an example on how to deal with this.
Alternatively, it is fine to just skip running this feature on older Rubies (1.8.7) (see `cucumber.yml` for an example of excluding 1.9).

pirj:
Installing 1.8.7 nowdays is challenging. I hope this helps https://github.com/rbenv/homebrew-tap

johnnymo87:
I've found that I have another problem if I get past the bundler version problem. The `gemspec` file has a `manifest` argument which is not compatible with the version of bundler that I can use with ruby 1.8.7 (bundler version 1.*). And if I comment out this repo's `manifest` block, it hits an error with the `manifest` block of `rspec-support`.

pirj:
> `gemspec` file has a `manifest` argument

Sorry, I can't find this `manifest`. Can you point it out?

For `rspec-support`, you can check out its repo next to `rspec-expectation`, and the [`Gemfile` will pick it up](https://github.com/rspec/rspec-expectations/blob/87588b1614a7be1c6560c5a024ccf8764fd125b9/Gemfile#L7).

johnnymo87:
Apologies, I misspoke, it was [metadata](https://github.com/rspec/rspec-expectations/blob/main/rspec-expectations.gemspec#L15-L21). In any case, I'm not sure **_what_** I was doing (24 hours ago) when I hit this problem. I found this very simple approach works:
1. Run the docker container as mentioned above.
2. `gem install bundler -v 1.17.3`
3. `mv Gemfile.lock Gemfile.lock.backup` (because it wants bundler version 2.3.8).
4. `script/legacy_setup.sh`
5. `bundle exec cucumber features/custom_matchers/define_matcher.feature:458`

And this works -- now I can reproduce the error from CI. So I'll keep working on this 🔨

pirj (approved):
Thank you, especially for being so persistent to push it through a cryptic 1.8.7 failure!

Out of curiosity, what was they problem of the previous approach of `be_close_to_pi` with 1.8.7?

johnnymo87:
@pirj Glad to have figured it out, thank you for the help!

Regarding `be_close_to_pi`, I noticed that the error message printed one less digit on ruby 1.8.7. And so I thought that even though the fuzzy matching of the `the output should contain all of these` step could handle this difference, I'd rather not burden future maintainers with this potentially tricky situation.

pirj:
Thank you for your contribution, @johnnymo87 !

PR #1374, opened 2022-06-01
Delegate kwargs when defining a method with `chain`

TylerRick:
Without this, it will raise:

     ArgumentError:
       wrong number of arguments (given 2, expected 1)

if you try to pass a keyword arg to a chain method on Ruby 3.

~In Ruby, you _must_ explicitly delegate keyword arguments. (See https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#delegation-ruby-3)~

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,10 @@
 ### Development
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.12.0...main)
 
+Bug Fixes:
+
+* Fix keyword arguments delegation for `chain` in custom matchers. (Tyler Rick, #1374)
+
 ### 3.12.0 / 2022-10-26
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.11.1...v3.12.0)
 
--- a/lib/rspec/matchers/dsl.rb
+++ b/lib/rspec/matchers/dsl.rb
@@ -306,6 +306,7 @@ module RSpec
             @chained_method_clauses.push([method_name, args])
             self
           end
+          ruby2_keywords method_name if respond_to?(:ruby2_keywords, true)
         end
 
         def assign_attributes(attr_names)
--- a/spec/rspec/matchers/dsl_spec.rb
+++ b/spec/rspec/matchers/dsl_spec.rb
@@ -354,9 +354,17 @@ module RSpec::Matchers::DSL
             (to_match <= five) || greater_than_ceiling(to_match) && not_divisible_by_divisor?(to_match)
           end
 
-          chain :and_smaller_than do |ceiling, inclusive: false|
-            @ceiling = ceiling
-            @ceiling_inclusive = inclusive
+          if RSpec::Support::RubyFeatures.kw_args_supported?
+            binding.eval(<<-CODE, __FILE__, __LINE__)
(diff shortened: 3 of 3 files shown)


pirj:
> if you try to pass a keyword arg to a chain method on Ruby 3.

As you may see, `**opts` delegation isn't really the best option. See https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html for more details.

A correct way to fix this is:
image

May I kindly ask you to separate the fix for this case from
```diff
- @divisor % to_match == 0
+ to_match % @divisor == 0
```
fixes? And potentially add an example that would indicate that the previous implementation was incorrect.

TylerRick:
> As you may see, `**opts` delegation isn't really the best option. See https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html for more details.
> 
> A correct way to fix this is: image

Good catch — thanks for the link! Pushed up a fixed commit.

Minor clarification: `**opts` delegation _is_ the best option if/when you only needs to support Ruby 3+ (i.e., as soon as we drop support for Ruby 2 someday :smile:), like I can do when developing _apps_. But in a library like this, yeah, you are right.
 
---

> May I kindly ask you to separate the fix for this case from
> 
> ```diff
> - @divisor % to_match == 0
> + to_match % @divisor == 0
> ```
> 
> fixes? And potentially add an example that would indicate that the previous implementation was incorrect.

I would be happy to, except that breaks my test.
The example I added here actually _does_ indicate that the previous implementation was incorrect. Without the fix, it fails with:

```
     Failure/Error: expect(10).to match
       expected 10 to be bigger than 5 and smaller than 10 and {:inclusive=>true} and divisible by 1
```
As you can see, 10 is divisible by 1.

I would rather just _leave off_ the `and_divisible_by(1)` part of this line altogether:
```
  match = matcher.and_smaller_than(10, inclusive: true).and_divisible_by(1)
```
since it's the test would be simpler without it, but when I tried that, it raised:

```ruby
     TypeError:
       nil can't be coerced into Integer
     # ./spec/rspec/matchers/dsl_spec.rb:381:in `%'
```

I considered doing further refactoring to make the `and_divisible_by` chain optional (as it should be), but figured that would cause a maintainer to complain about out-of-scope changes and that the bug-fix option I opted for instead would be just as trivial, and more appreciated.

Rather than wasting more time strategizing how best to incorporate this trivial test bug fix, may we please just include it here?

Or, if you'd prefer, would you mind just fixing the tests to however you'd like them? (I enabled the "Allow edits" option.)

pirj:
> just fix it now

Let's go for it :+1:

pirj:
```
 SyntaxError:
  /__w/rspec-expectations/rspec-expectations/spec/rspec/matchers/dsl_spec.rb:357: syntax error, unexpected tLABEL
            chain :and_smaller_than do |ceiling, inclusive: false|
```

I suggest extracting those specs with the matcher involving block kwargs into a separate one, as we still support Ruby < 2.
You can put it inside `binding.eval(<<-CODE, __FILE__, __LINE__)` as some other examples do.

JonRowe:
Migrated to the monorepo as rspec/rspec#130

PR #954, opened 2016-11-10
Prevent warning when exists matcher is used with files

JonRowe:
Fixes #947

PR #1383, opened 2022-09-11
Prevent deprecation warning from `expect(FileTest).to exist(x)`

r7kamura:
I found that `expect(FileTest).to exist(x)` also outputs deprecation warnings in the same way as `expect(File).to exist(x)` before https://github.com/rspec/rspec-expectations/pull/954.

```
FileTest.exists? is a deprecated name, use FileTest.exist? instead
```

I confirmed that this warning is displayed on Ruby 2.7.6 and Ruby 3.1.2.

So it would be nice that we also support `FileTest` as well as `File`. It won't be used that much, but it's not a big change and it would be better to support it.

### DIFF
--- a/lib/rspec/matchers/built_in/exist.rb
+++ b/lib/rspec/matchers/built_in/exist.rb
@@ -81,7 +81,7 @@ module RSpec
           end
 
           def deprecated(predicate, actual)
-            predicate == :exists? && File == actual
+            predicate == :exists? && (File == actual || FileTest == actual)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/rspec/matchers/built_in/exist.rb
```diff
@@ -81,7 +81,7 @@ def predicates
           end
 
           def deprecated(predicate, actual)
-            predicate == :exists? && File == actual
+            predicate == :exists? && [File, FileTest].include?(actual)
```
pirj, line 84: Cosmetic: `File == actual || FileTest == actual`?

pirj (approved):
Thank you!

PR #1361, opened 2022-02-28
Add an_array_matching alias for match_array

markaschneider:
This improves readability when used as an argument matcher.

```
expect(obj).to receive(:msg).with(an_array_matching([1, 2]))
```

PR #1385, opened 2022-10-16
Synchronise changes from main / 3-11-maintenance to 4-0-dev

pirj:
This brings 4-0 closer to parity with main.

This is the result of `git merge main` and some minor tweaks to remove what was mistakenly re-added during merge.

⚠️ I did not merge in `3-11-maintenance`, but at a glance, they are on par with `main`.

- [ ] needs a 3.12.0 to release an enhancement from #1361

Similar to:
 - https://github.com/rspec/rspec-core/pull/2972
 - https://github.com/rspec/rspec-support/pull/549


JonRowe:
I've closed this because I prefer we don't do merges into the core branches like this, its better in my opinion to cherry-pick merge commits across (this is impossible in some cases where rebase merges have happened which is something I've been meaning to mention to you, we should always be doing merge commits or squash merges) the 4.0 situation is something I'm hoping to address in the next few weeks but if you want to cherry pick the prs across I'm happy for you to attempt that.

PR #1389, opened 2022-11-20
Raise an exception when a `nil` is passed as the first argument to `raise_error`

pirj:
Follow-up to #1143

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,8 @@ Breaking Changes:
   (Phil Pirozhkov, #1285)
 * Remove deprecated `match_regex` matcher. (Phil Pirozhkov, #1288)
 * Remove deprecated `StartAndEndWith` matcher base class. (Phil Pirozhkov, #1288)
+* Raise an exception when a `nil` is passed as the first argument to `raise_error`.
+  (Phil Pirozhkov, #1389)
 
 ### 3.12.0 / 2022-10-26
 [Full Changelog](http://github.com/rspec/rspec-expectations/compare/v3.11.1...v3.12.0)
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -15,6 +15,8 @@ Breaking Changes:
   (Phil Pirozhkov, #1285)
 * Remove deprecated `match_regex` matcher. (Phil Pirozhkov, #1288)
 * Remove deprecated `StartAndEndWith` matcher base class. (Phil Pirozhkov, #1288)
+* Raise an exception when a `nil` is passed as the first argument to `raise_error`.
```
JonRowe, line 18:
```suggestion
* Raise `ArgumentError` when a `nil` is passed as the first argument to `raise_error`.
```

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -192,17 +185,14 @@ def warn_about_bare_error!
...
-                         "`RSpec::Expectations.configuration.on_potential_false" \
-                         "_positives = :nothing`")
+        def raise_nil_argument_error!
+          raise "Using the `raise_error` matcher with a `nil` error is probably " \
+                "unintentional, it risks false positives, since `raise_error` " \
+                "will match when Ruby raises a `NoMethodError`, `NameError` or " \
+                "`ArgumentError`, potentially allowing the expectation to pass " \
+                "without even executing the method you are intending to call. " \
+                "#{warning}"\
+                "Instead consider providing a specific error class or message."
```
JonRowe, line 195:
```suggestion
          raise ArgumentError, "Using the `raise_error` matcher with a `nil` " \
                "value is probably unintentional, it risks false positives, since " \
                "`raise_error` will match when Ruby raises a `NoMethodError`" \
                ", `NameError` or `ArgumentError`, potentially allowing the " \
                "expectation to pass without even executing the method you "\
                "are intending to call. " \
                "#{warning}"\
                "Instead provide a specific error class or message."
```

JonRowe (changes requested):
Looks good, just a tweak in that we should raise an ArgumentError and I propose rewording the message slightly given now you have to avoid the `nil` value.

PR #1392, opened 2022-12-12
Handle MockExpectationError in aggregate_failures

mrzasa:
This PR fixes an error that breaks the RSpec formatter and causes RSpec to exit prematurely (without printing failed examples).

With `aggregate_failures`, a failed mock expectation returns instead of failing - that's expected. The unexpected part is that it returns a reference to the array of failures. As a result the internal data structure (`FailureAggregator#failures`) is available for modification in a test example (and thus - in the object under test). It may happen that the test example modifies the `FailureAggregator#failures` array. If objects added to this array don't quack like `StandardError`,`ExceptionPresenter` is unable to print them and raises error:
```
/home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/formatters/exception_presenter.rb:402:in `with_truncated_backtrace': undefined method `backtrace' for {:foo=>2}:Hash (NoMethodError)
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/formatters/exception_presenter.rb:376:in `block (2 levels) in sub_failure_list_formatter'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/flat_map.rb:7:in `each'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/flat_map.rb:7:in `each_with_index'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/flat_map.rb:7:in `each'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/flat_map.rb:7:in `flat_map'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/flat_map.rb:7:in `flat_map'
        from /home/mrzasa/.rvm/gems/ruby-3.0.4/gems/rspec-core-3.12.0/lib/rspec/core/formatters/exception_presenter.rb:367:in `block in sub_failure_list_formatter'
```

Minimal reproducible example: https://gist.github.com/mrzasa/fc6aef6098c76b3f201dad8187f5c18c

This PR provides a minimal fix - instead of returning a reference to the internal array, we return a copy of this array. This change inreases memory usage, but I believe this increase shouldn't be significant.

### DIFF
--- a/lib/rspec/expectations/failure_aggregator.rb
+++ b/lib/rspec/expectations/failure_aggregator.rb
@@ -12,7 +12,6 @@ module RSpec
           'supressed to be aggregated into our failure report by returning ' \
           'this value, further errors can be ignored.'
 
-
         def inspect
           MESSAGE
         end


spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -265,6 +265,39 @@ def expect_error_included_in_aggregated_failure(error)
...
+
+      it 'handles MockExpectationError correctly' do
+        expect {
+          aggregate_failures do
+            allow(factory).to receive(:get).with(100) { [] }
+
+            array = do_foo(factory, 2)
+
+            expect(array).to include({ :foo => 9 })
+            expect(factory).to have_receive(:get).with(100)
```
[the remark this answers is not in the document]
mrzasa: fixed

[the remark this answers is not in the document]
mrzasa: It wouldn't, because we need to use the value returned from the double. But indeed, your suggestion helped me to simplify the spec :)

pirj (changes requested):

spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -265,6 +265,24 @@ def expect_error_included_in_aggregated_failure(error)
...
     end
 
+    describe 'with MockExpectationError' do
+      it 'does not allow modifications of the failure list in the test example' do
+        expect {
+          aggregate_failures do
+            dbl = double
+            allow(double).to receive(:get).with(1) { [] }
+            array = dbl.get(2)
+            array << { :foo => :bar }
```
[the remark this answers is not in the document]
mrzasa: > you'll get a weird message:

          NoMethodError:
            undefined method `upcase' for [#<RSpec::Mocks::MockExpectationError: #<Double (anonymous)> 

That's what happens now and what I've seen in my real-world test runs. It's surprising indeed ;-).

It's a tricky issue. I did considered returning `nil` but I wanted to make a minimal change the original implementation. I think there is no good way to solve it. We can't return the object specified by `and_return` because the expectation is not met. Whatever else we return, we'll get `NoMethodError` sooner or later (because we're returning a wrong object).

All I can think of is making the error message less surprising. Maybe we could return a special object that would raise an informative message from its `method_missing`, sth like `Previous mock expectation wasn't met. This object is only a dummy replacement`. WDYT?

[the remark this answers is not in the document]
pirj: Strongly agree.
Since `array <<`/`string.upcase` are more likely to be happening in the SUT, and the chain of calls can be rather long, I'd prefer to fail early. And `nil` seems like a good tradeoff in terms of the clarity of the message and complexity of the handling.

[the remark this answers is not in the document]
mrzasa: OK, replacing with `nil` then

pirj (approved):
Thank you, @mrzasa !

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -48,6 +48,8 @@ def call(failure, options)
         @seen_source_ids[source_id] = true
         assign_backtrace(failure) unless failure.backtrace
         failures << failure
+
+        nil
```
JonRowe, line 52:
```suggestion
        FailedMethod
```

JonRowe: Where this is declared somewere as:
```
FailedMethod = :a_failed_method_return_value.freeze
```
Or something

JonRowe (changes requested):
Generally this looks fine and a good catch, but I think this should return a magic constant to indicate this is a failure.

lib/rspec/expectations/failure_aggregator.rb
JonRowe: It could also be a more sophisticated value object (but should still be frozen object).

pirj: Side note: symbols are frozen, no need to `.freeze` explicitly.

mrzasa: Thanks for the review!
I've added an `AggregatedFailure` value object, please take a look. I'll be OK to fall back to a simple symbol if you think it's a better solution

lib/rspec/expectations/aggregated_failure.rb
```diff
@@ -0,0 +1,26 @@
...
+      def self.build(failure)
+        new(failure).freeze
+      end
+
+      def initialize(failure)
+        @failure = failure
+      end
+
+      def inspect
+        "AggregatedFailure: #{@failure}"
```
JonRowe, line 22: I'd be ok with this being a static "singleton" instance within the other file that had an inspect of something like:
```
AggregatedFailure: This method caused a failure which has been supressed to be aggregated into our failure report by returning this value, further errors can be ignored.
```

but this doesn't need to be its own file or instance.,

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -48,6 +48,8 @@ def call(failure, options)
         @seen_source_ids[source_id] = true
         assign_backtrace(failure) unless failure.backtrace
         failures << failure
+
+        AggregatedFailure.build(failure)
```
JonRowe, line 52: My point about a magic object was to avoid an instantiation here, its simply to differentiate that its not a random nil being returned from the object, but a failed method call

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -4,6 +4,17 @@ module Expectations
     class FailureAggregator
       attr_reader :block_label, :metadata
 
+      class AggregatedFailure
```
JonRowe, line 7:
```suggestion
      # @private
      class AggregatedFailure
```

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -4,6 +4,17 @@ module Expectations
...
+        MESSAGE =
```
JonRowe, line 8:
```suggestion
        # @private
        MESSAGE =
```

spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -265,6 +265,36 @@ def expect_error_included_in_aggregated_failure(error)
...
+        }.to raise_error(
+          be_a(MultipleExpectationsNotMetError).and have_attributes(
+            :failures => [
+              be_a(RSpec::Mocks::MockExpectationError)
+            ],
+            :other_errors => [
+              # Checking only the class name, not the full message, because the hehaviour differes bettern Ruby 2 and 3.
+              # Ruby 3 uses #inspect in NoMethodError, while Ruby 2.x uses format that I can't override:
+              # NoMethodError (undefined method `<<' for #<RSpec::Expectations::FailureAggregator::AggregatedFailure:0x000055e4bac69ba8>)
+              # even if I override #to_s
```
pirj, line 290: 3.0.3:
```
#<NoMethodError: undefined method `<<' for AggregatedFailure: This method caused a failure which has been supressed to be aggregated into our failure report by returning this value, further errors can be ignored.:RSpec::Expectations::FailureAggregator::AggregatedFailure>
```
2.7.7:
```
#<NoMethodError: undefined method `<<' for #<RSpec::Expectations::FailureAggregator::AggregatedFailure:0x00007fdadf06ee08>>
```
🤔

spec/rspec/expectations/failure_aggregator_spec.rb
```diff
@@ -265,6 +265,36 @@ def expect_error_included_in_aggregated_failure(error)
       end
     end
 
+    describe 'with MockExpectationError' do
+      it 'does not allow modifications of the failure list in the test example' do
+        other_error_message =
+          'undefined method `<<\' for AggregatedFailure: This method caused a failure which has been supressed '\
+          'to be aggregated into our failure report by returning this value, further errors can be ignored.'\
+          ':RSpec::Expectations::FailureAggregator::AggregatedFailure'
+
```
JonRowe, line 274:
```suggestion
```

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -4,6 +4,19 @@ module Expectations
     class FailureAggregator
       attr_reader :block_label, :metadata
 
+      # @private
+      class AggregatedFailure
+        # @private
+        MESSAGE =
+          'AggregatedFailure: This method caused a failure which has been supressed to be aggregated into our failure report by returning this value, further errors can be ignored.'
```
JonRowe, line 11:
```suggestion
          'AggregatedFailure: This method caused a failure which has been ' \
          'supressed to be aggregated into our failure report by returning ' \
          'this value, further errors can be ignored.'

```

lib/rspec/expectations/failure_aggregator.rb
```diff
@@ -4,6 +4,22 @@ module Expectations
...
       attr_reader :block_label, :metadata
 
+      # @private
+      class AggregatedFailure
+        # @private
+        MESSAGE =
+          'AggregatedFailure: This method caused a failure which has been ' \
+          'supressed to be aggregated into our failure report by returning ' \
+          'this value, further errors can be ignored.'
+
```
JonRowe, line 14:
```suggestion
```

JonRowe:
Thanks! 🙌

JonRowe:
Released in 3.12.1

pirj:
Thanks a lot!

PR #1398, opened 2023-01-04
Fix `Dir#exists?` deprecation warning.

bigzed:
Fixes the deprecation warning for `Dir`.


spec/rspec/matchers/built_in/exist_spec.rb
```diff
@@ -131,6 +131,12 @@ def send
           expect(FileTest).to exist __FILE__
         end
       end
+
+      context 'Dir has deprecated exists?' do
+        it 'will not call exists? triggering the warning' do
+          expect(Dir).to exist __dir__
```
pirj, line 137:
```suggestion
          expect(Dir).to exist Dir.pwd
```

spec/rspec/matchers/built_in/exist_spec.rb
```diff
@@ -131,6 +131,12 @@ def send
           expect(FileTest).to exist __FILE__
         end
       end
+
+      context 'Dir has deprecated exists?' do
+        it 'will not call exists? triggering the warning' do
+          expect(Dir).to exist Dir.pwd
+        end
+      end
```
pirj, line 139:
```suggestion
      if RUBY_VERSION > "1.9"
        context 'Dir has deprecated exists?' do
          it 'will not call exists? triggering the warning' do
            expect(Dir).to exist Dir.pwd
          end
        end
      end
```

pirj:
Thank you!

bigzed:
Thank you too!

JonRowe:
Released in 3.12.2

PR #1416, opened 2023-04-12
Fix `include` matcher fuzzy matched key errors when encountering hash errors

JonRowe:
When a class that looks like a hash but has custom behaviour (such as Racks header hash which case insensitively compares keys and always assumes its asked for keys with a downcase method) we currently still try to shortcut key lookup using `key?` which than fails for fuzzy matching using matchers, instead ignore that error and use the fallback method.

### DIFF
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -153,8 +153,15 @@ module RSpec
         def actual_hash_has_key?(expected_key)
           # We check `key?` first for perf:
           # `key?` is O(1), but `any?` is O(N).
-          actual.key?(expected_key) ||
-          actual.keys.any? { |key| values_match?(expected_key, key) }
+
+          has_exact_key =
+            begin
+              actual.key?(expected_key)
+            rescue
+              false
+            end
+
+          has_exact_key || actual.keys.any? { |key| values_match?(expected_key, key) }
         end
 
         def actual_collection_includes?(expected_item)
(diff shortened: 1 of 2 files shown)


lib/rspec/matchers/built_in/include.rb
```diff
@@ -153,8 +153,15 @@ def comparing_hash_keys?(expected_item)
...
           # `key?` is O(1), but `any?` is O(N).
-          actual.key?(expected_key) ||
-          actual.keys.any? { |key| values_match?(expected_key, key) }
+
+          has_exact_key =
+            begin
+              actual.key?(expected_key)
+            rescue
+              false
+            end
```
pirj, line 162: Would the following be more intention-revealing?
```suggestion
          has_exact_key =
            if Matchers.is_a_matcher?(expected_key)
              false
            else
              actual.key?(expected_key)
            end
```

pirj (approved):
Fantastic, thank you!

lib/rspec/matchers/built_in/include.rb
JonRowe: No because I want to catch errors in custom implementations, if you matched on any non string key, rack would break, despite any object being a valid hash key.

pirj: This raises a question if our existing specs are sufficient to cover this. I’ve checked with my suggestion and nothing failed.
So it’s either the ‘is_a_matcher?’ that detects a regexp as a matcher, or the rack-like mock’s ‘key?’ that is not breaking when a non-string (e.g. regexp) is not breaking to mimic how rack does it.

JonRowe:
(This was released in 3.12.3)

Issue #1422, opened 2023-06-02
Include string encoding differences in failure messages


adfoster-r7:
### Your environment

* Ruby version: ruby 3.0.5p211 (2022-11-24 revision ba5cf0f7c5) [x86_64-darwin20]
* rspec-expectations version: 3.12.3

### Steps to reproduce

Create the following test file:

```ruby
require 'spec_helper'

RSpec.describe 'Example' do
  it 'Example' do
    actual = 'abc'.encode('UTF-16LE')
    expected = 'abc'.force_encoding('ASCII-8BIT')

    expect(actual).to eql(expected)
  end
end
```

### Expected behavior

If the string's encoding is the reason for failure, this information could additionally be shown to the user:

```
Example F

  1) Example Example
     Failure/Error: expect(actual).to eq(expected)
     
       expected: #<Encoding:ASCII-8BIT> "abc"
            got: #<Encoding:UTF-16LE> "abc"
     
       (compared using ==)
```

### Actual behavior

There's no visual indication as to why the string comparison as failed:

```
Example F

  1) Example Example
     Failure/Error: expect(actual).to eq(expected)
     
       expected: "abc"
            got: "abc"
     
       (compared using ==)
```

pirj:
Would you like to submit a PR to fix this?

adfoster-r7:
I'll take a look :+1:

PR #1425, opened 2023-06-29
Update matchers to show detected string encoding differences

AlanFoster:
closes https://github.com/rspec/rspec-expectations/issues/1422

Updates the `eq` and `eql` matchers to show detected string encoding differences - https://github.com/rspec/rspec-expectations/issues/1422

### Before

Unclear why the test is failing. For an average user this might be confusing:

```
Example F

  1) Example Example
     Failure/Error: expect(actual).to eq(expected)

       expected: "abc"
            got: "abc"

       (compared using ==)
```

### After

Immediately clear that the string encodings are not the same and the tests might be impacted by this difference:

```
Example F

  1) Example Example
     Failure/Error: expect(actual).to eq(expected)

       expected: #<Encoding:ASCII-8BIT> "abc"
            got: #<Encoding:UTF-16LE> "abc"

       (compared using ==)
```

### DIFF
--- a/lib/rspec/matchers/built_in/base_matcher.rb
+++ b/lib/rspec/matchers/built_in/base_matcher.rb
@@ -167,20 +167,35 @@ module RSpec
           # @return [Boolean] True if the actual and expected string encoding are different.
           #   i.e. the failure may be related to encoding differences and the encoding
           #   should be shown to the user. false otherwise.
-          def string_encoding_differs?
-            actual.is_a?(String) && expected.is_a?(String) &&
-              # Older Ruby versions do not support #encoding
-              actual.respond_to?(:encoding) && expected.respond_to?(:encoding) &&
-              actual.encoding != expected.encoding
+          if String.method_defined?(:encoding)
+            def string_encoding_differs?
+              actual.is_a?(String) && expected.is_a?(String) && actual.encoding != expected.encoding
+            end
+          else
+            # @api private
+            # @return [Boolean] False always as the curent Ruby version does not support String encoding
+            def string_encoding_differs?
+              false
+            end
           end
           module_function :string_encoding_differs?
 
-          # @api private
-          # Formats a String's encoding as a human readable string
-          # @param value [String]
-          # @return [String]
-          def format_encoding(value)
-            "#<Encoding:#{value.encoding.name}>"
+          if String.method_defined?(:encoding)
+            # @api private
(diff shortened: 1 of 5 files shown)


AlanFoster:
> 1 workflow awaiting approval
> This workflow requires approval from a maintainer.

@pirj Sorry for the ping, CI approval would be appreciated 🎉

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -161,6 +161,32 @@ def improve_hash_formatting(inspect_string)
...
+          # @api private
+          # @return [Boolean] True if the actual and expected string encoding are different.
+          #   i.e. the failure may be related to encoding differences and the encoding
+          #   should be shown to the user. false otherwise.
+          def string_encoding_differs?
+            actual.is_a?(String) && expected.is_a?(String) &&
+              # Older Ruby versions do not support #encoding
+              actual.respond_to?(:encoding) && expected.respond_to?(:encoding) &&
+              actual.encoding != expected.encoding
+          end
```
JonRowe, line 175: We prefer to conditionally define things like this:
```suggestion
          if String.method_defined?(:encoding)
            def string_encoding_differs?
              actual.is_a?(String) && expected.is_a?(String) && actual.encoding != expected.encoding
            end
          else
            def string_encoding_differs?
              false
            end
          end
```

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -161,6 +161,32 @@ def improve_hash_formatting(inspect_string)
...
+          module_function :string_encoding_differs?
+
+          # @api private
+          # Formats a String's encoding as a human readable string
+          # @param value [String]
+          # @return [String]
+          def format_encoding(value)
+            "#<Encoding:#{value.encoding.name}>"
+          end
```
JonRowe, line 184: This should also be conditionally defined to protect against accidental usage

JonRowe (changes requested):
Stylistically we prefer to conditonally define methods rather than do runtime checks for old rubies, makes it easier to remove the backports when the time comes. We do a lot of similar work on encoded string in rspec-support

JonRowe:
Thanks 🙏

JonRowe:
Released in 3.13.0 apologies for the delay.

PR #1428, opened 2023-08-13
Prevent custom matcher is called twice without compound

AlecksJohannes:
Hey lovely folks :wave:, I hope you all are doing well 

This PR here is to address https://github.com/rspec/rspec/issues/142 which is to prevent a custom matcher from calling twice without compound like `.and` or `.or`

### DIFF
--- a/spec/rspec/matchers/dsl_spec.rb
+++ b/spec/rspec/matchers/dsl_spec.rb
@@ -81,6 +81,48 @@ RSpec.describe "a matcher defined using the matcher DSL" do
     expect(4).to be_just_like(4)
   end
 
+  describe 'check method call count' do
+    RSpec::Matchers.define :be_fine do
+      match { |actual| expect(actual).to eq('ok') }
+    end
+
+    RSpec::Matchers.define :have_keys do |attributes|
+      match { |actual| match_keys(actual, attributes) }
+
+      def match_keys(actual, attributes)
+        if attributes.is_a?(Array)
+          attributes.each do |attr|
+            expect(actual).to have_keys(attr)
+          end
+        else
+          expect(actual).to include(attributes)
+        end
+      end
+    end
+
+    it do
+      expect(RSpec).not_to receive(:warning)
+      expect(ok).to be_fine.and be_fine
+      expect(ok).to be_fine.or be_fine
+    end
+
+    context 'when custom matcher is called recursively' do
+      it 'does not raise a warning' do
+        expect(RSpec).not_to receive(:warning)
+        expect({ 'user' => 'Minh', 'id' => 1 }).to have_keys(%w[user id])
+      end
+    end
+
+    context 'when a custom matcher is called twice without using compound' do
+      it 'raises a warning' do
+        expect(RSpec).to receive(:warning).with(/The custom matcher `be_fine` or its negated method is called/)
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks for starting to tackle this, the issue as I see it is somehow this method is being responded to by the matcher, this is what needs to be fixed `.matcher().matcher` should cause a natural `NoMethodError` e.g.

```
     NoMethodError:
       undefined method `eq' for #<RSpec::Matchers::BuiltIn::Eq:0x00000001111948f0 @expected="ok">

               expect(ok).to eq(ok).eq
```

spec/rspec/matchers/dsl_spec.rb
```diff
@@ -479,20 +521,20 @@ def foo
     end
 
     it "allows helper methods to be defined with #define_method to have access to matcher parameters" do
-      matcher = new_matcher(:name, 3, 4) do |a, b|
+      matcher = new_matcher(:matcher, 3, 4) do |a, b|
```
JonRowe, line 524: Please revert these

AlecksJohannes:
> Thanks for starting to tackle this, the issue as I see it is somehow this method is being responded to by the matcher, this is what needs to be fixed `.matcher().matcher` should cause a natural `NoMethodError` e.g.
> 
> ```
>      NoMethodError:
>        undefined method `eq' for #<RSpec::Matchers::BuiltIn::Eq:0x00000001111948f0 @expected="ok">
> 
>                expect(ok).to eq(ok).eq
> ```

Hello @JonRowe  thank you, sure I can raise `NoMethodError` rather than using `RSpec.warning` WDYT ?

JonRowe:
> Hello @JonRowe thank you, sure I can raise NoMethodError rather than using RSpec.warning. WDYT ?

The fix should be to find where the method is being responded to and prevent it, I suspect its probably including a module it shouldn't do.

AlecksJohannes:
> > Hello @JonRowe thank you, sure I can raise NoMethodError rather than using RSpec.warning. WDYT ?
> 
> The fix should be to find where the method is being responded to and prevent it, I suspect its probably including a module it shouldn't do.

Hey there JonRowe, sorry I have updated my comment above.

AlecksJohannes:
I'll paste it here just to be sure 

I believe I understand what you said and I agree with you too. I tried to do that but I am not sure about removing or adjusting this piece of code, I played around with our code before and I can make it to work exactly like you said but we need to deal with 👇 

```rb
def respond_to_missing?(method, include_private=false)
   super || @matcher_execution_context.respond_to?(method, include_private)
end

def method_missing(method, *args, &block)
   if @matcher_execution_context.respond_to?(method)
      @matcher_execution_context.__send__ method, *args, &block
   else
       super(method, *args, &block)
   end
end
```

Otherwise, even calling `matcher().matcher` it will still work even though the method does not exist. Because of this blocker,  I had to provide another solution

P/S: If we were somehow to remove that `method_missing` I think you won't be able to call that custom matcher recursively which I think resulting in a breaking changes.

pirj:
What concerns me the most, why are matchers available as methods on an i stance of a matcher in the first place?

JonRowe:
Migrated to the monorepo as rspec/rspec#148

PR #1436, opened 2023-11-11
New matcher idea: `match_pattern` for Ruby's pattern-matching

r7kamura:
## Summary

Pattern-matching has been available since Ruby 3.0. How about providing this as a RSpec matcher?

### Basic example

```ruby
# Pass
expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])

# Fail
expect([1, 2, 3]).to match_pattern([Integer, Integer, String])
```

### More realistic example

```ruby
node = Nokogiri::HTML5.parse(response.body).at('.foo')
expect(node).to match_pattern(name: 'a', href: 'http://example.com')
```

This is done using nokogiri's pattern-matching feature:

- https://github.com/sparklemotion/nokogiri/discussions/2360
- https://github.com/sparklemotion/nokogiri/pull/2523

## Background

As a background, I am usually involved in Rails app development, and since recent Rails makes it easier to do pattern matching using `response.parsed_body`, so I thought it would be nice if RSpec had the ability to support this.

- https://github.com/rails/rails/pull/49003

## Implementation

I have come up with two implementation ideas and I think either is fine, but the latter seems closer to what RSpec is aiming for, so I decided to propose this one for the time being, even if it is a bit magical.

The former has its advantages here because it is simple to implement and easy to understand. Please let me know what you think.

### Using block

First in thinking about this, I referred to `assert_pattern` in minitest:

- https://github.com/minitest/minitest/pull/936

Minitest's pattern-matching feature provides an interface that encloses the entire expression in a block, as shown below:

```ruby
assert_pattern { [1, 2, 3] => [Integer, Integer, Integer] }
```

A similar implementation could be done in RSpec. In short, we can have a `match_pattern` that does the following:

```ruby
expect { [1, 2, 3] => [Integer, Integer, Integer] }.not_to raise_error(NoMatchingPatternError)
expect { [1, 2, 3] => [Integer, Integer, Integer] }.to match_pattern
``` 

### Using `instance_eval`

In RSpec, it is common to write `expect(actual).to ... `, so it would be nice if we could provide pattern-matching in a way that follows this convention.

I thought of a way to achieve this by using `inspect` and `instance_eval` to carry around the given expected pattern as a value.

```ruby
expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])
```

In this implementation, the following code will be evaluated in the above code:

```ruby
[1, 2, 3] in [Integer, Integer, Integer]
```

### DIFF
--- a/features/built_in_matchers/match_pattern.feature
+++ b/features/built_in_matchers/match_pattern.feature
@@ -1,0 +1,27 @@
+Feature: `match_pattern` matcher
+
+  Use the `match_pattern` matcher to specify that a value matches with
+  expected patterns by Ruby's pattern-matching.
+
+    ```ruby
+    expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])
+    expect([1, 2, 3]).to match_pattern([Integer, Integer, String])
+    ```
+
+  Scenario: Pattern usage
+    Given a file named "example_spec.rb" with:
+      """ruby
+      RSpec.describe [1, 2, 3] do
+        it { is_expected.to match_pattern([Integer, Integer, Integer]) }
+        it { is_expected.not_to match_pattern([Integer, Integer, String]) }
+
+        # deliberate failures
+        it { is_expected.to match_pattern([Integer, Integer, String]) }
+        it { is_expected.not_to match_pattern([Integer, Integer, Integer]) }
+      end
+      """
+    When I run `rspec example_spec.rb`
+    Then the output should contain all of these:
+      | 4 examples, 2 failures                                              |
+      | expected [1, 2, 3] to match pattern [Integer, Integer, String]      |
+      | expected [1, 2, 3] not to match pattern [Integer, Integer, Integer] |
--- a/lib/rspec/matchers/built_in/match_pattern.rb
+++ b/lib/rspec/matchers/built_in/match_pattern.rb
@@ -1,0 +1,47 @@
+module RSpec
+  module Matchers
+    module BuiltIn
+      # @api private
+      # Provides the implementation for `match_pattern`.
+      # Not intended to be instantiated directly.
+      class MatchPattern < BaseMatcher
+        def match(expected, actual) # rubocop:disable Lint/UnusedMethodArgument
+          raise_not_supported_ruby_version_error unless supported_ruby_version?
+
+          begin
+            instance_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+              actual in #{expected.inspect}
+            RUBY
+          rescue SyntaxError
+            raise_invalid_pattern_error
+          end
+        end
+
+        def failure_message
+          "expected #{description_of(@actual)} to match pattern #{@expected.inspect}"
+        end
+
+        def failure_message_when_negated
+          "expected #{description_of(@actual)} not to match pattern #{@expected.inspect}"
+        end
+
+        private
+
+        def raise_invalid_pattern_error
+          raise SyntaxError, "The #{matcher_name} matcher requires that " \
+                               "the expected object can be used as Ruby's " \
+                               "pattern-matching but a `SyntaxError` was raised instead."
+        end
+
+        def raise_not_supported_ruby_version_error
+          raise NotImplementedError, "The #{matcher_name} matcher is only " \
+                                     "supported on Ruby 3 or higher."
+        end
+
+        def supported_ruby_version?
+          RUBY_VERSION.to_f >= 3.0
+        end
+      end
+    end
+  end
+end
--- a/spec/rspec/matchers/built_in/match_pattern_spec.rb
+++ b/spec/rspec/matchers/built_in/match_pattern_spec.rb
@@ -1,0 +1,40 @@
+RSpec.describe 'match_pattern matcher' do
+  it_behaves_like 'an RSpec value matcher', :valid_value => [1, 2, 3],
+                                            :invalid_value => [1, 2, '3'] do
+    let(:matcher) { match_pattern([Integer, Integer, Integer]) }
+  end
+
+  context 'when expected pattern matches with actual' do
+    it 'passes' do
+      expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])
+    end
+  end
+
+  context 'when expected pattern does not match with actual' do
+    it 'fails' do
+      expect {
+        expect([1, 2, 3]).to match_pattern([Integer, Integer, String])
+      }.to fail_with('expected [1, 2, 3] to match pattern [Integer, Integer, String]')
+    end
+  end
+
+  context 'when expected pattern cannot be used as pattern-matching' do
+    it 'raises SyntaxError' do
+      expect {
+        expect([1, 2, 3]).to match_pattern(Object.new)
+      }.to raise_error(SyntaxError, "The match_pattern matcher requires that the expected object can be used as Ruby's pattern-matching but a `SyntaxError` was raised instead.")
+    end
+  end
+
+  context 'when pattern-matching is not supported on the current Ruby version' do
+    before do
+      stub_const('RUBY_VERSION', '2.7.0')
+    end
+
+    it 'raises NotImplementedError' do
+      expect {
+        expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])
+      }.to raise_error(NotImplementedError, 'The match_pattern matcher is only supported on Ruby 3 or higher.')
+    end
+  end
+end
--- a/lib/rspec/matchers.rb
+++ b/lib/rspec/matchers.rb
@@ -719,6 +719,17 @@ module RSpec
       desc.sub("contain exactly", "an array containing exactly")
     end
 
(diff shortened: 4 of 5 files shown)


pirj:
@zverok what do you think?

zverok:
So... Just thinking out loud!

My design space analysis is as follows:

1. I personally would like any new matcher to be fully on the right side of `to` (i.e., allowing to always write `expect(subject).to matcher` or `expect { subject }.to matcher`, without any change to the left side — also allowing implicit subjects); which, for me, excludes the first option.
2. I personally believe I would be confused if the new feature would be _pretending_ to accept a Ruby's pattern-matching, but actually wouldn't, i.e., accepted some subset of Ruby that "looks like" some patterns but wouldn't be handled with the same syntactical rules actually. I believe it is a great source of possible confusion. There are many valid options that people _might_ want, and which are valid patterns yet invalid in other contexts (and will be therefore invalid in this "imitation" syntax), like find patterns `[*, {id: 3}, *]`, variable pinning: `{id:, ... {linked_id: ^id}}`, or even just (if "we have a full pattern matching here! it is more expressive!") trying to use `{x:, y:}` instead of `hash_including(:x, :y)`. Which, basically, makes the second option undesirable.
3. An aside note: Ideally, and for realistic large data structures (like JSON responses), it would be cool to provide a detailed report of "which part exactly didn't match"

So, what options do I see viable for this matcher?

1\. A block on the right side that re-accepts the subject and just allows to write the check:
```ruby
expect([1, 2, 3]).to match_pattern { _1 in [Integer, String, Integer] }
# expected to match pattern [Integer, String, Integer], but didn't match
```
This is currently possible in almost the same form this way:
```ruby
expect([1, 2, 3]).to satisfy { _1 in [Integer, String, Integer] }
# expected [1, 2, 3] to satisfy expression `_1 in [Integer, String, Integer]`
```
(I believe the expression is extracted by just reading the file source, and the same can be done by `match_pattern`, just cleaning up `_1 in` part)

2\. The same block, yet also (or only) expecting you to write `=>` which then analyzes the error:
```ruby
expect([1, 2, 3]).to match_pattern { _1 => [Integer, String, Integer] }
# expected to match pattern [Integer, String, Integer], but didn't match:
#  String === 2 does not return true
```
Here, we are doing the same as above yet also catching the `NoMatchingPattern` error and extract the relevant part of text from it. 

(That's how it will fail in this case... I don't think it is ideal, and probably there is some room for improvement _in Ruby core_ with good proposals for better structuring "what haven't matched" info, and probably even providing it in a structured form in the exception object!)

Obviously, both forms would be less appealing for those who HATE numbered block params (and those people are plenty!), so they'll need to write
```ruby
expect([1, 2, 3]).to match_pattern { |val| val => [Integer, String, Integer] }
```

With a bit of trickery (providing proper evaluation context), we can provide some nice name available inside a block instead (but this "trickery" might lead to undesirable effects either, and deciding on the name is not easy):
```ruby
expect([1, 2, 3]).to match_pattern { value => [Integer, String, Integer] }
```

3\. The third idea is much more unholy _and also currently impossible_, but looks more appealing as a matcher syntax:
```ruby
expect([1, 2, 3]).to match_pattern => [Integer, String, Integer]
```
In my imagination, it would work by having `match_pattern` just provide an object to match (the same as subject or some wrapper), and then adding some "custom handler" to RSpec core, which will catch `NoMatchingPattern` error, and treat it as "that matcher that added the handler, have failed". This is kinda tricky, but wouldn't work in the current Ruby nevertheless: `=>` has a "void context" (same as, say, `return`), and can't be put in the expression in this place.

4\. The fourth one is a compromise on the above:
```ruby
expect([1, 2, 3]).to { match_pattern => [Integer, String, Integer] }
```
This _is_ valid syntax, and while also requires adjusting RSpec core, the adjustment seems more understandable than "adding a global handler" (in this case, the exception needs to be handled in the limited scope of the "block passed to `to`", not in the scope of the "entire example"). Maybe fiddling a bit around "how it can be implemented" might lead to some insights about adjustments in RSpec core (`expect`/`to`) that would be useful... Or maybe not!

Anyway, atm, honestly, no option looks super-persuasive for myself either. Would I need a lot of PM-in-spec (unfortunately, on my dayjob I still stuck with Ruby 2.7 where pattern matching was experimental and we prefer to avoid that unfinished-yet syntax), I would start with just `satisfy { _1 in pattern }` and after several weeks/months would try to rethink it and see whether it requires some more tinkering.

r7kamura:
Very thoughtful input, thank you!

Indeed, the `instance_eval` idea is not a good one where a lot of pattern-matching syntax doesn't actually work properly.

The `expect([1, 2, 3]).to match_pattern { |val| val => [Integer, String, Integer] }` idea seems the most reasonable of the current proposals for me, although I think the drawback is that it looks a bit clunky. Even from my personal observation, the use of numbered parameters does not seem to be widespread yet, nor does it seem to be that popular, so many people will probably use the normal block argument format.

Regarding the "what haven't matched" info, I was thinking that since pattern matching has variable capturing feature, it would be nice to be able to partially verify it using aggregated failures as follows:

```ruby
expect(hash).to match_pattern { _1 => { user: user } }
expect(user).to match_pattern { _1 => { name: 'foo' } }
```
However, blocks create local variable scopes, so we have to do something tricky to make this work well.

Thinking about it this way, I wonder if it is no longer necessary to use any matcher and just do the assignment as usual...🤔 

```ruby
it 'returns something' do
  subject => { user: user }
  user => { name: 'foo' }
end
```

r7kamura:
By the way, I heard that `it` might be available in Ruby 3.4 instead of `_1`.

- https://bugs.ruby-lang.org/issues/18980

If that happens, we may be able to write the following:

```ruby
expect([1, 2, 3]).to match_pattern { it => [Integer, String, Integer] }
```

But interestingly (unfortunately), the naming conflicts with RSpec's `it`, which is very, very confusing...

spec/rspec/matchers/built_in/match_pattern_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'raises SyntaxError' do
+      expect {
+        expect([1, 2, 3]).to match_pattern(Object.new)
+      }.to raise_error(SyntaxError, "The match_pattern matcher requires that the expected object can be used as Ruby's pattern-matching but a `SyntaxError` was raised instead.")
+    end
+  end
+
+  context 'when pattern-matching is not supported on the current Ruby version' do
+    before do
+      stub_const('RUBY_VERSION', '2.7.0')
```
JonRowe, line 31: Considering we run builds on multiple rubies, you don't need to do this, you can add a flag to `rspec-support` for wether pattern matching is supported, then flag one set of tests as `:if => ...` and the other has `:unless => ...`

lib/rspec/matchers/built_in/match_pattern.rb
```diff
@@ -0,0 +1,47 @@
...
+        end
+
+        def raise_not_supported_ruby_version_error
+          raise NotImplementedError, "The #{matcher_name} matcher is only " \
+                                     "supported on Ruby 3 or higher."
+        end
+
+        def supported_ruby_version?
+          RUBY_VERSION.to_f >= 3.0
+        end
```
JonRowe, line 43: We would typically conditionally define the entire matcher on supported rubies, and then have a dummy implementated that raises the not implemented error.

lib/rspec/matchers/built_in/match_pattern.rb
```diff
@@ -0,0 +1,47 @@
...
+      # @api private
+      # Provides the implementation for `match_pattern`.
+      # Not intended to be instantiated directly.
+      class MatchPattern < BaseMatcher
+        def match(expected, actual) # rubocop:disable Lint/UnusedMethodArgument
+          raise_not_supported_ruby_version_error unless supported_ruby_version?
+
+          begin
+            instance_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+              actual in #{expected.inspect}
```
JonRowe, line 13: If theres a way to avoid using eval that would be ideal...

features/built_in_matchers/match_pattern.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  Use the `match_pattern` matcher to specify that a value matches with
+  expected patterns by Ruby's pattern-matching.
+
+    ```ruby
+    expect([1, 2, 3]).to match_pattern([Integer, Integer, Integer])
+    expect([1, 2, 3]).to match_pattern([Integer, Integer, String])
+    ```
+
+  Scenario: Pattern usage
```
JonRowe, line 11: This would also have to be tagged and excluded on unsupported ruby, and as it is used in our docs we would need a note about supported ruby versions in that,

JonRowe (commented):
I like this idea but would have a preference for this sort of style:
```
expect([1, 2, 3]).to match_pattern { _1 in [Integer, String, Integer] }
```

JonRowe:
Migrated to the monorepo rspec/rspec#149

PR #1438, opened 2023-12-09
Display default and custom failure messages together

jgaskins:
I've never needed to specify a custom message to a matcher in 14 years of using RSpec, but apparently some folks find it useful to have both the original message and the custom message. I don't really disagree with them. Anything that helps understand the failure more.

And honestly I'm doing this so maybe other open source maintainers can have less ammunition to [shit on other people's open source software](https://ruby.social/@zenspider/111547919694702860) for no reason. 🙃 Maybe OSS maintainers that do things like that might, I dunno, open a PR with the change they'd like to see instead of just whining on the internet if they realize it's possible. It could happen. I've seen stranger things.

### DIFF
--- a/lib/rspec/expectations/handler.rb
+++ b/lib/rspec/expectations/handler.rb
@@ -32,7 +32,10 @@ module RSpec
 
       def self.handle_failure(matcher, message, failure_message_method)
         message = message.call if message.respond_to?(:call)
-        message ||= matcher.__send__(failure_message_method)
+        message = [message, matcher.__send__(failure_message_method)]
+          .compact
+          .uniq
+          .join("\n")
 
         if matcher.respond_to?(:diffable?) && matcher.diffable?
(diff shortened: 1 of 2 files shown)


pirj:
I’m not opposed to that, but the reasoning is unclear.
I would use a custom message if the regular is unwieldy, eg a diff of a hundred lines.
Just like you, I don’t think I’ve ever used this feature, and I don’t think that reacting to public hate speech is a natural way to evolve.
I happily merge PRs provided rationale and stating someone’s need.

JonRowe:
If the author of minitest wishes to prove a point by deliberatly overridding our failure messages with a custom failure message that doesn't include the values he is welcome to, that is the point of providing a custom failure message.

jgaskins:
That all makes perfect sense. I very much appreciate y'all's replies, both in this PR and on Mastodon. 💚

PR #1455, opened 2024-05-07
Handle dynamic mocks specially for null-object doubles

nevinera:
This hopefully resolves #1132 and #1288 - essentially, if you leave `strict_predicate_matchers` disabled and use a null-object spy like this:

```
let(:thing) { spy('my thing') }

it "does the thing right" do
  do_it
  expect(thing).to have_receivedd(:foo)
end
```

It'll pass. Note the typo in `have_receivedd`? The dynamic matcher will call `has_receivedd?` on the spy, the spy will return itself (that's what it does), and the matcher will note that the result is truthy, and pass the spec. That's kind of an issue - turning off that setting _does_ solve it, but that won't be the default until rspec 4, and even then it's not the _ideal_ behavior, just a lot better.

This PR updates the dynamic matchers to notice when their target is a Double, and change behavior slightly - for Doubles, don't just check if they respond to the method, _also_ check if it's _in the `methods` list_. That'll suffice to confirm that it's explicitly mocked on the double, rather than being an automatic response.

### DIFF
--- a/lib/rspec/matchers/built_in/has.rb
+++ b/lib/rspec/matchers/built_in/has.rb
@@ -49,15 +49,18 @@ module RSpec
         # Catch a semi-frequent typo - if you have strict_predicate_matchers disabled and
         # expect(spy).to have_receieveddd(:foo) it would be evergreen - the dynamic matcher
         # queries `has_receiveddd?`, the spy _fakes_ the method, returning its (truthy) self.
-        def really_responds_to?(method)
-          if defined?(RSpec::Mocks::Double) && @actual.is_a?(RSpec::Mocks::Double)
-            @actual.respond_to?(method) && methods_include?(method)
-          else
+        if defined?(RSpec::Mocks::Double)
+          def really_responds_to?(method)
+            if RSpec::Mocks::Double === @actual
+              @actual.respond_to?(method) && methods_include?(method)
+            else
+              @actual.respond_to?(method)
+            end
+          end
+        else
+          def really_responds_to?(method)
             @actual.respond_to?(method)
           end
-        rescue NoMethodError
-          # Proxied BasicObjects don't have `is_a?`, or basically anything else.
-          @actual.respond_to?(method)
         end
 
         def predicate_accessible?
--- a/lib/rspec/matchers/built_in/has.rb
+++ b/lib/rspec/matchers/built_in/has.rb
@@ -55,6 +55,9 @@ module RSpec
           else
             @actual.respond_to?(method)
           end
+        rescue NoMethodError
+          # Proxied BasicObjects don't have `is_a?`, or basically anything else.
+          @actual.respond_to?(method)
         end
 
         def predicate_accessible?
--- a/lib/rspec/matchers/built_in/has.rb
+++ b/lib/rspec/matchers/built_in/has.rb
@@ -46,8 +46,19 @@ module RSpec
 
       private
 
+        # Catch a semi-frequent typo - if you have strict_predicate_matchers disabled and
+        # expect(spy).to have_receieveddd(:foo) it would be evergreen - the dynamic matcher
+        # queries `has_receiveddd?`, the spy _fakes_ the method, returning its (truthy) self.
+        def really_responds_to?(method)
+          if defined?(RSpec::Mocks::Double) && @actual.is_a?(RSpec::Mocks::Double)
+            @actual.respond_to?(method) && @actual.methods.include?(method)
+          else
+            @actual.respond_to?(method)
+          end
+        end
+
         def predicate_accessible?
-          @actual.respond_to? predicate
+          really_responds_to?(predicate)
         end
 
         # support 1.8.7, evaluate once at load time for performance
@@ -155,7 +166,7 @@ module RSpec
         end
 
         def predicate_accessible?
-          super || actual.respond_to?(present_tense_predicate)
+          super || really_responds_to?(present_tense_predicate)
         end
 
         def present_tense_predicate
--- a/lib/rspec/matchers/built_in/has.rb
+++ b/lib/rspec/matchers/built_in/has.rb
@@ -51,7 +51,7 @@ module RSpec
         # queries `has_receiveddd?`, the spy _fakes_ the method, returning its (truthy) self.
         def really_responds_to?(method)
           if defined?(RSpec::Mocks::Double) && @actual.is_a?(RSpec::Mocks::Double)
-            @actual.respond_to?(method) && @actual.methods.include?(method)
+            @actual.respond_to?(method) && methods_include?(method)
           else
             @actual.respond_to?(method)
           end
@@ -70,10 +70,18 @@ module RSpec
           def private_predicate?
             @actual.private_methods.include? predicate.to_s
           end
+
+          def methods_include?(method)
+            @actual.methods.include?(method.to_s)
+          end
           # :nocov:
         else
           def private_predicate?
             @actual.private_methods.include? predicate
+          end
+
+          def methods_include?(method)
+            @actual.methods.include?(method)
           end
         end
 
--- a/spec/rspec/matchers/built_in/be_spec.rb
+++ b/spec/rspec/matchers/built_in/be_spec.rb
@@ -87,7 +87,7 @@ RSpec.describe "expect(...).to be_predicate" do
     end
 
     it "allows dynamic matchers to pass for supplied methods on spies" do
-      thing = spy("thing", has_recv?: true)
+      thing = spy("thing", :has_recv? => true)
       expect(thing).to have_recv(:foo)
       expect(thing).to have_received(:has_recv?).with(:foo)
     end
--- a/spec/rspec/matchers/built_in/be_spec.rb
+++ b/spec/rspec/matchers/built_in/be_spec.rb
@@ -85,6 +85,21 @@ RSpec.describe "expect(...).to be_predicate" do
         expect(actual).to be_happy
       }.to fail_with("expected `#{actual.inspect}.happy?` to be truthy, got false")
     end
+
+    it "does not allow non-supplied dynamic matchers to pass on spies" do
+      thing = spy("thing", has_recv?: true)
+
+      expect(thing).to have_recv(:foo)
+      expect(thing).to have_received(:has_recv?).with(:foo)
+
+      begin
+        expect(thing).to have_receivex(:foo)
+      rescue RSpec::Expectations::ExpectationNotMetError
+        @reached = true
+      end
+      expect(@reached).to be_truthy
+      expect(thing).not_to have_received(:has_receivex?)
+    end
   end
 
   it "fails when actual does not respond to :predicate?" do
--- a/spec/rspec/matchers/built_in/be_spec.rb
+++ b/spec/rspec/matchers/built_in/be_spec.rb
@@ -86,19 +86,18 @@ RSpec.describe "expect(...).to be_predicate" do
       }.to fail_with("expected `#{actual.inspect}.happy?` to be truthy, got false")
     end
 
-    it "does not allow non-supplied dynamic matchers to pass on spies" do
+    it "allows dynamic matchers to pass for supplied methods on spies" do
       thing = spy("thing", has_recv?: true)
-
       expect(thing).to have_recv(:foo)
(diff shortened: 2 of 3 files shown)


spec/rspec/matchers/built_in/be_spec.rb
```diff
@@ -85,6 +85,21 @@
         expect(actual).to be_happy
       }.to fail_with("expected `#{actual.inspect}.happy?` to be truthy, got false")
     end
+
+    it "does not allow non-supplied dynamic matchers to pass on spies" do
+      thing = spy("thing", has_recv?: true)
+
+      expect(thing).to have_recv(:foo)
+      expect(thing).to have_received(:has_recv?).with(:foo)
```
pirj, line 93: Can you please split this into two examples? Otherwise the one that covers the “really” part is easily missed.

spec/rspec/matchers/built_in/be_spec.rb
```diff
@@ -85,6 +85,21 @@
...
+
+      begin
```
pirj, line 95: Can we say that this has two checks for the same thing that won’t fail separately?
One is that the method wasn’t called on the spy, and another that the expectation failed?

What if we just

expect {
         expect(thing).to have_receivex(:foo)
         }.to fail_with(/some message/)

I recall this is pretty common in our specs

pirj (commented):
Looks mostly good, thanks for the fix!

pirj:
> 22 deprecation warnings total

Wondering why we don’t print them 🧐

spec/rspec/matchers/built_in/be_spec.rb
```diff
@@ -85,6 +85,21 @@
...
+        expect(thing).to have_receivex(:foo)
+      rescue RSpec::Expectations::ExpectationNotMetError
+        @reached = true
+      end
+      expect(@reached).to be_truthy
+      expect(thing).not_to have_received(:has_receivex?)
+    end
```
JonRowe, line 102:
```suggestion
    it "does not allow non-supplied dynamic matchers to pass on spies" do
      thing = spy("thing")

      expect {
        expect(thing).to have_receivex(:foo)
      }.to fail
    end
    
    it "allows dynamic matchers to pass on spies" do
      thing = spy("thing", has_recv?: true)
      
      expect(thing).to have_recv(:foo)
      expect(thing).to have_received(:has_recv?).with(:foo)
    end
```

spec/rspec/matchers/built_in/has_spec.rb
```diff
@@ -159,6 +159,21 @@ def o.has_sym?(sym); sym == :foo; end
...
+      expect(thing).to have_received(:furg?).with(:foo)
+
+      begin
+        expect(thing).to be_blurm(:foo)
+      rescue RSpec::Expectations::ExpectationNotMetError
+        @reached = true
+      end
+      expect(@reached).to be_truthy
+      expect(thing).not_to have_received(:blurm?)
+    end
```
JonRowe, line 176:
```suggestion
    it "does not allow non-supplied dynamic matchers to pass on spies" do
      thing = spy("thing")

      expect {
        expect(thing).to be_blurm(:foo)
      }.to fail
    end
    
    it "allows dynamic matchers to pass on spies" do
      thing = spy("thing", furg?: true)
      
      expect(thing).to be_furg(:foo)
      expect(thing).to have_received(:furg?).with(:foo)
```

JonRowe (changes requested):
Can you please use our fail matcher in the specs

nevinera:
> Can you please use our fail matcher in the specs

Thank you! This was so ugly, I should have known there would be a matcher for that :-D

spec/rspec/matchers/built_in/be_spec.rb
nevinera: I was wavering on doing that in the first place, but without the fail-matcher the specs were so big and ugly that I didn't want to double them up. You guys were responding so fast, I figured it'd be better to just get something up and find out what you liked :-)

spec/rspec/matchers/built_in/be_spec.rb
nevinera: If I understand the question, then yes - the second assertion in each of the updated tests (did the spy register receiving the method) _could_ be tested separately, if that's preferred.

I've always had a little trouble deciding where the boundary should be for testing actions that have multiple related effects like this, but since the first part of the spec is probably the simplest way to set up the second part of the spec in each case, I think one of the new tests in each case would be identical to the current test, and the other would be just the first segment of each? My threshold for splitting in those cases has (lately) just been "can I phrase the test description cleanly enough to be clear with them together?".

In some of my projects I use a `rescued { do_the_thing }` spec helper method, if we have some analogue to that (or would like one) it would make the split a little cleaner, though with the fail-matcher it's probably less useful than usual.

Up to you, happy to work it however :-)

nevinera:
```
  1) Object#should on interpreters that have BasicObject does not break the deprecation check on BasicObject-subclassed proxy objects
     Failure/Error: @target.send(name, *args)

     NoMethodError:
       undefined method `send' for #<BasicObject:0x000000010493faf8>
```

.. BasicObject doesn't have `is_a?`, so [this line](https://github.com/rspec/rspec-expectations/pull/1455/files#diff-d3b3086ce048c1abcc8e9452f21c7afa019704480ba45fcf09c98819bcccdbc4R53) breaks for those. Good test, I'd have never thought of that!

I'm not sure how to safely detect that I'm _dealing_ with a BasicObject though, since it doesn't have `respond_to?` or `class` methods (or anything else I can think of). Incidentally, it also doesn't have a `send` method, so this bit [here](https://github.com/rspec/rspec-expectations/blob/main/spec/rspec/expectations/extensions/kernel_spec.rb#L37-L39) just produces a `NoMethodError: undefined method `send' for #<BasicObject:0x000000012ca06360>`, which seems probably unintended?

Any advice? I could just catch NoMethodError on `really_responds_to?` and _assume_ it means we're on a BasicObject (and therefore not a spy). I'll push that up, but I feel like there _ought_ to be a better way to do this - possibly behavior to add in the Proxy itself?

nevinera:
Oh, and please let me know if I'm using PRs wrong here - I've run into a variety of expectations regarding who resolves conversations, how much feedback to ask for, etc. You guys seem super-active, so I'm hoping I can be helpful.

pirj:
No worries on the code, it feels that we’re getting along quite good. Thanks for your tireless contributions!

nevinera:
Darn, looks like I still have some kind of problem on ruby-1.8.7. I've never actually gotten it to install on an m1, guess I'll spend some time on that today. It's obviously going to keep coming up :-)

lib/rspec/matchers/built_in/has.rb
```diff
@@ -46,8 +46,22 @@ def description
 
       private
 
+        # Catch a semi-frequent typo - if you have strict_predicate_matchers disabled and
+        # expect(spy).to have_receieveddd(:foo) it would be evergreen - the dynamic matcher
+        # queries `has_receiveddd?`, the spy _fakes_ the method, returning its (truthy) self.
+        def really_responds_to?(method)
+          if defined?(RSpec::Mocks::Double) && @actual.is_a?(RSpec::Mocks::Double)
```
pirj, line 53: Cosmetic. This can be optimized by extracting the defined? check outside of the method, and defining the method conditionally.
I hint at adding a ruby version check there. Don’t want to drown your time. I would wholeheartedly accept that we don’t fix that issue for users sticking to 1.8.7

lib/rspec/matchers/built_in/has.rb
```diff
@@ -46,8 +46,22 @@ def description
...
+          # Proxied BasicObjects don't have `is_a?`, or basically anything else.
+          @actual.respond_to?(method)
+        end
+
         def predicate_accessible?
-          @actual.respond_to? predicate
+          really_responds_to?(predicate)
         end
 
         # support 1.8.7, evaluate once at load time for performance
```
pirj, line 67: The line below can suggest the solution. Were object.methods a list if stings back in the 1.8.7?

lib/rspec/matchers/built_in/has.rb
nevinera: That was my suspicion, but I didn't want to waste anyone's time by checking through repeated CI runs I couldn't start. Either way, I'll need _some_ way to run the specs locally on 1.8.7 - if I don't sort it by this afternoon I'll push up a blind fix though.

pirj: Ci would start now after you have some contributions merged.

nevinera: Ah! Okay, I'll work that way while I thrash my way through docker-based development on the side then. Ignore me until I get these suites sufficiently green -.-

lib/rspec/matchers/built_in/has.rb
nevinera: Yup! I had no idea - I was only a young impressionable rubyist in those days, and did not use such magical methods :-D

https://rubydoc.info/stdlib/core/1.8.7/Object:methods

lib/rspec/matchers/built_in/has.rb
nevinera: I believe that resolved it, though there's a _segfault_ on one of the legacy builds. Is that just something that happens occasionally? (I rewrote and force-pushed the last commit to prompt a rebuild; I didn't see a cleaner way to do that.)

pirj: This happens time to time.

lib/rspec/matchers/built_in/has.rb
```diff
@@ -46,8 +46,22 @@ def description
...
+        def really_responds_to?(method)
+          if defined?(RSpec::Mocks::Double) && @actual.is_a?(RSpec::Mocks::Double)
+            @actual.respond_to?(method) && methods_include?(method)
+          else
+            @actual.respond_to?(method)
+          end
+        rescue NoMethodError
+          # Proxied BasicObjects don't have `is_a?`, or basically anything else.
+          @actual.respond_to?(method)
+        end
```
JonRowe, line 61: We typically define conditional behaviour from existence / feature support by defining the method conditionally, also we should probably use the equality matcher here in case `is_a?` is tampered with, which would probably solve the basic object problem too.
```suggestion
        if defined?(RSpec::Mocks::Double)
          def really_responds_to?(method)
            if RSpec::Mocks::Doubl === @actual
              @actual.respond_to?(method) && methods_include?(method)
            else
              @actual.respond_to?(method)
            end
          rescue NoMethodError
            # Proxied BasicObjects don't have `is_a?`, or basically anything else.
            @actual.respond_to?(method)
          end
        else
          def really_responds_to?(method)
            @actual.respond_to?(method)
          end
        end
```

lib/rspec/matchers/built_in/has.rb
```diff
@@ -56,11 +70,19 @@ def predicate_accessible?
...
+          end
           # :nocov:
         else
           def private_predicate?
             @actual.private_methods.include? predicate
           end
+
+          def methods_include?(method)
+            @actual.methods.include?(method)
+          end
```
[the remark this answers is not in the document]
nevinera: One version of this method calls `to_s` on the method name, because the list of methods returned used to be strings.

We _could_ do that check inline (or look for either `method` _or_ `method.to_s`), it seemed in-line with the prior decision about `private_predicate` to just conditionally define the `methods_include?` method to handle that case - are you sure you want me to collapse that to an inline check? It'd look like one of these, and would be less performant:

```
@actual.methods.include?(method) || @actual.methods.include?(method.to_s)
@actual.methods.include?(String === methods.first ? method.to_s : method)
```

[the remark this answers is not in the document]
JonRowe: Oh my bad, this is fine as is then.

lib/rspec/matchers/built_in/has.rb
nevinera: Huh, BasicObject doesn't even have `respond_to?`, which makes that question kind of moot. `===` still works, so your solution is much more elegant, thanks :-)

JonRowe (approved):

pirj:
Thank you!

JonRowe:
Released in 3.13.2

PR #1456, opened 2024-05-08
Give correct error message for `expect {}.not_to raise_error(/foo/)`

nevinera:
Found this while adding coverage - there was a test that _ought_ to have covered the line in question, but didn't because it was actually slightly too loose. Tightening it up shows that it's getting the wrong error message, because the RaiseError matcher is handling `expected_error_or_message` by checking the type, and it's looking for _String_, but not Regexp there.

Figured it'd make more sense as its own PR, since it's not just test coverage. (The line _is_ covered afterwards though!)

Failure with just the test change:
```
  1) expect { ... }.not_to raise_error(/message/) issues a warning
     Failure/Error: expect { raise 'blarg' }.not_to raise_error(/blah/)

       Kernel received :warn with unexpected arguments
         expected: (match /not_to raise_error\(message\)` risks false positives/ and match /Called from \/Users\/ericmueller\/src\/rspec-expectations\/spec\/rspec\/matchers\/built_in\/raise_error_spec.rb:309/)
              got: ("WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives, since lite...ons/spec/rspec/matchers/built_in/raise_error_spec.rb:309:in `block (2 levels) in <top (required)>'.")
     # ./spec/rspec/matchers/built_in/raise_error_spec.rb:309:in `block (2 levels) in <top (required)>'
```

### DIFF
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -25,7 +25,7 @@ module RSpec
           when nil, UndefinedValue
             @expected_error = Exception
             @expected_message = expected_message
-          when String
+          when String, Regexp
             @expected_error = Exception
             @expected_message = expected_error_or_message
           else
--- a/lib/rspec/matchers/built_in/raise_error.rb
+++ b/lib/rspec/matchers/built_in/raise_error.rb
@@ -13,6 +13,10 @@ module RSpec
         # argument. We can't use `nil` for that because we need to warn when `nil` is
         # passed in a different way. It's an Object, not a Module, since Module's `===`
         # does not evaluate to true when compared to itself.
+        #
+        # Note; this _is_ the default value supplied for expected_error_or_message, but
+        # because there are two method-calls involved, that default is actually supplied
+        # in the definition of the _matcher_ method, `RSpec::Matchers#raise_error`
         UndefinedValue = Object.new.freeze
 
         def initialize(expected_error_or_message, expected_message, &block)
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -291,25 +291,27 @@ RSpec.describe "expect { ... }.to raise_error.with_message(message)" do
 end
 
 RSpec.describe "expect { ... }.not_to raise_error('message')" do
-  it "issues a warning" do
+  it "issues a warning when configured to do so", :warn_about_potential_false_positives do
+    RSpec::Expectations.configuration.warn_about_potential_false_positives = true
     expect_warning_with_call_site __FILE__, __LINE__+1, /not_to raise_error\(message\)` risks false positives/
-    expect { raise 'blarg' }.not_to raise_error("blah")
+    expect { raise 'blarg' }.not_to raise_error('blah')
   end
 
-  it "can supresses the warning when configured to do so", :warn_about_potential_false_positives do
+  it "supresses the warning when configured to do so", :warn_about_potential_false_positives do
     RSpec::Expectations.configuration.warn_about_potential_false_positives = false
     expect_no_warnings
-    expect { raise 'blarg' }.not_to raise_error("blah")
+    expect { raise 'blarg' }.not_to raise_error('blah')
   end
 end
 
 RSpec.describe "expect { ... }.not_to raise_error(/message/)" do
-  it "issues a warning" do
+  it "issues a warning when configured to do so", :warn_about_potential_false_positives do
+    RSpec::Expectations.configuration.warn_about_potential_false_positives = true
     expect_warning_with_call_site __FILE__, __LINE__+1, /not_to raise_error\(message\)` risks false positives/
     expect { raise 'blarg' }.not_to raise_error(/blah/)
   end
 
-  it "can supresses the warning when configured to do so", :warn_about_potential_false_positives do
+  it "supresses the warning when configured to do so", :warn_about_potential_false_positives do
     RSpec::Expectations.configuration.warn_about_potential_false_positives = false
     expect_no_warnings
     expect { raise 'blarg' }.not_to raise_error(/blah/)
--- a/spec/rspec/matchers/built_in/raise_error_spec.rb
+++ b/spec/rspec/matchers/built_in/raise_error_spec.rb
@@ -290,9 +290,22 @@ RSpec.describe "expect { ... }.to raise_error.with_message(message)" do
   end
 end
 
-RSpec.describe "expect { ... }.not_to raise_error(message)" do
+RSpec.describe "expect { ... }.not_to raise_error('message')" do
   it "issues a warning" do
-    expect_warning_with_call_site __FILE__, __LINE__+1, /risks false positives/
+    expect_warning_with_call_site __FILE__, __LINE__+1, /not_to raise_error\(message\)` risks false positives/
+    expect { raise 'blarg' }.not_to raise_error("blah")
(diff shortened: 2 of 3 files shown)


spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -290,9 +290,22 @@ def to_s
   end
 end
 
-RSpec.describe "expect { ... }.not_to raise_error(message)" do
+RSpec.describe "expect { ... }.not_to raise_error('message')" do
   it "issues a warning" do
-    expect_warning_with_call_site __FILE__, __LINE__+1, /risks false positives/
+    expect_warning_with_call_site __FILE__, __LINE__+1, /not_to raise_error\(message\)` risks false positives/
+    expect { raise 'blarg' }.not_to raise_error("blah")
```
pirj, line 296: Just nitpicking over the use of single and double quotes on the same line.

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -290,9 +290,22 @@ def to_s
...
+  end
+
+  it "can supresses the warning when configured to do so", :warn_about_potential_false_positives do
```
pirj, line 299: Nitpicking - can or suppresses?

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -290,9 +290,22 @@ def to_s
...
+    RSpec::Expectations.configuration.warn_about_potential_false_positives = false
```
pirj, line 300: We have an helper ‘with_isolated_config’. If you think it’s overkill, or suspect that the example couldn’t be considered as balanced with an opposite example that doesn’t use it, I think it’s fine to restore it after ‘expect’.

spec/rspec/matchers/built_in/raise_error_spec.rb
pirj, line 299: Wondering why ‘warn_about_potential_false_positives’ metadata is needed here to warn, but it warns anyway in other examples? Is it to restore the setting to the default? It just reads confusingly when used this way.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -25,7 +25,7 @@ def initialize(expected_error_or_message, expected_message, &block)
           when nil, UndefinedValue
```
pirj, line 25: This usage of UndefinedValue confuses me. It was supposed be the default value in the method signature.

spec/rspec/matchers/built_in/raise_error_spec.rb
pirj: I see that we do the same in other examples in this file.

lib/rspec/matchers/built_in/raise_error.rb
```diff
@@ -25,7 +25,7 @@ def initialize(expected_error_or_message, expected_message, &block)
...
             @expected_error = Exception
             @expected_message = expected_message
-          when String
+          when String, Regexp
```
pirj, line 28: Is there any change in behaviour?
Is looks like that previously a regexp would have been used as an exception?

This change indeed seems the right thing to do. I’m wondering, why it worked the same way before?

Is it because we still were tolerant to whatever it is an exception or something matcheable with a string message?

```ruby
if values_match?(@expected_error, @actual_error) ||
               values_match?(@expected_error, actual_error_message)
```

pirj (approved):
No objections on the changes, just a few notes to spark a discussion on if this can be further improved. Partially caused by the lack of morning coffee.

spec/rspec/matchers/built_in/raise_error_spec.rb
nevinera: Whoops! I'm used to doubles, but can adapt either way. I must have copied the local style for the first string and missed it on the second one -.-

spec/rspec/matchers/built_in/raise_error_spec.rb
nevinera: I didn't know about this until you asked (and I wasn't sure until I tested it out), but adding a shared context like [this one](https://github.com/rspec/rspec-expectations/blob/main/spec/spec_helper.rb#L120-L124) (with a tag specified) causes that context to be automatically included in any context that supplies that same tag. So including this metadata causes an automatic cleanup to occur. I, of course, had just copied it from other tests without thinking :-)

spec/rspec/matchers/built_in/raise_error_spec.rb
nevinera: Ah, it's probably more accurate to explicitly enable it in the other test, I agree (if I'm following you?)

nevinera: (updated the existing test I copied from at the same time)

lib/rspec/matchers/built_in/raise_error.rb
nevinera: That looks like either another bug or dead code at this point, good eye! I'll go spelunking in a moment.

lib/rspec/matchers/built_in/raise_error.rb
nevinera: So.. there is a change in "behavior", but it pretty much just affects what failure messages come out. These two ivars are really only used for that purpose, and aren't exposed externally, thankfully; the only impact I can find is that a slightly wrong failure message gets generated (it acts as though you supplied an exception if you supplied only a regex). Could have had larger impact, but the main place we look at the `@expected_error`, we only check `@expected_error != Exception`, which is not true for Regexps either (though I suppose if you supplied `to raise_error(Exception, /foo/)`, you might also be able to get some wrong behavior?).

And the reason the specs didn't mind was just that the case checking this had a test that only checked the part of the failure message the outcomes had in common ("risks false positives"), which was still being produced.

lib/rspec/matchers/built_in/raise_error.rb
nevinera: Ah, it's not _wrong_, just referenced from afar: https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers.rb#L773

I think this will confuse future-us less with an explanatory comment, I'll add that here.

pirj: Right, now I recall it. Good find.

spec/rspec/matchers/built_in/raise_error_spec.rb
```diff
@@ -290,13 +290,28 @@ def to_s
...
+    expect_no_warnings
+    expect { raise 'blarg' }.not_to raise_error('blah')
+  end
+end
+
+RSpec.describe "expect { ... }.not_to raise_error(/message/)" do
+  it "issues a warning when configured to do so", :warn_about_potential_false_positives do
+    RSpec::Expectations.configuration.warn_about_potential_false_positives = true
+    expect_warning_with_call_site __FILE__, __LINE__+1, /not_to raise_error\(message\)` risks false positives/
```
pirj, line 310: Just to dissect it. Before the change, this was printing a slightly different warning:
```ruby
Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives
```
and now it's a more appropriate for `not_to raise_error(/regexp/)`:
```ruby
Using `expect { }.not_to raise_error(message)` risks false positives
```

My apologies that it took that long to figure out.

pirj (approved):
Thank you!

PR #1457, opened 2024-05-08
Drop dead (private) methods from BaseMatcher

nevinera:
Drop the `assert_ivars` and `present_ivars` methods from RSpec::Matchers::BuiltIn::BaseMatcher`.

They aren't referenced anymore, and apparently haven't been for ~10 years. I confirmed with git pickaxes, but then I also checked out a few random points in time over that period and looked for either string in the repository - no references in 2015, or at any point I looked since then.

It _is_ possible that (against your strong recommendations), people are subclassing this class for their custom matchers, but I think that's their problem. And this isn't a particularly obvious or natural method for them to be using anyway.

### DIFF
--- a/lib/rspec/matchers/built_in/base_matcher.rb
+++ b/lib/rspec/matchers/built_in/base_matcher.rb
@@ -126,22 +126,6 @@ module RSpec
 
       private
 
-        def assert_ivars(*expected_ivars)
-          return unless (expected_ivars - present_ivars).any?
-          ivar_list = EnglishPhrasing.list(expected_ivars)
-          raise "#{self.class.name} needs to supply#{ivar_list}"
-        end
-
-        if RUBY_VERSION.to_f < 1.9
-          # :nocov:
-          def present_ivars
-            instance_variables.map(&:to_sym)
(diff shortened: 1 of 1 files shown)


lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -126,22 +126,6 @@ def self.underscore(camel_cased_word)
 
       private
```
pirj, line 127:
```suggestion
```

A agree with RuboCop here, there are no method definitions below (anymore), and this access modifier is useless and can be removed.

My apologies, I’m from mobile and couldn’t mark two lines, the suggestion won’t apply clearly.

pirj (approved):
Thank you!
100% coverage already yields results

lib/rspec/matchers/built_in/base_matcher.rb
```diff
@@ -126,22 +126,6 @@ def self.underscore(camel_cased_word)
...
```
JonRowe, line 128:
```suggestion
```

PR #1458, opened 2024-05-08
100% test coverage

nevinera:
Well, aside from a few lines in two other PRs (https://github.com/rspec/rspec-expectations/pull/1457 and https://github.com/rspec/rspec-expectations/pull/1456) that needed non-spec changes.

First, I found and marked all of the segments that seem to be ruby-version specific, and set them off with `:nocov:` tags. That's in the first commit. Then I started tackling the remaining uncovered lines one at a time, adding tests  I believe there to be no non-spec, non-comment changes in this PR, but I'm confident that some of the specs can be done more elegantly. Please, enhance my approaches, rspec wizards :-D

### DIFF
--- a/spec/rspec/matchers/built_in/compound_spec.rb
+++ b/spec/rspec/matchers/built_in/compound_spec.rb
@@ -260,6 +260,26 @@ module RSpec::Matchers::BuiltIn
       end
     end
 
+    describe "expect(...).to custom_matcher.and(other_matcher)" do
+      it "treats a matcher that doesn't support value expectations correctly" do
+        expect {
+          expect([1, 2]).to include(1).and raise_error(/test/)
+        }.to fail_with(/but was not given a block/)
+      end
+
+      it "treats a matcher that does support value expectations correctly" do
+        expect([1, 2]).to include(1).and include(2)
+      end
+
+      it "treats a matcher that doesn't _specify_ as supporting value expectations" do
+        unsupporting_includes_class = Class.new(Object) do
+          def matches?(actual); actual.include?(2); end
+        end
+        unsupporting_includes = unsupporting_includes_class.new
+        expect([1, 2]).to include(1).and unsupporting_includes
+      end
+    end
+
     describe "expect(...).to matcher.and(other_matcher)" do
 
       it_behaves_like "an RSpec value matcher", :valid_value => 3, :invalid_value => 4, :disallows_negation => true do
--- a/spec/rspec/matchers/built_in/compound_spec.rb
+++ b/spec/rspec/matchers/built_in/compound_spec.rb
@@ -993,5 +993,30 @@ module RSpec::Matchers::BuiltIn
         }.to raise_error(NotImplementedError, /matcher.or matcher` is not supported/)
       end
     end
+
+    describe "#expects_call_stack_jump?" do
+      subject(:expects_call_stack_jump?) { matcher.expects_call_stack_jump? }
+      let(:matcher) { described_class.new(matcher_1, matcher_2) }
+      let(:matcher_1) { double("Matcher 1") }
+      let(:matcher_2) { double("Matcher 2") }
+
+      context "when neither matcher expects a call-stack jump" do
+        let(:matcher_1) { double("Matcher 1", :expects_call_stack_jump? => false) }
+        let(:matcher_2) { double("Matcher 2", :expects_call_stack_jump? => false) }
+        it { is_expected.to be_falsey }
+      end
+
+      context "when one of the matchers expects a call-stack jump" do
+        let(:matcher_1) { double("Matcher 1", :expects_call_stack_jump? => false) }
+        let(:matcher_2) { double("Matcher 2", :expects_call_stack_jump? => true) }
+        it { is_expected.to be_truthy }
+      end
+
+      context "when neither matcher _specifies_ that it expects a call-stack jump" do
+        let(:matcher_1) { Object.new }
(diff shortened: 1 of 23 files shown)


spec/rspec/matchers/built_in/compound_spec.rb
```diff
@@ -973,5 +993,30 @@ def actual
         }.to raise_error(NotImplementedError, /matcher.or matcher` is not supported/)
       end
     end
+
+    describe "#expects_call_stack_jump?" do
```
nevinera, line 997: This one I was least satisfied with - you've fairly consistently avoided thin unit testing in favor of behavioral/usage-flavored tests. But the tests I rigged up for that here were _enormous_, and I decided that they were complex enough to get in the way (I needed to make custom ruby-class block-accepting matchers that expected and didn't expect call-stack jumps). But I'll take another swing if I'm not imagining that preference, and you think it's worth the ugly.

spec/rspec/matchers/composable_spec.rb
```diff
@@ -12,6 +12,13 @@ module Matchers
         }.to fail_with(STDOUT.inspect)
       end
 
+      it "behaves correctly when surfacing descriptions from a readable IO object" do
+        allow(STDOUT).to receive(:each).and_yield
+        expect {
+          expect(3).to matcher_using_surface_descriptions_in(STDOUT)
+        }.to fail_with(STDOUT.to_s)
+      end
```
nevinera, line 20: This one was also ugly, because we can't get to the line in question without an instance of IO, and I didn't figure we'd want to open a fixture File for this spec (but if we do, that'd make it nicer).

pirj: Would StringIO work?

nevinera: My first thought as well, but it's actually not a subclass of IO, just duck-typed like it! I assumed so hard I wrote the test and debugged it to find out I was wrong -.-

nevinera:
The oldest rubies are failing two of the new tests:

```
  1) operator matchers RSpec::Matchers::BuiltIn::NegativeOperatorMatcher complains when negated
     Failure/Error: expect {
       expected /does not support `should_not != .*Use `should ==/ but nothing was raised
     # ./spec/rspec/matchers/built_in/operators_spec.rb:264

  2) operator matchers RSpec::Matchers::BuiltIn::PositiveOperatorMatcher complains when negated
     Failure/Error: expect {
       expected /does not support `should != .*Use `should_not ==/ but nothing was raised
     # ./spec/rspec/matchers/built_in/operators_spec.rb:241
```

I don't see anything _obviously_ version dependent here, the only thing that seems plausible is this `has_non_generic_implementation_of?` [method](https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers/built_in/operators.rb#L82-L86), but I don't see _how_ variation there would cause these expectations to skip failing entirely..

I think this might mean those matchers are allowed and broken in 1.8.7? Which would probably be .. well, I doubt anybody on that version expects first-class support, but once I get 1.8.7 working in a container I can probably fix it at least :-)

pirj (approved):
Looks good. Thank you!

nevinera:
@pirj - The other two PRs are merged now, so I rebased this on top of `main`. That means that it _actually_ has 100% coverage now - would you like me to bump [this value](https://github.com/rspec/rspec-expectations/blob/main/spec/spec_helper.rb#L5) up to 100% to keep it there, or leave it at 92% like it is now? I wasn't originally planning on doing so, but I noticed that we were actually trying to enforce a 100% value in rspec-core using such a config, so I thought I'd check.

pirj:
Let’s tighten it up to 100%, we can always loosen up if there’s something that we can’t reasonably cover.

nevinera:
@JonRowe, this one has been waiting for a while, and ought to be fairly safe as it's test-only. I've updated it with latest if you have a minute to comment; apparently *I* was the only one to introduce uncovered code since the last push 6 weeks ago >.<

spec/rspec/matchers/built_in/operators_spec.rb
```diff
@@ -228,22 +228,48 @@ def method_missing(name, *args, &block)
...
+
     it "works when the target has implemented #send" do
-      o = Object.new
-      def o.send(*_args); raise "DOH! Library developers shouldn't use #send!" end
       expect {
         o.should == o
       }.not_to raise_error
     end
+
+    it "complains when negated", :if => RUBY_VERSION.to_f >= 1.9 do
```
pirj, line 240: We'll be [removing](https://github.com/rspec/rspec-core/issues/1821) :if/:else in the next major.
```suggestion
    it "complains when negated", :skip => RUBY_VERSION.to_f < 1.9 do
```

spec/rspec/matchers/built_in/operators_spec.rb
```diff
@@ -228,22 +228,48 @@ def method_missing(name, *args, &block)
...
+
     it "works when the target has implemented #send" do
-      o = Object.new
-      def o.send(*_args); raise "DOH! Library developers shouldn't use #send!" end
       expect {
         o.should_not == :foo
       }.not_to raise_error
     end
+
+    it "complains when negated", :if => RUBY_VERSION.to_f >= 1.9 do
```
pirj, line 263:
```suggestion
    it "complains when negated", :skip => RUBY_VERSION.to_f < 1.9 do
```

pirj (approved):

pirj:
Thank you!

PR #1459, opened 2024-05-16
output matcher: add `as_tty` method to make the simulated stream behave as a TTY

porras:
## The problem

I have an app which prints some output only if the standard output is a TTY. I'd say it's a relative standard practice:

```ruby
puts "Whatever" if $stdout.tty?
```

Currently, this can't be tested with `expect { ... }.to output(...).to_stdout` because `$stdout` is replaced with a `StringIO`, which always returns `false` for `.tty?`, so the code never prints.

If this is a problem RSpec wants to fix (I guess you can argue that it's an antipattern and should be implemented in a different way), this PR does it.

## The fix

~`$stdout` is still replaced with a `StringIO` and the matcher works the same, but some specific methods are called on the original stream. I only added `.tty?` because that's my usecase, but I guess there are other methods for which this could make sense (now or in the future).~

~I haven't contributed to RSpec before so I'm not familiar with coding standards, style, etc., so I appreciate any feedback, particularly on the following problem: I had to add an exception for the libraries allowed to be loaded for `delegate` (in fact there is another one for `stringio` for the same reason). Is this acceptable? If not, there are at least two alternatives I could implement:~

* ~Do some trick to require `delegate` and define the class on runtime only once `output` has been called. This is already  done for the `tempfile` require, although it would be a bit trickier than that because the class needs to be defined to.~
* ~Not use `SimpleDelegator` and implement the behavior using `method_missing`.~

~Is any of these approaches preferred?~

Update: the description above no longer mathes the implementation, which I've changed according to the feedback received. It's simpler now and doesn't need such a long explanation :sweat_smile:

### DIFF
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -69,6 +69,17 @@ module RSpec
           self
         end
 
+        # @api public
+        # Tells the matcher to simulate the output stream not being a TTY.
+        # Note that that's the default behaviour if you don't call `as_tty`
+        # (since `StringIO` is not a TTY).
+        def as_not_tty
+          raise ArgumentError, '`as_not_tty` can only be used after `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
+
+          @stream_capturer.as_tty = false
+          self
+        end
+
         # @api private
         # @return [String]
         def failure_message
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -150,7 +150,8 @@ module RSpec
         attr_accessor :as_tty
 
         def tty?
-          as_tty || super
+          return super if as_tty.nil?
+          as_tty
         end
       end
 
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -63,7 +63,7 @@ module RSpec
         # Tells the matcher to simulate the output stream being a TTY.
         # This is useful to test code like `puts '...' if $stdout.tty?`.
         def as_tty
-          raise '`as_tty` can only be used with `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
+          raise ArgumentError, '`as_tty` can only be used after `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
 
           @stream_capturer.as_tty = true
           self
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -29,7 +29,7 @@ module RSpec
         # Tells the matcher to match against stdout.
         # Works only when the main Ruby process prints to stdout
         def to_stdout
-          @stream_capturer = CaptureStdout
+          @stream_capturer = CaptureStdout.new
           self
         end
 
@@ -37,7 +37,7 @@ module RSpec
         # Tells the matcher to match against stderr.
         # Works only when the main Ruby process prints to stderr
         def to_stderr
-          @stream_capturer = CaptureStderr
+          @stream_capturer = CaptureStderr.new
           self
         end
 
@@ -56,6 +56,16 @@ module RSpec
         # This is significantly (~30x) slower than `to_stderr`
         def to_stderr_from_any_process
           @stream_capturer = CaptureStreamToTempfile.new("stderr", $stderr)
+          self
+        end
+
+        # @api public
+        # Tells the matcher to simulate the output stream being a TTY.
+        # This is useful to test code like `puts '...' if $stdout.tty?`.
+        def as_tty
+          raise '`as_tty` can only be used with `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
+
+          @stream_capturer.as_tty = true
           self
         end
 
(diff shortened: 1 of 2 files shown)


spec/rspec/matchers/built_in/output_spec.rb
```diff
@@ -159,6 +159,14 @@ def print_to_stream(msg)
       }
     end
```
pirj, line 161: Side observation - let(:macher) on line 6 of this spec seems unnecessary as it’s overridden by a later define_method.

May i kindly ask you to try and remove it?

lib/rspec/matchers/built_in/output.rb
```diff
@@ -1,4 +1,5 @@
 require 'stringio'
+require 'delegate'
```
JonRowe, line 2: Please don't add this.

spec/rspec/expectations_spec.rb
```diff
@@ -3,6 +3,7 @@
 RSpec.describe "RSpec::Expectations" do
   allowed_loaded_features = [
     /stringio/, # Used by `output` matcher. Can't be easily avoided.
+    /delegate/, # Same.
```
JonRowe, line 6: Please don't add this it can be avoided.

lib/rspec/matchers/built_in/output.rb
```diff
@@ -135,23 +136,39 @@ def self.capture(_block)
         end
       end
 
+      # @private
+      # Initializes a new StringIO stream and delegates everything to it, except
+      # a couple of specific methods (e.g. `tty?`), which get delegated to the
+      # original stream.
+      class CapturedStream < SimpleDelegator
```
JonRowe, line 143: Instead inherit from `StringIO`

JonRowe (changes requested):

JonRowe:
I think its reasonable to respect the original `tty` behaviour as we're just capturing the output, another option would be to explicitly define this with something like `expect { ... }.to output(...).to_stdout.as_tty` which sets it explicitly in the test

spec/rspec/matchers/built_in/output_spec.rb
porras: mmmm 10 specs did actually fail when I removed it, I don't really understand why. I'll leave that change out.

porras:
What I just pushed is without implementing the `.as_tty` method, it just hardcodes it to true. I'll give it a try to it.

porras:
I just pushed an attempt at implementing `as_tty`. It works but it's admittedly ugly. Feedback on suggested refactoring or a different way to do it welcome :smile: I guess the ugliest thing is the duplication between `CaptureStdout` and `CaptureStderr` (which was already there, but it's more evident now), but it's difficult to refactor them together without resorting to some metaprogramming tricks, because they need to reassign the global variables.

pirj (approved):
Nice! Thank you.

Can you add a couple of examples covering those parts where it raises when as_tty is not compatible with “in any process”?

‘output’ calls can be chained, too? output('foo').to_stdout.and output('bar').to_stdout.as_tty. Will it work as expected? Should as_tty be specified for both?
I am sorry, Iam from my phone, does it even chain properly, or just my wishful thinking?

lib/rspec/matchers/built_in/output.rb
```diff
@@ -59,6 +59,16 @@ def to_stderr_from_any_process
           self
         end
 
+        # @api public
+        # Tells the matcher to simulate the output stream being a TTY.
+        # This is useful to test code like `puts '...' if $stdout.tty?`.
+        def as_tty
+          raise '`as_tty` can only be used with `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
```
JonRowe, line 66:
```suggestion
          raise ArgumentError, '`as_tty` can only be used after `to_stdout` or `to_stderr`' unless @stream_capturer.respond_to?(:as_tty=)
```

lib/rspec/matchers/built_in/output.rb
```diff
@@ -136,13 +146,25 @@ def self.capture(_block)
       end
 
       # @private
-      module CaptureStdout
-        def self.name
+      class CapturedStream < StringIO
+        attr_accessor :as_tty
+
+        def tty?
+          as_tty || super
```
JonRowe, line 153: Forcing this to false should also be possible.

```suggestion
          return super if as_tty.nil?
          as_tty
```

spec/rspec/matchers/built_in/output_spec.rb
```diff
@@ -214,5 +214,37 @@ def print_to_stream(msg)
...
+      end
+
+      it "captures output written to a TTY through stderr" do
+        expect { $stderr.print "foo" if $stderr.tty? }.to output("foo").to_stderr.as_tty
+      end
+
+      it "errors out nicely when attempting it without having set the stream" do
+        expect {
+          expect { print "foo" }.to output("foo").as_tty # <- wrong call, `to_std(out|err)` is missing
+        }.to raise_error(/can only be used with `to_stdout` or `to_stderr`/)
```
JonRowe, line 234:
```suggestion
        }.to raise_error(/can only be used after `to_stdout` or `to_stderr`/)
```

JonRowe (changes requested):
Nice, a few suggestions

lib/rspec/matchers/built_in/output.rb
JonRowe: I was suggestting this to leave a `as_not_tty` or `as_tty(false)` future [or could even be added here] to be explicit about what this setting is as `StringIO` is not under our control and could be monkey patched.

porras: Makes sense, thank you!

JonRowe:
Great! Thanks!

PR #1460, opened 2024-05-19
Support compound stream matchers

### LINKED ISSUE #89: Compound expectations of OutputMatcher do not receive captured output
### Compound expectations with OutputMatcher not receiving captured output
:heart: Thanks for any help on this in advance - it might just be user error.

When constructing a compound expectation using the `output` matcher, I'd expect something like this to pass:
```ruby
expect { puts "foo\nbar" }.to(output(/foo/).to_stdout & output(/bar/).to_stdout)
# or to use the `.and` style
expect { puts "foo\nbar" }.to output(/foo/).to_stdout
                          .and output(/bar/).to_stdout
```

Instead a failure is reported:
```
Failure/Error: expect { puts "foo\nbar" }.to(output(/foo/).to_stdout & output(/bar/).to_stdout)
       expected block to output /bar/ to stdout, but output nothing
```

Running them individually passes however:
```ruby
expect { puts "foo\nbar" }.to output(/foo/).to_stdout
expect { puts "foo\nbar" }.to output(/bar/).to_stdout
# => Passes
```

### Your environment
* Ruby version: 3.1.0
* rspec-expectations version: 3.12.0

### Steps to reproduce
```ruby
expect { puts "foo\nbar" }.to(output(/foo/).to_stdout & output(/bar/).to_stdout)
# or to use the `.and` style
expect { puts "foo\nbar" }.to output(/foo/).to_stdout
                          .and output(/bar/).to_stdout
```

### Expected behavior

I'd expect that compound statements would both use the same STDOUT output and match against that

### Actual behavior
```
Failure/Error: expect { puts "foo\nbar" }.to(output(/foo/).to_stdout & output(/bar/).to_stdout)
       expected block to output /bar/ to stdout, but output nothing
```

nevinera:
Currently, compound stream matchers aren't properly supported, because when the CompoundMatcher nests them, the inner matcher doesn't expose the captured output to the outer matcher. This results in rspec/rspec#89.

While resolving the issue for the `to_stderr_from_any_process` though, I found an additional problem, which is that this matcher didn't properly _restore_ the `RSpec::Support::StdErrSplitter` to its original state after capturing (it leaves it pointed at a now-dead Tempfile stream).

This PR will be in draft state until I work out a resolution for that.

Fixes rspec/rspec#89

### DIFF
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -186,7 +186,7 @@ module RSpec
           captured_stream.string
         ensure
           $stdout = original_stream
-          $stdout.write(captured_stream.string) unless $stdout == STDOUT
+          $stdout.write(captured_stream.string) unless $stdout == STDOUT # rubocop:disable Style/GlobalStdStream
         end
       end
 
@@ -210,7 +210,7 @@ module RSpec
           captured_stream.string
         ensure
           $stderr = original_stream
-          $stderr.write(captured_stream.string) unless $stderr == STDERR
+          $stderr.write(captured_stream.string) unless $stderr == STDERR # rubocop:disable Style/GlobalStdStream
         end
       end
 
@@ -225,20 +225,46 @@ module RSpec
           # thread, fileutils, etc), so it's worth delaying it until this point.
           require 'tempfile'
 
+          # This is.. awkward-looking. But it's written this way because of how
+          # compound matchers work - we essentially need to be able to tell if
+          # we're in an _inner_ matcher, so we can pass the stream-output along
+          # to the outer matcher for further evaluation in that case. Added to
+          # that, it's fairly difficult to _tell_, because the only actual state
+          # we have access to is the stream itself, and in the case of stderr,
+          # that stream is really a RSpec::Support::StdErrSplitter (which is why
+          # we're testing `is_a?(File)` in such an obnoxious way).
+          inner_matcher = stream.to_io.is_a?(File)
+
+          # Careful here - the StdErrSplitter is what is being cloned; we're
+          # relying on the implemented clone method of that class (in
+          # rspec-support) to actually clone the File for ensure-reopen.
           original_stream = stream.clone
+
           captured_stream = Tempfile.new(name)
 
           begin
             captured_stream.sync = true
             stream.reopen(captured_stream)
             block.call
-            captured_stream.rewind
-            captured_stream.read
+            read_contents(captured_stream)
           ensure
+            captured_content = inner_matcher ? read_contents(captured_stream) : nil
             stream.reopen(original_stream)
-            captured_stream.close
-            captured_stream.unlink
+            stream.write(captured_content) if captured_content
+            clean_up_tempfile(captured_stream)
           end
+        end
+
+        private
+
+        def read_contents(strm)
+          strm.rewind
+          strm.read
+        end
+
+        def clean_up_tempfile(tempfile)
+          tempfile.close
+          tempfile.unlink
         end
       end
     end
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -257,9 +257,9 @@ module RSpec
 
         private
 
-        def read_contents(strm)
-          strm.rewind
-          strm.read
+        def read_contents(captured_stream)
+          captured_stream.rewind
+          captured_stream.read
         end
 
         def clean_up_tempfile(tempfile)
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -186,6 +186,7 @@ module RSpec
           captured_stream.string
         ensure
           $stdout = original_stream
+          $stdout.write(captured_stream.string) unless $stdout == STDOUT
         end
       end
 
@@ -209,6 +210,7 @@ module RSpec
           captured_stream.string
         ensure
           $stderr = original_stream
+          $stderr.write(captured_stream.string) unless $stderr == STDERR
         end
       end
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ Enhancements:
 
 * Improve the IO emulation in the output capture matchers (`output(...).to_stdout` et al)
   by adding `as_tty` and `as_not_tty` to change the `tty?` flags. (Sergio Gil Pérez de la Manga, #1459)
+* Add support for compound output matchers. (Eric Mueller, #1460)
 
 ### 3.13.1 / 2024-06-13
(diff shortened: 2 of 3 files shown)


nevinera:
I think this is the solution: https://github.com/rspec/rspec-support/pull/598

nevinera:
There does still seem to be some issue in JRuby - any insight into what the problem might be? I saw a couple of comments that were jruby-specific, but it didn't seem like they indicated a lack of support for anything I'm using here.. I've never used jruby, I'll need to spend some time getting an install working I think -.-

pirj:
I can reproduce those two failures on JRuby 9.4.7.0 + OpenJDK 22, which were a breeze to install with rvm/homebrew.

The only difference I could find is:

image

But `to_io` makes them return `true` in both cases for `is_a?(File)`, even though both before and after `to_io` those are of a class `Tempfile`.

Is there anything else I can dig for you? What are your suspects?

Gemfile
```diff
@@ -16,6 +16,8 @@ branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
   end
 end
 
+gem "rspec-support", :git => "https://github.com/nevinera/rspec-support.git", :branch => "nev--override-stderr-splitter-clone"
```
pirj, line 19: This is not needed anymore when https://github.com/rspec/rspec-support/pull/598 is merged?

nevinera: I think that's correct, but I'll need a few minutes to refresh; I'll take a look later today. :-)

nevinera: Ah, so the resolution on the other PR was just to declare that the behavior would remain incorrect on jruby for now, and skip the relevant tests there? Should I update this PR to match? I didn't feel comfortable skipping support for such a significant ruby, but I admit I didn't make any progress on it, and haven't found time to come back to the problem in a while now :-\

nevinera: I am having a surprising amount of trouble successfully skipping these tests for the right containers. It sure seems like "JRuby 1.7 1.8 mode" has `RSpec::Support::Ruby.jruby?` returning true; do you know what's special about that setup? (I'll poke around again this evening)

pirj: > "JRuby 1.7 1.8 mode" has RSpec::Support::Ruby.jruby? returning true

I think this is as expected - it's JRuby, too.

pirj: >  the resolution on the other PR was just to declare that the behavior would remain incorrect on jruby for now

Please accept my apologies. I was wrong, and under the impression that the code behaves, and it's just that we can't write a test for JRuby because the Tempfile is implemented there differently.
But it surprised me that the code is affected, too.

Fixing this for other platforms, but not JRuby will certainly be an odd thing to do.

I'll need some time to have a closer look at this.

nevinera: No, I'm just failing to type correctly. I'm seeing that "JRuby 1.7 1.8 mode" has `RSpec::Support::Ruby.jruby?` NOT returning true; I tried to set up the two specs in question to skip on all jruby, and it worked for every container except this one :-\

> Please accept my apologies. I was wrong, and under the impression that the code behaves, and it's just that we can't write a test for JRuby because the Tempfile is implemented there differently.
> But it surprised me that the code is affected, too.

For what it's worth, the behavior jruby should stay broken in the same way that it is currently broken on all platforms. The change already merged in https://github.com/rspec/rspec-support/pull/598 actually fixes _that_ bug only on non-jruby platforms; I believe that's the reason.

pirj: According to [this](https://github.com/rspec/rspec-support/pull/598#issuecomment-2200779633):

> both versions of Tempfile should behave the same from an API perspective

Can you please ask again, but with a code example where `clone` doesn’t work as expected?

nevinera: I called it a "bug" because it does appear to be intentional, but it's been under discussion as a problem in MRI a few times too; as far as I can tell, this is the only place in all of ruby that an object can change its type after instantiation. I would generally say that rubyists should avoid being concerned with the type of the object in the first place; that's just the only place the information I needed to check (is the the original tempfile, or has it been reopened?) seems to be held.

I.. could maybe patch File in rspec-support to track that explicitly; how offensive would that be?

lib/rspec/matchers/built_in/output.rb
```diff
@@ -223,21 +225,47 @@ def capture(block)
...
 
+          # This is.. awkward-looking. But it's written this way because of how
+          # compound matchers work - we essentially need to be able to tell if
+          # we're in an _inner_ matcher, so we can pass the stream-output along
+          # to the outer matcher for further evaluation in that case. Added to
+          # that, it's fairly difficult to _tell_, because the only actual state
+          # we have access to is the stream itself, and in the case of stderr,
+          # that stream is really a RSpec::Support::StdErrSplitter (which is why
+          # we're testing `is_a?(File)` in such an obnoxious way).
+          inner_matcher = stream.to_io.is_a?(File)
```
JonRowe, line 236: Is there not a way to assert against one of our internal classes here? checking to see if we're already in a splitter etc

JonRowe (changes requested):

Gemfile
pirj: A wild idea - can we use a proxy that would additionally tell us if it’s a fake stream (and use a Tempfile underneath)?

lib/rspec/matchers/built_in/output.rb
nevinera: Only in some of the cases. We're interacting with whatever's in `$stderr` and `$stdout`, so I think we're reliably getting an internal class for the former but not the latter. Wrapping with an internal class for the latter as well did occur to me as a potential solution, but it was large and involved enough that I wasn't willing to attempt it for such a minor issue :-\

lib/rspec/matchers/built_in/output.rb
```diff
@@ -223,21 +225,47 @@ def capture(block)
...
+            clean_up_tempfile(captured_stream)
           end
         end
+
+        private
+
+        def read_contents(strm)
+          strm.rewind
+          strm.read
+        end
```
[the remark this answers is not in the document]
nevinera: It doesn't have to be `strm`, but it'll need to be something other than `stream` - that's taken, as it's one of the struct's [accessors](https://github.com/rspec/rspec-expectations/blob/829f7fec70764ff6850b9db3ea5f2a6c94ce2d82/lib/rspec/matchers/built_in/output.rb#L218).

I could perhaps call it `stream_being_read`, or something more verbose?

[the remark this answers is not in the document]
JonRowe: Fair point `captured_stream` then as thats what it is elsewhere

JonRowe:
I'm closing this as part of the monorepo migration because its a draft, but if you wanted to reopen it @nevinera it should be easier to work with there.

PR #1467, opened 2024-06-28
add example on how to use raise_error with block

glaucocustodio:
I had to do some digging on google to find that, having on the official docs might help some people :)


pirj:
Thank you! It's a very useful feature for those who use advanced exception features.

PR #1472, opened 2024-07-15
Use in built rspec warnings rather than kernel

### LINKED ISSUE #1461: Use `RSpec.warning` instead of `::Kernel.warn`
I expected this to raise an error when `Rspec.configure { |config| config.raise_on_warning = true }`

https://github.com/rspec/rspec-expectations/blob/0139c22a62624849c09dd33980554215e63ab9eb/lib/rspec/expectations/handler.rb#L7-L11

JonRowe:
Fixes #1461


JonRowe:
Released in 3.13.2

PR #1476, opened 2024-08-04
Prevent mismatched composed use of value and block matchers

pirj:
```diff
  expect { arr << :foo }
-   .to change { arr }.to be_one
+   .to change { arr }.to(be_one)
    .and change { arr }.to include(:foo)
```
This is to detect mismatches like this.


JonRowe:
Released in 3.13.2

PR #1477, opened 2024-08-04
Allow `satisfy` to match the block expectation return value

### LINKED ISSUE #91: `return_value` matcher
I have a proof of concept of a matcher that works with block expectations, but checks the value returned by the block.
Works as a compound matcher, too:

```ruby
$bar = 0
def foo
  $bar += 1
  2
end

RSpec.describe do
  it do
    $bar=0
    expect { foo }
      .to change { $bar }.to(1)
      .and return_value(2)
  end

  it do
    $bar=0
    expect { foo }
      .to return_value(satisfy { |x| x == 2 })
      .and change { $bar }.to(1)
  end
end
```

Existing known alternatives:
```ruby
expect {
  expect(foo).to eq(2)
}.to change { $bar }.to(1)
```

PS it took me [five years](https://github.com/rspec/rspec-expectations/issues/805#issuecomment-494380144)

Is this something worth continuing working on, WDYT @JonRowe ?

pirj:
fixes https://github.com/rspec/rspec/issues/91
idea https://github.com/rspec/rspec-expectations/issues/805#issuecomment-494380144

```ruby
ary = [1, 2]
expect { ary.shift }
  .to change { ary }.to([2])
  .and satisfy { |returned_value| returned_value == 1 }
```

### DIFF
--- a/lib/rspec/matchers/built_in/compound.rb
+++ b/lib/rspec/matchers/built_in/compound.rb
@@ -162,8 +162,13 @@ module RSpec
 
             inner, outer = order_block_matchers
 
+            returned = nil
             @match_results[outer] = outer.matches?(Proc.new do |*args|
-              @match_results[inner] = inner.matches?(inner_matcher_block(args))
+              p = Proc.new { |*args2|
+                returned = inner_matcher_block(args).call(*args2)
+              }
+              @match_results[inner] = inner.matches?(p)
+              returned
             end)
           end
 
--- a/spec/rspec/matchers/built_in/satisfy_spec.rb
+++ b/spec/rspec/matchers/built_in/satisfy_spec.rb
@@ -120,3 +120,54 @@ RSpec.describe "expect(...).not_to satisfy { block }" do
     end
   end
 end
+
+RSpec.describe "expect { ... }.to satisfy { block }" do
+  it_behaves_like "an RSpec block matcher", :disallows_negation => true, :skip_deprecation_check => true do
+    let(:matcher) { satisfy { |v| v == 1 } }
+    before { @k = 0 }
+    def valid_block
+      1
+    end
+    def invalid_block
+      2
+    end
+  end
+
+  let(:ary) { [1, 2] }
+
+  it "matches the returned value" do
+    expect { ary.shift }.to satisfy { |returned_value| returned_value == 1 }
+  end
+
+  it "provides a sensible failure message", :skip => !RSpec::Support::RubyFeatures.ripper_supported? do
+    expect {
+      expect { ary.shift }.to satisfy { |returned_value| returned_value == :other }
+    }.to fail_with("expected 1 to satisfy expression `returned_value == :other`")
+  end
+
+  context "when negated" do
+    it "passes when the returned value doesn't match" do
+      expect { ary.shift }.not_to satisfy { |returned_value| returned_value == 2 }
+    end
+
+    it "fails when the retuned value matches", :skip => !RSpec::Support::RubyFeatures.ripper_supported? do
+      expect {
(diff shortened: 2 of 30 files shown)


pirj:
cc @zverok WDYT?

spec/rspec/matchers/built_in/satisfy_spec.rb
```diff
@@ -120,3 +120,54 @@
...
+    it "works as a root matcher" do
+      expect { ary.shift }.to satisfy { |returned_value| returned_value == 1 }.and change { ary }.to([2])
+    end
+
+    it "works as a supplemental matcher" do
+      expect { ary.shift }.to change { ary }.to([2]).and satisfy { |returned_value| returned_value == 1 }
+    end
+  end
+
+  pending "allows a matcher as an argument" do
```
JonRowe, line 170: I'd rather not introduce pending specs unless its for something version specific, what needs to be done to make this work?

spec/rspec/matchers/built_in/satisfy_spec.rb
```diff
@@ -120,3 +120,54 @@
...
+    end
+
+    it "fails when the retuned value matches", :skip => !RSpec::Support::RubyFeatures.ripper_supported? do
+      expect {
+        expect { ary.shift }.not_to satisfy { |returned_value| returned_value == 1 }
+      }.to fail_with("expected 1 not to satisfy expression `returned_value == 1`")
+    end
+  end
+
+  describe "composed usage" do
```
JonRowe, line 160: This should follow on from the example group description
```suggestion
  describe "when composed with other matchers" do
```

lib/rspec/matchers/built_in/compound.rb
```diff
@@ -162,8 +162,13 @@ def initialize(actual, matcher_1, matcher_2)
 
             inner, outer = order_block_matchers
 
+            returned = nil
             @match_results[outer] = outer.matches?(Proc.new do |*args|
-              @match_results[inner] = inner.matches?(inner_matcher_block(args))
+              p = Proc.new { |*args2|
+                returned = inner_matcher_block(args).call(*args2)
+              }
```
JonRowe, line 169: This needs to take care of ruby2 keywords (e.g. marked as such, not having keywords itself) so that they get passed through, I'm also a bit nervous about `returned` why isn't it just implict from the assignment?

JonRowe (changes requested):
Sorry for the delay reviewing this @pirj definietly a good idea and good work but needs some tweaks

zverok:
(A bit of a philosophical note, as I was tagged...)

I can’t say that I am fond of this solution. The original idea for `return_value` matcher, for me, is more in the spirit of RSpec (and I actually [have one in saharspec](https://github.com/zverok/saharspec?tab=readme-ov-file#expect--block-to-retvalue-matcher), called just `ret`, as you are probably aware).

The solution with `satisfy { boolean check }` seems to go against the RSpec spirit of “every check should be built from combining matchers,” the spirit that allows to produce clear explanations of “what is expected” in `--format doc` mode, but which requires to write matchers in “RSpec DSL” instead of pure Ruby (e.g., we write `expected.to eq 5` not `expected.to == 5`).

This is a choice that is questioned by some, but it allows to build powerful, combinable checks, which very clearly report “what went wrong” when the specs are red, able to produce nice diffs depending on the context of the check and so on.

Even the original `satisfy { any code }` looked like an “escape hatch” for me (basically, “when you are in a hurry, or the check is so unique that it is easier then make a new small clear matcher”), and extending its usage for new cases... I am not sure if I see it justified.

May I ask why the original `return_value(other_matcher)` idea was abandoned?..

JonRowe:
Migrated to the monorepo as rspec/rspec#150

PR #1479, opened 2024-08-18
Require an argument for the Include matcher

### LINKED ISSUE #1478: `to include()` should fail
### Subject of the issue

I just made a not-so-obvious mistake in one of my tests. I think the `include` matcher is missing some edge cases when it is used incorrectly. For example, `include` should always require at least 1 argument. 

### Your environment
* Ruby version: ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]
* rspec-expectations version: rspec-expectations-3.13.0

### Steps to reproduce

```
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "3.7.0" # Activate the gem and version you are reporting the issue against.
end

require 'rspec/autorun'

RSpec.describe 'additions' do
  it 'does not have b' do
    expect([{ a: 1 }]).to include
    {
      b: 1
    }
  end

  it 'really does not have b' do
    expect([{ a: 1 }]).not_to include({
      b: 1
    })
  end
end
```

### Expected behavior

The first test should fail. 

### Actual behavior

The first test passes.

nevinera:
Resolves #1478 

## Purpose

There are some edge-case ways you can try to invoke this matcher that end up parsing as an argument-less call and a separate expression; this will stop that from happening.

## Implementation
Treat instantiation of the `Include` matcher with no arguments as an ArgumentError.

### Note

This might count as a breaking change - while it _seems_ like something nobody would ever want to do, the case where this can happen legitimately comes up when the caller is taking an array to splat _into_ the `include` matcher, and ignoring the empty case because it currently passes.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@ Bug Fixes:
   (Eric Mueller, #1455)
 * Use `RSpec.warning` for an expectation warning rather than `Kernel.warn`. (Jon Rowe, #1472)
 * Prevent mismatched use of block and value matchers in compound expectations. (Phil Pirozhkov, #1476)
+* Raise an error when passing no arguments to the `include` matcher. (Eric Mueller, #1479)
 
 Enhancements:
 
--- a/lib/rspec/matchers/built_in/include.rb
+++ b/lib/rspec/matchers/built_in/include.rb
@@ -13,6 +13,7 @@ module RSpec
 
         # @api private
         def initialize(*expecteds)
+          raise(ArgumentError, 'Include matcher requires at least one argument') if expecteds.empty?
           @expecteds = expecteds
         end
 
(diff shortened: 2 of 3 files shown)


juanca:
Thanks for making the PR! Looks good to me 🫠

pirj (approved):
Thank you!

pirj:
@JonRowe we’re [thinking](https://github.com/rspec/rspec-expectations/issues/1478#issuecomment-2295754795) about making this a warning, to follow how we did with `raise_error(nil)`. What is your take on this?

JonRowe:
I'm happier with as an error, its a bug in the matcher due to implementation that should never have worked. I actually played around with this a bit improving the error and there are other things that can break (like the description) when there are no expected.

JonRowe:
Thank you @nevinera

JonRowe:
Released in 3.13.2

matthew-puku:
Thanks for this! Caught a few `include(*empty_array)` cases in our test suite.

PR #1485, opened 2024-09-05
Fix `include(/regexp/)` when used without a count.

### LINKED ISSUE #1481: Behaviour of `include` against a String with a Regexp matcher "feels" inconsistent
### Subject of the issue

With an rspec matcher such as:

```ruby
expect("Hello, World!").to include(/Hello/)
```

there are some unclear behaviours/messages

### Your environment
* Ruby version: 3.3.3
* rspec-expectations version: 3.13.2

### Steps to reproduce

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue LoadError => e
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rspec", "~> 3.13.0"
  gem "rspec-expectations", "~> 3.13.0"
end

require 'rspec/autorun'

RSpec.describe "Include Regexp" do
  it "should work with a regexp without a count" do
    expect("Hello, World!").to include(/Hello/)
  end
  it "works with a regexp and a count" do
    expect("Hello, World!").to include(/l/).exactly(3).times
  end
  it "works with a non-matching regexp" do
    expect("Hello, World!").to_not include(/bob/)
  end
  it "should work with a non-matching regexp and count" do
    # I realize that this expectation is ambiguous, as a literal reading means that both 
    # "no matches" or "more than one match" would pass, but at the very least
    # a better error message would be helpful
    expect("Hello, World!").to_not include(/bob/).once
  end
end
```

### Expected behavior

All specs should pass

### Actual behavior

```
❯ ruby test.rb
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
Ruby version is: 3.3.3
F.F.

Failures:

  1) Include Regexp should work with a regexp without a count
     Failure/Error: expect("Hello, World!").to include(/Hello/)

     TypeError:
       no implicit conversion of Regexp into String
     # test.rb:22:in `block (2 levels) in <main>'

  2) Include Regexp works with a non-matching regexp
     Failure/Error: expect("Hello, World!").to_not include(/bob/)

     TypeError:
       no implicit conversion of Regexp into String
     # test.rb:28:in `block (2 levels) in <main>'
```

JonRowe:
Fixes #1481 there was a bug where a regexp and a count contraint would work with the `include` matcher, but not a bare regexp e.g.

```
# worked
expect("astring").to include(/a/).once

# failed
expect("astring").to include(/a/)
```

This makes this consistent with the docs as working, in another version of RSpec we may look to deprecate this usage as it is not doing an `include?` call but using scan instead.

### DIFF
--- a/features/built_in_matchers/include.feature
+++ b/features/built_in_matchers/include.feature
@@ -4,6 +4,7 @@ Feature: `include` matcher
 
   ```ruby
   expect("a string").to include("a")
+  expect("a string").to include(/str..g/)
   expect("a string").to include(/a|str/).twice
   expect("a string").to include("str", "g")
   expect("a string").not_to include("foo")
@@ -80,6 +81,7 @@ Feature: `include` matcher
       RSpec.describe "a string" do
         it { is_expected.to include("str") }
         it { is_expected.to include("a", "str", "ng") }
+        it { is_expected.to include(/str..g/) }
         it { is_expected.to include(/a|str/).twice }
         it { is_expected.not_to include("foo") }
         it { is_expected.not_to include("foo", "bar") }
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thanks!

JonRowe:
Released in 3.13.3

PR #1486, opened 2024-09-19
Fix output.to_stdout_from_any_process when $stdout has been reassigned

jdelStrother:
(An alternate fix for https://github.com/rspec/rspec-core/pull/3113)

When running under rspec bisect runner, $stdout & $stderr are reassigned to a StringIO, which would cause "can't convert Tempfile into StringIO" when trying to reopen the stream in CaptureStreamToTempfile

Since CaptureStreamToTempfile reopens the stream rather than reassigning it, I _think_ it makes sense to use STDOUT rather than whatever the current value of $stdout is.

(Please feel free to sanity-check that assumption. I've gone round in circles a few too many times on this)

### DIFF
--- a/lib/rspec/matchers/built_in/output.rb
+++ b/lib/rspec/matchers/built_in/output.rb
@@ -46,7 +46,7 @@ module RSpec
         # Works when subprocesses print to stdout as well.
         # This is significantly (~30x) slower than `to_stdout`
         def to_stdout_from_any_process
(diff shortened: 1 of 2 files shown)


jdelStrother:
I've pushed a possible fix, where instead of reopening `$stderr`, it reopens `STDERR` and reassigns `$stderr`.
I'm not entirely sure it's going to work, but see what you think and I can squash it if desired.

JonRowe:
Migrated to the monorepo as rspec/rspec#151

